PR #1, opened 2015-04-13
Configuration file edits

Kegsay:
This modifies the structure of ``config.yaml``, specifically the IRC ``servers`` part. A lot of the keys there "organically" grew as more features were added, without much thought to how things should be structured going forward. Given the config file is *effectively the main end-user API* I thought this had to be changed. It's now structured something like:
```
ircService:
  servers:
    irc.example.com:
      botConfig:
        nick: "dev_as_bot"
        password: "helloworld"

      privateMessages:
        enabled: true

      dynamicChannels:
        enabled: false
        visibility: "public"
        aliasTemplate: "#irc_$SERVER_$CHANNEL"

      mappings:
        "#foobar": ["!gPXedTwNCdZGAsaHGd:localhost"]

      matrixClients:
        userTemplate: "@irc_$NICK"
        displayName: "$NICK (on $SERVER)"

      ircClients:
        nickTemplate: "M-$DISPLAY"
        maxClients: 30
```
Related fields are now clumped together, and there is support for ``$VARIABLES``. This gives much more flexibility than the prefix system which existed before. 

All of this adds more complexity to the config file however, and I knew that without any form of validation this wasn't going to work. This is why this PR adds a dependency on ``JaySchema`` which is a JSON Schema library. This lets the AS define a schema for ``config.yaml`` to prevent the user missing required fields, using wrong types, not specifying some $VARIABLES (e.g. a matrix user ID without ``$NICK``), etc. Finally, this is all wrapped up in a ``Config`` object which sets sensible defaults, meaning that the core library doesn't have to be programmed as defensively as it was in the past, as we can ensure certain fields will always be defined.

Documentation has been modified to explain each field, and I'm opting to link to ``config.sample.yaml`` rather than maintaining two sets of configs (one in ``README`` and one in the sample) to reduce maintenance burden.


ara4n:
having the name of the config block be the server hostname seems a mistake - what if you need to specify a port for the server? or whether the server is SSL or not? or use some weird ipv4/ipv6 address? Surely it's better to use an array of servers, or possibly a map of labels to servers, and then separately have the server hostname/port/ssl/etc config embedded within the service defn?

Kegsay:
Both ``port`` and ``ssl`` are keys within the server itself. I'll admit, I'm unsure how much I like having the server hostname as a key, and I'm open for changing it.

PR #2, opened 2015-07-06
Room history visibility

erikjohnston:
Add support for setting `m.room.history_visibility`.

This requires a yet unreleased version of synapse.

### DIFF
--- a/lib/mxlib/matrix.js
+++ b/lib/mxlib/matrix.js
@@ -168,16 +168,11 @@ MatrixLib.prototype.createRoomWithAlias = function(alias, name, topic, joinRule,
             return setJoinRule(
               client, roomId, joinRuleToSet
             ).then(function(roomId) {
-              var d = setRoomHistoryVisibility(client, roomId, "joined");
-              self.log.info("setRoomHistoryVisibility2 -> %s", d);
-              return d;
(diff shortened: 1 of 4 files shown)


erikjohnston:
Also, I haven't figured out how to fix the tests yet.

erikjohnston:
The tests should now be fixed.

lib/mxlib/matrix.js
```diff
@@ -322,7 +319,19 @@ var setJoinRule = function(client, roomId, joinRule) {
...
+    });
+};
+
+var setRoomHistoryVisibility = function(client, roomId, history_visibility) {
+    return client.sendStateEvent(
+        roomId, "m.room.history_visibility", {
+            "history_visibility": history_visibility
+        }, ""
+    ).then(function() {
+        return roomId;
```
ara4n: Why do we return roomId here?  Surely we already know it and nothing's happened that would cause it to change?

erikjohnston: Its purely to make the chaining of the deferreds above slightly nicer. In the end it probably doesn't make that much difference, since we would just have to move the `.then(function() {return roomId;})` to the call sites above.

Kegsay:
LGTM

PR #3, opened 2015-08-04
Matrix membership list mirroring

Kegsay:
This PR:
 - Adds support for membership list syncing
 - Reimplements the IRC client connection logic
 - Sends a new metric (IRC connections) to statsd
 - Adds LONGNAM~1 username support (with irc_clients.db)


Kegsay:
This has been stable for at least a week, merging into `master`.

PR #4, opened 2015-08-10
Hot reloading

Kegsay:
This PR refactors the structure of the IRC AS and adds incomplete support for hot reloading. Specifically, it listens for `SIGUSR2` and doesn't do a lot else. This PR is being made now as the decision has been made with Matthew to focus on modularising the bridging semantics from the IRC-specific semantics. This would mean that there would be "modules" which could interact with the bridge layer; which means we can load/reload them at will to get hot code reloading behaviour.


Kegsay:
Closed in favour of PR #5

PR #5, opened 2015-08-11
Kill off matrix bridge and restructure project

Kegsay:
This PR adds an `enabled: true|false` option to `botConfig`. If `enabled: false`, the bot will not be present on any channel. Messages from IRC are relayed via collective sharing on the IRC client connections (see https://github.com/matrix-org/matrix-appservice-irc/blob/4f3f42bf36ab88844edfa8a768d178e83e227d64/lib/irclib/client-events.js#L2 for an overview of the logic). This PR also changes the behaviour of Matrix -> IRC images/files, which are now sent as NOTICEs instead of coming from the bot.

This PR also refactors the structure of the IRC AS and adds incomplete support for hot reloading. Specifically, it listens for SIGUSR2 and doesn't do a lot else. This PR is being made now as the decision has been made with Matthew to focus on modularising the bridging semantics from the IRC-specific semantics. This would mean that there would be "modules" which could interact with the bridge layer; which means we can load/reload them at will to get hot code reloading behaviour.


leonerd:
Any particular reason for using `SIGUSR2` for this? `SIGHUP` is the traditional "please reload things" signal for this kind of purpose.

Kegsay:
To quote the Node.js docs:
> SIGHUP is generated on Windows when the console window is closed, and on other platforms under various similar conditions, see signal(7). It can have a listener installed, however node will be unconditionally terminated by Windows about 10 seconds later. On non-Windows platforms, the default behaviour of SIGHUP is to terminate node, but once a listener has been installed its default behaviour will be removed.

I would use `SIGUSR1` but Node.js reserves that for attaching a debugger.

Kegsay:
Been stable for about a week, nothing ominous in the logs/metrics.

PR #6, opened 2015-10-21
Implement dynamic channel federation option

stevenhammerton:


lib/irclib/server.js
```diff
@@ -25,6 +25,10 @@ IrcServer.prototype.getJoinRule = function() {
     return this.config.dynamicChannels.joinRule;
 };
 
+IrcServer.prototype.getFederate = function() {
```
Kegsay: `shouldFederateRooms` may be more descriptive.

spec/integ/dynamic-channels.spec.js
```diff
@@ -82,6 +82,92 @@ describe("Dynamic channels", function() {
         });
     });
 
+    it("should create non federated room when joining channel and federation is enabled",
```
Kegsay: wha-? Do you mean "should create **a federated room**" rather than "non"?

stevenhammerton:
Kegsay, both comments addressed :)

Kegsay:
LGTM! Thanks!

PR #7, opened 2015-11-17
Patches to allow for Self Signed IRC Certs

bkruger99:
See subject.

Ran into an issue where I couldn't connect to a server with a self-signed cert / server that doesn't provide a CA Intermediate.   This solves that problem. (Though the error messages were less than useful that it was a SSL issue, but I've not looked at node.js before today)


Kegsay:
LGTM - Thanks!

PR #9, opened 2015-12-06
Add support for unique per-user IPv6 addresses

davidar:
I've tried to clean up [Tor's patch](https://xt.gg/matrix.ircbridge.ipv6range.diff.txt) for [BOTS-113](https://matrix.org/jira/browse/BOTS-113). I haven't had a chance to test this yet, but wanted to submit it for feedback in the meantime.

CC: @ara4n

### DIFF
--- a/HOWTO.md
+++ b/HOWTO.md
@@ -313,6 +313,16 @@ Logging is configurable in the yaml, but there is also an extra verbose setting
 you can enable. This is done by passing ``--verbose`` or ``-v`` to 
 ``node app.js``.
 
+### Unique IPv6 addresses
+You may want to assign unique IPv6 addresses to the generated IRC clients
+e.g. to scope bans to Matrix users rather than the entire application service.
+To enable this:
+```yaml
+ircService:
+  uniqueIPv6:
+    prefix: "2001:0db8:85a3::"  # modify appropriately
+```
+
 Contributing
 ------------
 
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -241,3 +241,9 @@ ircService:
   # dump .db files to. This is relative to the project directory.
   # Required.
   databaseUri: "nedb://data"
+  
+  # Optional. The IPv6 prefix to use for generating unique addresses for each
+  # connected user. If not specified, all users will connect from the same
+  # (default) address.
+  uniqueIPv6:
+    prefix: "2001:0db8:85a3::"
--- a/lib/irc-appservice.js
+++ b/lib/irc-appservice.js
@@ -11,6 +11,7 @@ var matrixToIrc = require("./bridge/matrix-to-irc.js");
 var ircToMatrix = require("./bridge/irc-to-matrix.js");
 var membershiplists = require("./bridge/membershiplists.js");
 var IrcServer = require("./irclib/server.js").IrcServer;
+var ircClient = require("./irclib/client.js");
 var ircLib = require("./irclib/irc.js");
 var matrixLib = require("./mxlib/matrix");
 var MatrixUser = require("./models/users").MatrixUser;
@@ -35,6 +36,9 @@ module.exports.configure = function(config) {
     if (config.ident.enabled) {
         ident.configure(config.ident);
         ident.run();
+    }
+    if (config.uniqueIPv6.prefix) {
+        ircClient.ip6Prefix = config.uniqueIPv6.prefix;
     }
 
     dbConnPromise = store.connectToDatabase(config.databaseUri);
--- a/lib/irclib/client.js
+++ b/lib/irclib/client.js
@@ -13,6 +13,12 @@ var log = require("../logging").get("irc-client");
 // The length of time to wait before trying to join the channel again
 var JOIN_TIMEOUT_MS = 15 * 1000; // 15s
 
+// A global counter for generating unique numerical user IDs
+var userCount = 0;
+
+// IPv6 prefix to use
+var ip6Prefix = null;
+
 /**
  * Create a new bridged IRC client.
  * @constructor
@@ -60,6 +66,9 @@ BridgedClient.prototype.setIrcUserInfo = function(ircUser) {
     this.server = ircUser.server;
     this.nick = ircUser.nick;
     this.password = ircUser.password;
+    if (ircUser.localAddress) {
+        this.localAddress = ircUser.localAddress;
+    }
     this.userId = this.matrixUser ? this.matrixUser.userId : ircUser.username;
 };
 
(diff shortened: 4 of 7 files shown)


HOWTO.md
```diff
@@ -313,6 +313,16 @@ Logging is configurable in the yaml, but there is also an extra verbose setting
 you can enable. This is done by passing ``--verbose`` or ``-v`` to 
 ``node app.js``.
 
+### Unique IPv6 addresses
+You may want to assign unique IPv6 addresses to the generated IRC clients
+e.g. to scope bans to Matrix users rather than the entire application service.
+To enable this:
+```yaml
+ircService:
+  uniqueIPv6:
```
Kegsay: Please drop the `unique` on this to be just `ipv6`.

lib/irc-appservice.js
```diff
@@ -36,6 +37,9 @@ module.exports.configure = function(config) {
         ident.configure(config.ident);
         ident.run();
     }
+    if (config.uniqueIPv6.prefix) {
+        ircClient.ip6Prefix = config.uniqueIPv6.prefix;
```
Kegsay: Everything in the config will already be assigned to the `IrcServer` - see https://github.com/matrix-org/matrix-appservice-irc/blob/master/lib/irclib/server.js#L17 - Please get the IPv6 prefix from there via an accessor like https://github.com/matrix-org/matrix-appservice-irc/blob/master/lib/irclib/server.js#L61

config.sample.yaml
```diff
@@ -241,3 +241,9 @@ ircService:
   # dump .db files to. This is relative to the project directory.
   # Required.
   databaseUri: "nedb://data"
+  
+  # Optional. The IPv6 prefix to use for generating unique addresses for each
+  # connected user. If not specified, all users will connect from the same
+  # (default) address.
+  uniqueIPv6:
```
Kegsay: This is *not* assigning unique IPv6 addresses for each *Matrix* connection, so it doesn't make much sense being here. Please move this to the `ircClients` section.

lib/irclib/client.js
```diff
@@ -13,6 +13,12 @@ var log = require("../logging").get("irc-client");
 // The length of time to wait before trying to join the channel again
 var JOIN_TIMEOUT_MS = 15 * 1000; // 15s
 
+// A global counter for generating unique numerical user IDs
+var userCount = 0;
```
Kegsay: This will not create a unique 1:1 mapping from user ID to IRC client. Restarting the bridge will lead to users getting each others IPv6 addresses. Given these addresses are used for identification for bans etc, we can't really do this. They need to be calculated and stored similar to ident usernames.

lib/irclib/client.js
```diff
@@ -60,6 +66,9 @@ BridgedClient.prototype.setIrcUserInfo = function(ircUser) {
     this.server = ircUser.server;
     this.nick = ircUser.nick;
     this.password = ircUser.password;
+    if (ircUser.localAddress) {
+        this.localAddress = ircUser.localAddress;
+    }
```
Kegsay: Why are you doing this? When will `ircUser` ever have `localAddress` set?

lib/irclib/client.js
```diff
@@ -85,13 +94,22 @@ BridgedClient.prototype.connect = function(callbacks) {
...
             nick: nameInfo.nick,
             username: nameInfo.username,
             realname: nameInfo.realname,
             password: self.password
-        }, function(inst) {
+        };
+        if (ip6Prefix) {
+            var suffix = userCount.toString(16);
+            // insert colons: "deadbeef" => "dead:beef"
+            suffix = suffix.replace(/\B(?=(.{4})+(?!.))/g, ':');
```
Kegsay: `suffix.match(/.{1,4}/g).join(":")` is a simpler regex to grok.

lib/irclib/client.js
davidar: Not sure, it was in Tor's original patch, so I assumed it was there for a reason, but perhaps not.

config.sample.yaml
davidar: I modelled it after the `ident` config, but that does make more sense, thanks.

lib/irclib/client.js
davidar: That doesn't work properly though --- it produces 'abcd:ef' when it should be 'ab:cdef'

lib/irclib/client.js
davidar: Hmm, that makes things trickier. How should I do this?

lib/irclib/client.js
Kegsay: Then can you either:
 - explain what that regex is doing better than `// insert colons: "deadbeef" => "dead:beef"` which implies it is inserting `:` every 4 characters.
 - Rewrite the regex so it is easier to grok.

lib/irclib/client.js
Kegsay: Indeed, this is in fact the trickiest part of ipv6 support. We already have to do this when generating ident usernames, so it should be possible to crib the control flow mechanics from it.

See https://github.com/matrix-org/matrix-appservice-irc/blob/master/lib/irclib/names.js for how ident username generation works - you're interested in the queueing mechanics and https://github.com/matrix-org/matrix-appservice-irc/blob/master/lib/irclib/names.js#L139 which persists the user to the DB. You will need to:
 - Add `ipv6address` to https://github.com/matrix-org/matrix-appservice-irc/blob/master/lib/models/users.js#L41
 - Add it as a persisted key in https://github.com/matrix-org/matrix-appservice-irc/blob/master/lib/store.js#L451
 - Repopulate the field on the getters - https://github.com/matrix-org/matrix-appservice-irc/blob/master/lib/store.js#L434 and https://github.com/matrix-org/matrix-appservice-irc/blob/master/lib/store.js#L488

Then you need to add the glue to make it work on each client connection:
 - The ident stuff does it here: https://github.com/matrix-org/matrix-appservice-irc/blob/4a2dc0ed0a09e746aef640fa05d30bc7534dfc8f/lib/irclib/client.js#L83
 - You would add another promise in the chain to `getIpv6Address` or something, then use it in `clientConnection.create()`

Testing wise I would expect:
 - A test which does *not* set `localAddress` if there is no ipv6 prefix supplied.
 - A test which does set `localAddress` if there is an ipv6 prefix supplied (which is correct)
 - A test which checks your queueing behaviour (sending 2 matrix messages from the same user ID to the bridge on the same tick of the event loop, should get assigned the same ipv6 address (if locks are not taken out correctly this could race))
 - A test which checks your counting (sending 2 matrix messages, one from user A, one from user B, should get sequentially higher addresses assigned.

How would the bridge respond if it consumed all the ipv6 block it had been allocated? Refuse to make new connections? Cycle them around from 0 again (how would that interact with bans on the network)?

lib/irclib/client.js
Kegsay: Remove it and see what breaks! :D

Kegsay:
I'll most likely be taking up the helm for IPv6 support within the next 2 weeks.

Kegsay:
Better late than never: I should finally be adding experimental support for IPv6 in the next few days on the `develop` branch.

Kegsay:
This will land hopefully tomorrow, if not Friday.

Kegsay:
Closing this PR as support has now landed on `develop`. No idea how well it works though since we don't have an IPv6 env here which I can test with! Please let me know how it goes.

PR #13, opened 2016-01-19
Switch promises from q to bluebird

illicitonion:


lib/bridge/membershiplists.js
```diff
@@ -68,14 +69,15 @@ module.exports.getChannelsToJoin = function(server) {
...
+        var onDone = function() {
             var chans = Object.keys(channels);
             log.debug(
                 "getChannelsToJoin => %s should be synced: %s",
                 chans.length, JSON.stringify(chans)
             );
             defer.resolve(chans);
-        });
+        };
+        Promise.all(promises).then(onDone, onDone);
```
Kegsay: Is this actually equivalent with `allSettled`?

lib/irclib/irc.js
```diff
@@ -185,13 +186,15 @@ var getBotClient = function(server) {
...
 module.exports.connect = function() {
-    var defer = q.defer();
+    var defer = promiseutil.defer();
     store.waitForDatabase().done(function() {
         var promises = [];
         servers.forEach(function(server) {
             promises.push(loginToServer(server));
         });
-        q.allSettled(promises).then(function() {
+        Promise.all(promises).then(function() {
```
Kegsay: `allSettled`
Returns a promise that is fulfilled with an array of promise state snapshots, but only **after all the original promises have settled**, i.e. become either fulfilled **or rejected**.

`Promise.all`
Given an Iterable(arrays are Iterable), or a promise of an Iterable, which produces promises (or a mix of promises and values), iterate over all the values in the Iterable into an array and return a promise that is fulfilled when all the items in the array are fulfilled. The promise's fulfillment value is an array with fulfillment values at respective positions to the original array. **If any promise in the array rejects**, the returned promise is rejected with the rejection reason

So AFAIK, the promise will resolve early with bluebird the moment any one of the promises rejects, but this doesn't happen with Q.

spec/util/test.js
```diff
@@ -76,5 +76,9 @@ module.exports.beforeEach = function(testCase, env) {
         });
         env.mockAsapiController = env.asapiMock.create();
     }
+
+    process.on("unhandledRejection", function(reason, promise) {
+        throw new Error("Unhandled rejection: " + reason);
```
Kegsay: I worry this will stop any useful stack traces. Can you confirm stack traces are still okay?

lib/irclib/irc.js
Kegsay: Looks like you want `.reflect()` - http://bluebirdjs.com/docs/api/reflect.html

lib/bridge/membershiplists.js
illicitonion: Implemented allSettled for realsies.

I can see promiseutil getting more used in the future :(

lib/irclib/irc.js
illicitonion: Alas even those docs lied. Implemented the whole thing myself.

spec/util/test.js
illicitonion: The stack trace is useless, but the error bubbles up to the top level, which is the important thing.

I'm not sure how to get a better stack trace, given this is all happening in an event loop...

illicitonion: Done

lib/irclib/irc.js
illicitonion: Done

lib/bridge/membershiplists.js
illicitonion: Done

Kegsay:
LGTM

PR #14, opened 2016-01-25
matrix-to-irc.js still uses q, should use bluebird

roblabla:
q has been removed in 3fd3ab5a75f5d29d5ee6ce9027a6acb4fd00a8f2 but not every references to it has been switched to bluebird. This PR fixes it.


Kegsay:
Thanks for the PR! I saw this when doing https://github.com/matrix-org/matrix-appservice-irc/pull/16 and fixed it there, so I'm going to close this one.

PR #15, opened 2016-01-25
Add an entry in "bin"

roblabla:
With this patch, `npm install -g matrix-appservice-irc` puts a `matrix-appservice-irc` command in the PATH.


Kegsay:
LGTM - Thanks!

PR #19, opened 2016-01-28
Use the Bridge class from matrix-appservice-bridge

Kegsay:
This PR basically takes a bunch of code from `main.js` and shoves it in a new file which can be instanced(ish). There's still a lot wrong with it, but this is a good stopping point to prevent the diff getting too bad.

Notably, it finally makes a `Bridge` from `matrix-appservice-bridge`. This is then injected via proxyquire for tests (so it's actually also executing `matrix-appservice-bridge` code, which has subtly different semantics wrt return values, hence functional changes on tests).

The roadmap for finishing this off:
 - Remove `models/requests.js` - We can use the bridge implementation now.
 - Remove `mxlib` entirely - We can use `Intent` and `ClientFactory` from the bridge lib now.
 - Remove `models/users.js` and `models.rooms.js` - We can replace them with bridge-classes for users and rooms respectively.
 - Finally, replace `store.js` and point the bridge impl at the real thing (we currently have 2 stores side-by-side as of this PR, the bridge one is unused).

Once that is done, the IRC AS will "officially" be using `matrix-appservice-bridge` in its entirety. At this point we could possibly do a release to `master` since the churn will have stopped. We can then:
 - Persist nicks
 - Persist IPv6 addresses
 - Fix error semantics (`throw`ing actual `Error` objects)
 - etc...

### DIFF
--- a/lib/bridge/instance.js
+++ b/lib/bridge/instance.js
@@ -17,7 +17,7 @@ var log = logging.get("IrcBridge");
 var Bridge = require("matrix-appservice-bridge").Bridge;
 
 const DELAY_TIME_MS = 10 * 1000;
-const DEAD_TIME_MS = 60 * 1000 * 5; // 5min
+const DEAD_TIME_MS = 5 * 60 * 1000;
 
 
 function IrcBridge(config, registration) {
@@ -137,6 +137,7 @@ IrcBridge.prototype._onEvent = Promise.coroutine(function*(request, context) {
         if (!event.content || !event.content.membership) {
             return;
         }
+        // MatrixUser(userId, displayName, isVirtual)
         var target = new MatrixUser(event.state_key, null, null);
         var sender = new MatrixUser(event.user_id, null, null);
         if (event.content.membership === "invite") {
--- a/lib/bridge/instance.js
+++ b/lib/bridge/instance.js
@@ -70,11 +70,10 @@ IrcBridge.prototype.run = Promise.coroutine(function*(port) {
         userStore: "_unused_users.db", // we use our own for now
         suppressEcho: false, // we use our own dupe suppress for now
         queue: {
-            type: "per_room",
-            perRequest: true
+            type: "none",
+            perRequest: false
         }
     });
-
 
     if (this.config.appService) {
         console.warn(
@@ -116,7 +115,7 @@ IrcBridge.prototype.run = Promise.coroutine(function*(port) {
     this.ircServers.forEach(function(server) {
         membershiplists.sync(server);
     });
-    this.bridge.run(port);
+    yield this.bridge.run(port);
     this.bridge.getRequestFactory().addDefaultTimeoutCallback((req) => {
         this.onLog("[" + req.getId() + "] DELAYED (" + req.getDuration() + "ms)");
     }, DELAY_TIME_MS);
@@ -132,7 +131,7 @@ IrcBridge.prototype.onEvent = function(request, context) {
 IrcBridge.prototype._onEvent = Promise.coroutine(function*(request, context) {
     var event = request.getData();
     if (event.type === "m.room.message" || event.type === "m.room.topic") {
-        yield matrixToIrc.onMessage(event)
+        yield matrixToIrc.onMessage(request, event)
     }
     else if (event.type === "m.room.member") {
         if (!event.content || !event.content.membership) {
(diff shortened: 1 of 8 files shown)


lib/bridge/instance.js
```diff
@@ -0,0 +1,175 @@
...
+var names = require("../irclib/names.js");
+var IrcServer = require("../irclib/server.js").IrcServer;
+var matrixLib = require("../mxlib/matrix");
+var MatrixUser = require("../models/users").MatrixUser;
+var logging = require("../logging");
+var log = logging.get("IrcBridge");
+var Bridge = require("matrix-appservice-bridge").Bridge;
+
+const DELAY_TIME_MS = 10 * 1000;
+const DEAD_TIME_MS = 60 * 1000 * 5; // 5min
```
illicitonion: 5 * 60 * 1000 or 1000 * 60 * 5 would both be more clear than 60 * 1000 * 5, and probably wouldn't need that comment

lib/bridge/instance.js
```diff
@@ -0,0 +1,175 @@
...
+function IrcBridge(config, registration) {
+    this.config = config;
+    this.registration = registration;
+    this.ircServers = [];
+    this.bridge = null; // Bridge
+}
+
+IrcBridge.prototype.run = Promise.coroutine(function*(port) {
+    // connect to the DB, blow away any old config mappings, we're setting new ones now.
+    yield store.connectToDatabase(this.config.ircService.databaseUri);
```
illicitonion: This is surprisingly side-effecty. Surely this should return a database connection, rather than presumably set global state?

lib/bridge/instance.js
```diff
@@ -0,0 +1,175 @@
...
+    yield store.rooms.removeConfigMappings();
+
+    // maintain a list of IRC servers in-use
+    let serverDomains = Object.keys(this.config.ircService.servers);
+    for (var i = 0; i < serverDomains.length; i++) {
```
illicitonion: forEach?

lib/bridge/instance.js
```diff
@@ -0,0 +1,175 @@
...
+IrcBridge.prototype._onEvent = Promise.coroutine(function*(request, context) {
+    var event = request.getData();
+    if (event.type === "m.room.message" || event.type === "m.room.topic") {
+        yield matrixToIrc.onMessage(request, event)
+    }
+    else if (event.type === "m.room.member") {
+        if (!event.content || !event.content.membership) {
+            return;
+        }
+        var target = new MatrixUser(event.state_key, null, null);
```
illicitonion: Throw in /*displayName=*/ and /*isVirtual=*/ before the nulls, for clarity

lib/bridge/instance.js
```diff
@@ -0,0 +1,175 @@
...
+    yield matrixToIrc.onUserQuery(matrixUser.userId);
+    return null; // don't provision, we already do atm
+});
+
+IrcBridge.prototype.onAliasQuery = Promise.coroutine(function*(alias, aliasLocalpart) {
+    yield matrixToIrc.onAliasQuery(alias);
+    return null; // don't provision, we already do atm
+});
+
+IrcBridge.prototype.onLog = function(line, isError) {
```
illicitonion: isError is never true?

lib/bridge/instance.js
Kegsay: I completely agree, but you're having a laugh if you think I'm doing that in this PR :)

lib/bridge/instance.js
Kegsay: It can be; `appservice-bridge` lib calls this.

lib/bridge/instance.js
Kegsay: done

lib/bridge/instance.js
Kegsay: Can't really do this *nicely*. `forEach` means a new `function` is required, and inside the `for` we `yield`, so you need to `forEach` a generator function. You can't even do arrow functions `*() =>` with generators :/ and need to IIFE the function `Promise.coroutine()` returns. In all, it looks terrible and is confusing.

lib/bridge/instance.js
Kegsay: done

illicitonion:
LGTM

PR #20, opened 2016-01-29
Escape HTML entities when sending IRC -> Matrix

Kegsay:
This prevents munging occurring between IRC formatting and textual < element >
references, whereby if you sent a tag and some colour codes from IRC it would
not escape the tag and therefore send invalid HTML to Matrix.

Added integration test for this case.

### DIFF
--- a/lib/irclib/formatting.js
+++ b/lib/irclib/formatting.js
@@ -33,6 +33,22 @@ var STYLE_UNDERLINE = '\u001f';
 var STYLE_CODES = [STYLE_BOLD, STYLE_ITALICS, STYLE_UNDERLINE];
 var RESET_CODE = '\u000f';
 
+function escapeHtmlChars(text) {
+    return text
+         .replace(/&/g, "&amp;")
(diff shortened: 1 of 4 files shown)


lib/irclib/formatting.js
```diff
@@ -33,6 +33,22 @@ var STYLE_UNDERLINE = '\u001f';
...
+         .replace(/&/g, "&amp;")
+         .replace(/</g, "&lt;")
+         .replace(/>/g, "&gt;")
+         .replace(/"/g, "&quot;")
+         .replace(/'/g, "&#39;"); // to work on HTML4 (&apos; is HTML5 only)
+}
+
+module.exports.stripIrcFormatting = function(text) {
+    // strip colors
+    return text.replace(/(\x03\d{0,2}(,\d{0,2})?|\u200B)/g, '').replace(
```
illicitonion: Wrap on .s not (s

illicitonion:
LGTM

PR #21, opened 2016-01-29
Use the RequestFactory and ClientFactory from appservice-bridge

Kegsay:
- Standardise on `appservice-bridge` `Request` classes
- Make `onLeave` use coroutines.
- Removed `requests.js`
- Removed `cs-extended-sdk.js`
- This also starts to unpick the singletons to work on instances, but there's a ton of work left to do there.


illicitonion:
LGTM

PR #23, opened 2016-02-16
Deterministically generate config file

illicitonion:


leonerd:
LGTM I guess; since it is oberved to DTRT.

PR #24, opened 2016-03-03
Remove mxlib (Phase 1)

Kegsay:
This is the first stage to removing `mxlib` entirely. Some functions have been removed. M->I and I->M logic is now instantiated and aren't singletons. The bridge now relies on `Intent` from the bridge lib.


lib/bridge/instance.js
```diff
@@ -107,11 +116,14 @@ IrcBridge.prototype.run = Promise.coroutine(function*(port) {
             This release will use values from this config file. This will produce
             a fatal error in a later release.`
         );
+        this.domain = this.config.appService.homeserver.domain;
+        this.appServiceUserId = (
+            "@" + (this.config.appService.localpart || IrcBridge.DEFAULT_LOCALPART) + ":" +
+            this.config.homeserver.domain
```
illicitonion: Use this.domain?

Right now we have three things:

this.domain == this.config.appService.homeserver.domain ?= this.config.homeserver.domain

lib/bridge/instance.js
```diff
@@ -107,11 +116,14 @@ IrcBridge.prototype.run = Promise.coroutine(function*(port) {
...
+        );
+
         matrixLib.setMatrixClientConfig({
-            baseUrl: this.config.appService.homeserver.url,
-            accessToken: this.config.appService.appservice.token,
-            domain: this.config.appService.homeserver.domain,
-            localpart: this.config.appService.localpart || IrcBridge.DEFAULT_LOCALPART
+            domain: this.config.appService.homeserver.domain
```
illicitonion: this.domain

lib/bridge/instance.js
```diff
@@ -121,28 +133,27 @@ IrcBridge.prototype.run = Promise.coroutine(function*(port) {
                 "FATAL: Registration file is missing a sender_localpart and/or AS token."
             );
         }
+        this.domain = this.config.homeserver.domain;
+        this.appServiceUserId = (
+            "@" + this.registration.getSenderLocalpart() + ":" +
+            this.config.homeserver.domain
```
illicitonion: this.domain

lib/bridge/instance.js
```diff
@@ -121,28 +133,27 @@ IrcBridge.prototype.run = Promise.coroutine(function*(port) {
...
+        );
+
         matrixLib.setMatrixClientConfig({
-            baseUrl: this.config.homeserver.url,
-            accessToken: this.registration.getAppServiceToken(),
-            domain: this.config.homeserver.domain,
-            localpart: this.registration.getSenderLocalpart()
+            domain: this.config.homeserver.domain
```
illicitonion: this.domain

lib/bridge/matrix-handler.js
```diff
@@ -103,11 +110,20 @@ var handleInviteFromUser = Promise.coroutine(function*(req, event, invitedIrcUse
...
     req.log.info("Joined %s to room %s", invitedUser.userId, event.room_id);
-    let isPmRoom = yield req.mxLib.isPmRoom(
-        invitedUser.userId, event.room_id, event.user_id
+
+    // check if this room is a PM room or not.
+    let intent = this.ircBridge.bridge.getIntent(invitedUser.userId);
+    let roomState = yield intent.roomState(event.room_id);
+    let joinedMembers = roomState.filter((ev) => {
+        return ev.type === "m.room.member" && ev.content.membership === "join";
+    }).map((ev) => {
```
illicitonion: (ev) => ev.state_key

Kegsay:
add to whitelist

Kegsay:
ok to test

lib/bridge/matrix-handler.js
```diff
@@ -303,9 +319,9 @@ var _onInvite = Promise.coroutine(function*(req, event, inviter, invitee) {
     // mark this room as being processed in case we simultaneously get
     // messages for this room (which would fail if we haven't done the
     // invite yet!)
-    processingInvitesForRooms[event.room_id + event.state_key] = req.getPromise();
-    req.getPromise().finally(function() {
-        processingInvitesForRooms[event.room_id + event.state_key] = undefined;
+    this._processingInvitesForRooms[event.room_id + event.state_key] = req.getPromise();
+    req.getPromise().finally(() => {
+        this._processingInvitesForRooms[event.room_id + event.state_key] = undefined;
```
illicitonion: Why not delete this._processingInvitesForRooms[event.room_id + event.state_key] ?

Kegsay:
ok to test

lib/bridge/matrix-handler.js
Kegsay: They're for all intents and purposes equivalent. `delete` probably makes the intent clearer though, so will change.

lib/bridge/matrix-handler.js
```diff
@@ -568,25 +584,78 @@ let _onAliasQuery = Promise.coroutine(function*(req, roomAlias) {
...
+                    creation_content: {
+                        "m.federate": channelInfo.server.shouldFederate()
+                    }
+                }
+            });
+
+            newRoomId = res.room_id;
+
+            if (shouldSetJoinRule) {
+                yield botIntent.sendStateEvent(newRoomId, "m.room.join_rules", "", {
```
illicitonion: Why not include these initial states under the initial_state param of create_room, so that the server applies them atomically?

lib/bridge/matrix-handler.js
```diff
@@ -568,25 +584,78 @@ let _onAliasQuery = Promise.coroutine(function*(req, roomAlias) {
...
-            channelInfo.server.getJoinRule(),
-            channelInfo.server.shouldFederate(),
-            channelInfo.server.shouldPublishRooms()
-        );
+        req.log.info("Bot is now tracking IRC channel.");
+
+        // ======== Create the Matrix room
+        let newRoomId = null;
+        let botIntent = this.ircBridge.bridge.getIntent();
+        try { // make the matrix room
```
illicitonion: Drop the scope of your try to the things which could actually fail (i.e. the createRoom call, and if they remain, the sendStateEvent calls)

lib/bridge/matrix-handler.js
```diff
@@ -568,25 +584,78 @@ let _onAliasQuery = Promise.coroutine(function*(req, roomAlias) {
...
+                });
+            }
+
+            yield botIntent.sendStateEvent(newRoomId, "m.room.history_visibility", "", {
+                history_visibility: "joined"
+            });
+        }
+        catch (e) {
+            if (e && e.errorcode === "M_UNKNOWN") {
+                // alias already taken, must be us. Join the room alias.
```
illicitonion: The intent pattern seems to be to try to join, and then do the creation if it failed. Why is this the other way around?

lib/mxlib/matrix.js
```diff
@@ -25,36 +24,12 @@ var mkUserId = function(localpart) {
...
-        });
-        promiseutil.allSettled(promises).then(function() {
-            d.resolve();
-        });
-    });
-    return d.promise;
-};
-
-module.exports.decodeMxc = function(mxcUri) {
+module.exports.decodeMxc = function(mxcUri) { // actions.js (de-singleton) TODO
```
illicitonion: What does this comment mean?

package.json
```diff
@@ -30,7 +30,7 @@
     "jayschema": "^0.3.1",
     "js-yaml": "^3.2.7",
     "matrix-appservice": "^0.2.3",
-    "matrix-appservice-bridge": "^0.1.5",
+    "matrix-appservice-bridge": "matrix-org/matrix-appservice-bridge#develop",
```
illicitonion: eww??

lib/bridge/instance.js
Kegsay: You're absolutely right, and I think this may have been causing the flakiness before when the freenode bridge was on `master` (it would've set an empty domain for user IDs).

lib/bridge/instance.js
Kegsay: done

lib/bridge/instance.js
Kegsay: done

lib/bridge/instance.js
Kegsay: done

lib/bridge/matrix-handler.js
Kegsay: done

lib/bridge/matrix-handler.js
Kegsay: done

lib/bridge/instance.js
Kegsay: done

package.json
Kegsay: This is just temporary so Jenkins doesn't whine (since it `npm installs` from scratch). I don't want to constantly do releases each time I require a bug fix / new method to be added to the bridge lib atm.

lib/mxlib/matrix.js
Kegsay: `actions.js` needs to have extra context of the HS URL so it can convert MXC URIs to HTTP URIs correctly. This could be done either by instantiating actions with some surrounding context, or explicitly passing the HS URL around the place when creating messages.

lib/bridge/matrix-handler.js
Kegsay: Good idea! I forgot that existed until recently.

lib/bridge/matrix-handler.js
Kegsay: Done

lib/bridge/matrix-handler.js
Kegsay: Done

lib/bridge/matrix-handler.js
Kegsay: We don't expect the room to exist, so the nominal case is faster if we make a room with the correct alias. Occasionally that will fail, in which case we can then join the room. In contrast, the intent pattern by default assumes if you try to join a room that you must think it already exists.

lib/bridge/matrix-handler.js
```diff
@@ -568,25 +582,69 @@ let _onAliasQuery = Promise.coroutine(function*(req, roomAlias) {
...
-        );
+        req.log.info("Bot is now tracking IRC channel.");
+
+        // ======== Create the Matrix room
+        let newRoomId = null;
+        let botIntent = this.ircBridge.bridge.getIntent();
+        try { // make the matrix room
+            let res = yield botIntent.createRoom({
+                options: {
+                    room_alias_name: roomAlias.split(":")[0].substring(1), // localpart
```
illicitonion: Does anything guarantee roomAlias has non-zero length?

Kegsay: No, but that is fine. `<empty string>.split(":") == [ "" ]` which then `[0] == ""` which then `"".substring(1) == ""`

illicitonion:
LGTM

PR #29, opened 2016-05-01
Fix config being broken with multiple irc servers

### LINKED ISSUE #26: Settings are mixed up between IRC servers
Settings are mixed up between IRC servers.
For example:

```
ircService:
  servers:
    irc.server1.net:
      port: 9999
    irc.server2.net:
      port: 6667
```

matrix-appservice-irc will sometimes try to connect to irc.server1.net on port 6667, or irc.server2.net on port 9999. It seems to be order dependent; I can get it to work by reordering the servers.

roblabla:
The extend module works by doing an in-place replacement on its first argument.

Fixes #26


Kegsay:
LGTM, thanks!

Issue #68, opened 2016-06-08
DB performance when using `matrix-appservice-bridge@0.3.6` is not acceptable


Kegsay:
*This issue is currently blocking a release from `develop` to `master`, and is nominally affecting the Freenode deployment on Matrix.org (the other IRC networks like Moznet are running `develop` happily as their workloads are less intense).*

---

### The issue
When you try to run `develop` on the Freenode deployment, things start well but quickly begin to slow down to a crawl before almost stopping. Logs begin to slow down then stop. CPU usage on the box indicates it is spinning at 100% CPU usage.

### Investigation
After CPU profiling with `node-inspector`, it became clear this was a problem with NeDB. It was starving the event loop from executing tasks because it kept running this function: https://github.com/louischatriot/nedb/blob/master/lib/datastore.js#L309

More investigation indicated that the cause of this was that the IRC bridge was triggering tens of thousands of `find` queries, which NeDB was attempting to service. Adding in trace logging to the `DataStore` calls indicates that this is a problem with `getIrcChannelsForRoomId` which would be called with around 1000 room IDs on startup. This triggers 2 `find` queries (one to fetch the mapping, one to fetch the mapped room data).

### Proposed solutions
 - Batch `find` statements using `$in` - The bridge knows the full set of rooms it needs to query, but then queries them one at a time.
 - Change the database format to store 1 row per complete mapping, ditching the idea of room entries. This would halve the number of queries as a single `find` would produce all the data required, rather than triggering another `find` to retrieve the rest of the room data.

Kegsay:
So I spent time batching up `getIrcChannelsForRoomId` to reduce startup delay and it worked, which is great. However, general traffic is still generating more work than the event loop can handle, resulting in a gradual drift of latency for messages to go through the bridge: about 250ms every 5 minutes (so after 20 minutes there is a solid 1s extra delay, ad infinitum).

Kegsay:
CPU profiling shows that NeDB is *still* a huge component of the overall time spent (98% of CPU time over 3.3 minutes (!) - still in the same function, spending its life doing `match`.

The only times we touch NeDB at all is via the `DataStore` class which I have now instrumented with trace logging, so we can see which methods are being called when the bridge is operating normally. With a bit of Python-fu to parse the logs, the main methods invoked are (tallys were over a 10 minute period in the logs):
```javascript
{
  'getIrcChannelsForRoomId': 519,
  'getMatrixUserByLocalpart': 296,
[...]

PR #75, opened 2016-06-17
Improve performance of room store

Kegsay:
This is the culmination of https://github.com/matrix-org/matrix-appservice-irc/issues/68 - the following changes have been made:
 - Make the IRC bridge rely on `matrix-appservice-bridge@1.0.0` to pull in the new database format.
 - Add an upgrade script `scripts/upgrade-db-0.2-to-0.3.js` to port over to the new format.
 - Add a batch version of `getLinkedRemoteRooms` to reduce the number of queries at startup.
 - Add indexes on Room Store's `matrix_id`, `remote_id` and User Store's `localpart` to hasten lookups as they are called frequently.
 - Cache the promise for `_getSyncableRooms` to prevent 2x `/initialSync`s being triggered on startup (one to get the list of channels to join, one to inject join events).

This PR also tidies up a lot of metrics and adds a few more:
 - Join/Part metrics for IRC/Matrix are now sent as counters.
 - If a server doesn't mirror joins/parts, it will not treat the joining/parting request as `FAILED`, but instead ignore it (so as to avoid skewing error metrics).
 - Fix a problem where metrics I->M would not be sent.

### DIFF
--- a/lib/DataStore.js
+++ b/lib/DataStore.js
@@ -121,23 +121,21 @@ DataStore.prototype.setPmRoom = function(ircRoom, matrixRoom, userId, virtualUse
 DataStore.prototype.getMatrixPmRoom = function(realUserId, virtualUserId) {
     log.debug("getMatrixPmRoom " + realUserId);
     var linkKey = createPmLinkKey(realUserId, virtualUserId);
-    return this._roomStore.getLinksByKey(linkKey).then(function(links) {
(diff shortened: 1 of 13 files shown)


Kegsay:
Keeping this PR around until Monday to see how it performs. If there are no disasters, this will be merged to `develop`.

Kegsay:
No disasters as far as I can see :)

PR #90, opened 2016-07-08
Add the `!whois` command for the bridge

dakrone:
This adds a new command - `!whois` for retrieving whois information
about a user in the same way an IRC user would use `/whois <nick>`

b7e

I'm not a Javascript programmer by any means, but I'm happy to make any changes if someone will guide me in what they'd like!


lib/irc/BridgedClient.js
```diff
@@ -218,6 +218,19 @@ BridgedClient.prototype.changeNick = function(newNick) {
     return d.promise;
 };
 
+/**
+ * Get the whois info for an IRC user
+ * @param {string} whoisNick : The nick to call /whois on
+ * @return {Promise<String>} Which resolves to a message to be sent to the user.
+ */
+BridgedClient.prototype.whois = function(whoisNick) {
```
Kegsay: There is already a function with this name on `:280`. This old function used to be used to check if a nick existed on a server, and looks very similar to this function. The old function is not used anywhere anymore (referenced in `IrcBridge.checkNickExists` which is not called). I would use that one instead, and rely on the caller to craft a human-readable message from the result. You may want to expand the returned args from the old `whois` to include more information beyond just "yep they exist!".

Kegsay:
Please base your branch off `develop` rather than `master` :)

Structure/code-wise, this is wonderful, thank you!

dakrone:
Thanks! I'll have to re-open this against `develop`

Issue #81, opened 2016-06-27
Expand non-federate options to PMs


Kegsay:
At the moment, there is a coarse `federate: true|false` option which applies to both dynamic channels and PMs. It would be nice if the bridge could be configured to:
 - Accept/Reject incoming PMs from non-federated sources.
 - Accept/Reject outgoing PMs from non-federated sources.

PR #114, opened 2016-07-26
Optional federation for PM rooms

lukebarnard1:
PM rooms now have the option of being federated or not. Use ```config.privateMessages.federate```. 
By default set to true.

So when a room is created, "m.federate" is set to the value in config. This doesn't stop random private messages from other federated HSs going to the IRC - I will make commits for this shortly.

Integration test written for this (passing locally).

(refers to #81)


spec/integ/pm.spec.js
```diff
@@ -294,4 +294,144 @@ describe("IRC-to-Matrix PMing", function() {
...
+
+    it("should create a federated (by default) 1:1 matrix room and invite the real matrix user when " +
+    "it receives a PM directed at a virtual user from a real IRC user",
+    test.coroutine(function*() {
+        // mock create room impl
+        let createRoomPromise = new Promise(function(resolve, reject) {
+            sdk.createRoom.andCallFake(function(opts) {
+                expect(opts.visibility).toEqual("private");
+                expect(opts.invite).toEqual([tRealUserId]);
+                expect(opts.creation_content["m.federate"]).toEqual(true);
```
Kegsay: This is the only line which is different to the earlier test; I would just do this assertion in that test and then we can remove this test entirely :D

spec/integ/pm.spec.js
```diff
@@ -294,4 +294,144 @@ describe("IRC-to-Matrix PMing", function() {
...
+            roomMapping.server, tRealMatrixUserNick
+        );
+        client.emit(
+            "message", tRealIrcUserNick, tRealMatrixUserNick, tText
+        );
+
+        yield createRoomPromise;
+        yield sentMessagePromise;
+    }));
+});
```
Kegsay: Nit: New line at the end please!

Kegsay:
ok to test

Kegsay:
LGTM aside from rather small comments :smile:

Kegsay:
Ohhhh, but you should be basing off `develop`, not `master`!

PR #115, opened 2016-07-27
Optional federation for PM rooms (based from develop)

lukebarnard1:
Continuation of PR #114, rebased on develop (I also pushed some commits directly to develop - whoops)


lukebarnard1:
Proper testing has now been done with multiple federated HSs:

Two federated HSs, A and B. A has an IRC AS registered. IRC server connected to AS

[IRC]----[AS]----[A]----[B]

When privateMessages.federate = true:
-     PM from A to IRC allowed
-     PM from B to IRC disallowed

When privateMessages.federate = false:
-   PM from A to IRC allowed
-    PM from B to IRC allowed

Issue #28, opened 2016-05-01
Generated regexes in appservice config accept any HS


roblabla:
The namespaces generated by the generate command are too liberal, in that they will also forward messages from remote rooms that have their own aliases.

For example, my AS is configured with a `#freenode_` alias. I would expect it to only bridge channels that have a LOCAL alias e.g. `#freenode_#matrix:cmc.im`. However, the generated config : 

```
namespaces:
  users:
    - exclusive: true
      regex: '@freenode_.*:.*'
  aliases:
    - exclusive: true
      regex: '#freenode_.*:.*'
```

means that my HS is sending events from #freenode_#matrixhq:matrix.org to my AS.

lukebarnard1:
Seems the IrcServer is giving the ```:.*``` part of the regex here: https://github.com/matrix-org/matrix-appservice-irc/blob/master/lib/irc/IrcServer.js#L265

This is then added to the registration file when generated. So just need replace ```.*``` with the server domain.

eternaleye:
#190 and #232 seem to be duplicates of this - was the fix incomplete?

Kegsay:
Yes, there are still global catch all regexps around.

PR #117, opened 2016-07-28
Regex in generated registration file restricted

lukebarnard1:
The regex for user ids and aliases has been modified to include by default the configured HS domain from config.yaml.

Refers to #28

### DIFF
--- a/lib/irc/IrcServer.js
+++ b/lib/irc/IrcServer.js
@@ -276,7 +276,7 @@ IrcServer.prototype.getAliasRegex = function(hsdomain) {
             "$CHANNEL": ".*"  // the nick is unknown, so replace with a wildcard
         },
         // Only match the domain of the HS
-        ":" + hsdomain
+        ":" + escapeRegExp(hsdomain)
     );
 };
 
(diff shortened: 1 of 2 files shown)


lib/irc/IrcServer.js
```diff
@@ -266,7 +266,7 @@ IrcServer.prototype.getNick = function(userId, displayName) {
     return nick;
 };
 
-IrcServer.prototype.getAliasRegex = function() {
+IrcServer.prototype.getAliasRegex = function(hsdomain) {
```
Kegsay: `hsDomain` or `homeserverDomain` please 0:)

lib/irc/IrcServer.js
```diff
@@ -275,13 +275,12 @@ IrcServer.prototype.getAliasRegex = function() {
         {
             "$CHANNEL": ".*"  // the nick is unknown, so replace with a wildcard
         },
-        // The regex applies to the entire alias, so add a wildcard after : to
-        // match all domains.
-        ":.*"
+        // Only match the domain of the HS
+        ":" + hsdomain
```
Kegsay: This is returning a regular expression. Domains unsurprisingly include `.`s. We should escape them.

Issue #56, opened 2016-06-02
Auto-pastebin large messages to avoid spamming


Kegsay:
https://matrix.org/jira/browse/BOTS-64

sbts:
Ideally this would be configurable per channel (command to the AS user)
General public channels would want pastebining, while special purpose channels may not, eg: #ledgersmb we are very happy with the current behavior.

Kegsay:
I'm not sure how much chan ops should get a say though - as I was under the impression anti-spam bots are network-wide rather than per-channel -- So even if you guys are happy with it, I can't allow it if it means it leaves black marks on the bridge (see #101)

jansol:
AFAIK there are network-wide flood limits, but some channels employ their own bots for enforcing even stricter rules.

sbts:
@Kegsay well the current rate-limiting behaviour doesn't seem to trigger any issues with the network side anti-spambots, and we are trying to encourage users in general to move over to native matrix clients instead of IRC.
However we do need to retain the IRC bridging for a long long time to come as it's been the primary support channel for many many years and there are users that will simply expect it to exist when they have an issue in 10 years time :-(

Kegsay:
> some channels employ their own bots for enforcing even stricter rules

That's fine, since it will just affect the user sending the messages being kicked and/or banned. My primary concern is the overall bridge health.

> well the current rate-limiting behaviour doesn't seem to trigger any issues with the network side anti-spambots

[It did trigger issues](https://github.com/matrix-org/matrix-appservice-irc/issues/101), which is why I'm pushing for this issue to be fixed.

> However we do need to retain the IRC bridging for a long long time to come as it's been the primary support channel for many many years and there are users that will simply expect it to exist when they have an issue in 10 years time :-(

I'm not following your thought pattern here. Bridging will still work, just large (for whatever definition of large) messages will be uploaded into Matrix first (effectively as a `.txt` file) and then that URL will be sent instead - much like how we currently do for images. What am I missing here?

PR #118, opened 2016-07-29
Large Blocks of SPAM Uploaded as File

lukebarnard1:
In response to issue #56, the IRC bridge will now protect itself against large amounts of SPAM (which has been defined arbitrarily as anything with more than N lines) being sent across to IRC, and thus getting itself banned. When messages that are too long are sent from the HS, the bridge attempts to upload a plaintext file to the HS. If successful, it will send a URL to IRC in the same way IRC deals with files and images. If it fails to upload the file, the first N lines will be sent to IRC with a message explaining why the file could not be uploaded.

N has arbitrarily been set to 2. This should be refactored to the config file, which I can do. Any ideas as to what the default number of lines should be?

### DIFF
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -297,6 +297,18 @@ IrcBridge.prototype.sendMatrixAction = function(room, from, action, req) {
     return Promise.reject(new Error("Unknown action: " + action.type));
 };
 
+IrcBridge.prototype.sendMatrixPlaintextFile = function(plaintext, req) {
+    if (req) {
+        req.log.info("sendMatrixPlaintextFile");
+    }
+    let intent = this._bridge.getIntent();
+    return intent.getClient().uploadContent({
+        stream: new Buffer(plaintext),
+        name: Date().replace(/\s/g,'_') + ".txt",
+        type: "text/plain",
+    });
+};
+
 IrcBridge.prototype.getMatrixUser = Promise.coroutine(function*(ircUser) {
     let matrixUser = null;
     let userLocalpart = ircUser.server.getUserLocalpart(ircUser.nick);
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -302,7 +302,7 @@ IrcBridge.prototype.sendMatrixPlaintextFile = function(fileName, plaintext, req)
         req.log.info("sendMatrixPlaintextFile");
     }
     let intent = this._bridge.getIntent();
-    
+
     return intent.getClient().uploadContent({
         stream: new Buffer(plaintext),
         name: fileName,
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -297,14 +297,15 @@ IrcBridge.prototype.sendMatrixAction = function(room, from, action, req) {
     return Promise.reject(new Error("Unknown action: " + action.type));
 };
 
-IrcBridge.prototype.sendMatrixPlaintextFile = function(plaintext, req) {
+IrcBridge.prototype.sendMatrixPlaintextFile = function(fileName, plaintext, req) {
     if (req) {
         req.log.info("sendMatrixPlaintextFile");
     }
     let intent = this._bridge.getIntent();
+    
     return intent.getClient().uploadContent({
         stream: new Buffer(plaintext),
-        name: Date().replace(/\s/g,'_') + ".txt",
+        name: fileName,
         type: "text/plain",
     });
 };
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -297,13 +297,12 @@ IrcBridge.prototype.sendMatrixAction = function(room, from, action, req) {
     return Promise.reject(new Error("Unknown action: " + action.type));
 };
 
-IrcBridge.prototype.sendMatrixPlaintextFile = function(fileName, plaintext, req) {
+IrcBridge.prototype.uploadTextFile = function(fileName, plaintext, req) {
     if (req) {
-        req.log.info("sendMatrixPlaintextFile");
+        req.log.info("uploadTextFile");
     }
-    let intent = this._bridge.getIntent();
 
-    return intent.getClient().uploadContent({
+    return this._bridge.getIntent().getClient().uploadContent({
         stream: new Buffer(plaintext),
         name: fileName,
         type: "text/plain",
--- a/lib/bridge/MatrixHandler.js
+++ b/lib/bridge/MatrixHandler.js
@@ -770,64 +770,77 @@ MatrixHandler.prototype._onMessage = Promise.coroutine(function*(req, event) {
 });
 
 MatrixHandler.prototype._sendIrcAction = Promise.coroutine(
-    function*(req, ircRoom, ircUser, ircAction, event) {
+    function*(req, ircRoom, ircClient, ircAction, event) {
     let self = this;
 
     if (event.content.msgtype !== "m.text" ||
-        !(ircUser.unsafeClient && ircUser.unsafeClient.wouldSend)) {
-        yield self.ircBridge.sendIrcAction(ircRoom, ircUser, ircAction);
+        !(ircClient.unsafeClient && ircClient.unsafeClient.wouldSend)) {
+        yield self.ircBridge.sendIrcAction(ircRoom, ircClient, ircAction);
+        return;
     }
-    else {
-        let text = event.content.body;
-        let wouldSend = ircUser.unsafeClient.wouldSend(ircRoom.channel, text);
-        let lineLimit = 2;
-        if (wouldSend.length > lineLimit) {
-            // Message body too long - send a file to HS instead and send link to IRC
 
-            try {
-                // Upload as a file and get URI
+    let text = event.content.body;
 
-                let fileName = Date().replace(/\s/g, '_') + ".txt";
-                let response = yield self.ircBridge.sendMatrixPlaintextFile(fileName, text);
-                let result = JSON.parse(response);
-                event.content.url = result.content_uri;
+    // Generate an array of individual messages that would be sent
+    let potentialMessages = ircClient.unsafeClient.wouldSend(ircRoom.channel, text);
+    let lineLimit = 3;
+    if (potentialMessages.length < lineLimit) {
+        yield self.ircBridge.sendIrcAction(ircRoom, ircClient, ircAction);
+        return;
+    }
 
-                // Send the link to IRC
-                event.content.msgtype = "m.file";
-                event.content.body = fileName;
+    // Message body too long, upload to HS instead
+    try {
+        // Use the current time as the name of the file
+        //  strip off day of week and timezone
+        //  replace whitespace with underscores
+        // result e.g : Aug_01_2016_13:41:44.txt
+        let fileName = new Date().toString()
+            .match(/.*?\s(.*\:[0-9]*\:[0-9]*).*/)[1]
+            .replace(/\s/g,'_') + ".txt";
 
-                let bigFileIrvAction = IrcAction.fromMatrixAction(
-                    MatrixAction.fromEvent(
-                        self.ircBridge.getAppServiceBridge().getClientFactory().getClientAs(), event
-                    )
-                );
+        // Try to upload as a file and get URI
+        //  (this could fail, see the catch statement)
+        let response = yield self.ircBridge.sendMatrixPlaintextFile(fileName, text);
+        let result = JSON.parse(response);
 
-                yield self.ircBridge.sendIrcAction(ircRoom, ircUser, bigFileIrvAction);
-            }
-            catch (err) {
-                req.log.error("Failed to upload file, will truncate and send instead", err);
+        // Alter event object so that it is treated as if a file has been uploaded
+        event.content.url = result.content_uri;
+        event.content.msgtype = "m.file";
+        event.content.body = fileName;
 
-                event.content = {
-                    url : undefined,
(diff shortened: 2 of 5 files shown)


lib/bridge/MatrixHandler.js
```diff
@@ -770,6 +769,68 @@ MatrixHandler.prototype._onMessage = Promise.coroutine(function*(req, event) {
     yield Promise.all(promises);
 });
 
+MatrixHandler.prototype._sendIrcAction = Promise.coroutine(
+    function*(req, ircRoom, ircUser, ircAction, event) {
+    let self = this;
```
Kegsay: You don't need `self` here at all - use `this` directly instead.

ara4n:
anything more than 3 consecutive lines is generally frowned on - so i 
suggest n=3.

lib/bridge/IrcBridge.js
```diff
@@ -297,6 +297,19 @@ IrcBridge.prototype.sendMatrixAction = function(room, from, action, req) {
     return Promise.reject(new Error("Unknown action: " + action.type));
 };
 
+IrcBridge.prototype.sendMatrixPlaintextFile = function(fileName, plaintext, req) {
+    if (req) {
+        req.log.info("sendMatrixPlaintextFile");
+    }
+    let intent = this._bridge.getIntent();
```
Kegsay: You can probably in-line this into line `:306`.

lib/bridge/MatrixHandler.js
```diff
@@ -770,6 +769,68 @@ MatrixHandler.prototype._onMessage = Promise.coroutine(function*(req, event) {
     yield Promise.all(promises);
 });
 
+MatrixHandler.prototype._sendIrcAction = Promise.coroutine(
+    function*(req, ircRoom, ircUser, ircAction, event) {
```
Kegsay: I know you're just using the same named variables as before, but the name `ircUser` is terrible - that variable *isn't* an IRC user, it's a `BridgedClient`. \*chastises past self\* - please rename this to `client` or `ircClient` or something like that.

lib/bridge/MatrixHandler.js
```diff
@@ -770,6 +769,68 @@ MatrixHandler.prototype._onMessage = Promise.coroutine(function*(req, event) {
...
+    let self = this;
+
+    if (event.content.msgtype !== "m.text" ||
+        !(ircUser.unsafeClient && ircUser.unsafeClient.wouldSend)) {
+        yield self.ircBridge.sendIrcAction(ircRoom, ircUser, ircAction);
+    }
+    else {
```
Kegsay: As a rule of thumb, try to return early so you can outdent e.g this:
```javascript
function foo() {
  if (some condition) {
    // if block
    return;
  }
  // else block goes here
}
```

Rather than what you have here:
```javascript
function foo() {
  if (some condition) {
    // if block
  }
  else {
    // else block
  }
}
```

Two reasons for this:
 - Less indentation (which is always good)
 - Less chance of you accidentally executing code on the `if` block you weren't expecting to e.g:

    ```javascript
    function foo() {
      if (some condition) {
        // if block
      }
      else {
        // else block
      }
      // extra code you didn't mean to execute on the if block
    }
```

You will also find this follows your mental model a lot better as the "success path" of the function tends down the page rather than being sprawled with splits from conditional statements.

File: lib/bridge/MatrixHandler.js [hunk 5]
```diff
@@ -770,6 +769,68 @@ MatrixHandler.prototype._onMessage = Promise.coroutine(function*(req, event) {
...
+        let text = event.content.body;
+        let wouldSend = ircUser.unsafeClient.wouldSend(ircRoom.channel, text);
+        let lineLimit = 2;
+        if (wouldSend.length > lineLimit) {
```
Kegsay: `wouldSend` is a misleading variable name - I intuitively thought it would be a `boolean` but it turns out it is an `array`?!

lib/bridge/MatrixHandler.js
```diff
@@ -770,6 +769,68 @@ MatrixHandler.prototype._onMessage = Promise.coroutine(function*(req, event) {
...
+                        self.ircBridge.getAppServiceBridge().getClientFactory().getClientAs(), event
+                    )
+                );
+
+                yield self.ircBridge.sendIrcAction(ircRoom, ircUser, trucatedIrcAction);
+            }
+        }
+        else {
+            yield self.ircBridge.sendIrcAction(ircRoom, ircUser, ircAction);
+        }
```
Kegsay: Invert your if statement so you can put this first. There are many lines between when you did that check and here, making it non-obvious when this will be executed. E.g:
```javascript
if (wouldSend.length <= lineLimit) {
  yield self.ircBridge.sendIrcAction(ircRoom, ircUser, ircAction);
  return;
}
// text is too long and needs to be uploaded
// <upload code>
```

lib/bridge/MatrixHandler.js
```diff
@@ -770,6 +769,68 @@ MatrixHandler.prototype._onMessage = Promise.coroutine(function*(req, event) {
...
+        let text = event.content.body;
+        let wouldSend = ircUser.unsafeClient.wouldSend(ircRoom.channel, text);
+        let lineLimit = 2;
+        if (wouldSend.length > lineLimit) {
+            // Message body too long - send a file to HS instead and send link to IRC
+
+            try {
+                // Upload as a file and get URI
+
+                let fileName = Date().replace(/\s/g, '_') + ".txt";
```
Kegsay: It would be polite to state what you're trying to do here e.g:

```javascript
// Use the current time as the name of the file and remove whitespace
let fileName = Date().replace(/\s/g, '_') + ".txt";
```

lib/bridge/MatrixHandler.js
```diff
@@ -770,6 +769,68 @@ MatrixHandler.prototype._onMessage = Promise.coroutine(function*(req, event) {
...
+                let response = yield self.ircBridge.sendMatrixPlaintextFile(fileName, text);
+                let result = JSON.parse(response);
+                event.content.url = result.content_uri;
+
+                // Send the link to IRC
+                event.content.msgtype = "m.file";
+                event.content.body = fileName;
```
Kegsay: Your comment is unhelpful, as you actually do the sending on `:805`. You don't explain why you're clobbering fields on the event though. You should say why you're doing that.

lib/bridge/MatrixHandler.js
```diff
@@ -770,6 +769,68 @@ MatrixHandler.prototype._onMessage = Promise.coroutine(function*(req, event) {
...
+
+                let bigFileIrvAction = IrcAction.fromMatrixAction(
```
Kegsay: `IrvAction` huh? 0:)

lib/bridge/MatrixHandler.js
```diff
@@ -770,6 +769,68 @@ MatrixHandler.prototype._onMessage = Promise.coroutine(function*(req, event) {
...
+                    )
+                );
+
+                yield self.ircBridge.sendIrcAction(ircRoom, ircUser, bigFileIrvAction);
+            }
+            catch (err) {
+                req.log.error("Failed to upload file, will truncate and send instead", err);
+
+                event.content = {
+                    url : undefined,
```
Kegsay: Well done for catching this, but it is subtle and worthy of a comment why you're doing this, as someone else looking at the code might not spot the problem if this were not done.

lib/bridge/MatrixHandler.js
```diff
@@ -770,6 +769,68 @@ MatrixHandler.prototype._onMessage = Promise.coroutine(function*(req, event) {
...
+                    msgtype : "m.text",
+                    body : wouldSend.splice(0, lineLimit).join('\n') +
+                     "\n(Could not convert large text (truncated above) to file: \"" +
+                     err.message +
+                     "\")"
```
Kegsay: There are 2 ways to make this string nicer rather than having to escape `"` all the time. As in Python, you can use `'` instead which means you don't need to escape the `"`s. An [even nicer form](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Template_literals) exists in ES6.

Kegsay:
LGTM aside from comments.

lib/bridge/MatrixHandler.js
```diff
@@ -770,6 +769,83 @@ MatrixHandler.prototype._onMessage = Promise.coroutine(function*(req, event) {
...
+        yield this.ircBridge.sendIrcAction(ircRoom, ircClient, bigFileIrcAction);
+    }
+    catch (err) {
+        // Uploading the file to HS could fail
+        req.log.error("Failed to upload file, will truncate and send instead", err);
+
+        // Modify the event to become a truncated version of the original
+        //  the truncation limits the number of lines to lineLimit
+        event.content = {
+            url : undefined,
```
Kegsay: You still aren't explaining why you're doing this.

lib/bridge/MatrixHandler.js
```diff
@@ -770,6 +769,83 @@ MatrixHandler.prototype._onMessage = Promise.coroutine(function*(req, event) {
...
+        };
+
+        // Recreate action from modified event
+        let trucatedIrcAction = IrcAction.fromMatrixAction(
+            MatrixAction.fromEvent(
+                this.ircBridge.getAppServiceBridge().getClientFactory().getClientAs(), event
+            )
+        );
+
+        // Send the truncated message action
```
Kegsay: Remove this redundant comment, it's obvious from the code.

lib/bridge/MatrixHandler.js
```diff
@@ -770,6 +769,83 @@ MatrixHandler.prototype._onMessage = Promise.coroutine(function*(req, event) {
...
+        //  strip off day of week and timezone
+        //  replace whitespace with underscores
+        // result e.g : Aug_01_2016_13:41:44.txt
+        let fileName = new Date().toString()
+            .match(/.*?\s(.*\:[0-9]*\:[0-9]*).*/)[1]
+            .replace(/\s/g,'_') + ".txt";
+
+        // Try to upload as a file and get URI
+        //  (this could fail, see the catch statement)
+        let response = yield this.ircBridge.sendMatrixPlaintextFile(fileName, text);
```
Kegsay: Naming: `send` makes me think it is sending the message but that's not what it's really doing, it's *uploading* the text. I would prefer if this were called `uploadTextFile` or equiv (we don't have anywhere else that we can upload to, so the "matrix" bit is somewhat redundant).

lib/bridge/MatrixHandler.js
```diff
@@ -770,6 +769,83 @@ MatrixHandler.prototype._onMessage = Promise.coroutine(function*(req, event) {
...
+        // Recreate action from modified event
+        let bigFileIrcAction = IrcAction.fromMatrixAction(
+            MatrixAction.fromEvent(
+                this.ircBridge.getAppServiceBridge().getClientFactory().getClientAs(), event
+            )
+        );
+
+        // Send the recreated action in place of ircAction
+        yield this.ircBridge.sendIrcAction(ircRoom, ircClient, bigFileIrcAction);
+    }
```
Kegsay: After thinking about this, I think this try/catch is too general and should be made more precise. There's a few reasons for this:
 - The `catch` block doesn't know how far along the `try` got to, whether it uploaded the text file, read the response, or sent the IRC action, so it can't reasonably retry any bits without doing the lot again.
 - Because there is so much in the `try` block, it need to do weird subtle things like `:831` in the `catch` (where if the URL property is set on the `event` but sending the IRC action fails means you then need to back out the changes you made to the event).
 - The `catch` block is assuming failure is due to network errors (as shown by the log line), but this may not be true.

lib/bridge/MatrixHandler.js
```diff
@@ -770,6 +769,83 @@ MatrixHandler.prototype._onMessage = Promise.coroutine(function*(req, event) {
...
+
+    // Message body too long, upload to HS instead
+    try {
+        // Use the current time as the name of the file
+        //  strip off day of week and timezone
+        //  replace whitespace with underscores
+        // result e.g : Aug_01_2016_13:41:44.txt
+        let fileName = new Date().toString()
+            .match(/.*?\s(.*\:[0-9]*\:[0-9]*).*/)[1]
+            .replace(/\s/g,'_') + ".txt";
```
Kegsay: > Note: parsing of date strings with the Date constructor (and Date.parse, they are equivalent) is strongly discouraged due to browser differences and inconsistencies.

https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Date#Several_ways_to_create_a_Date_object

There's no guarantees this will work on other Node versions as they are well within their rights to change it. Node fully supports ES5 though, so you *can* use `new Date().toISOString()` and replace from that string representation instead.

lib/bridge/MatrixHandler.js
```diff
@@ -770,6 +769,90 @@ MatrixHandler.prototype._onMessage = Promise.coroutine(function*(req, event) {
...
+    }
+
+    // Message body too long, upload to HS instead
+        // Use the current time as the name of the file
+        //  strip off day of week and timezone
+        //  replace whitespace with underscores
+        // result e.g : Aug_01_2016_13:41:44.txt
+    let fileName = new Date().toString()
+        .match(/.*?\s(.*\:[0-9]*\:[0-9]*).*/)[1]
+        .replace(/\s/g, '_') + ".txt";
```
Kegsay: Over-indented comment lines `:795 ~ :798`.

Also, my point on https://github.com/matrix-org/matrix-appservice-irc/pull/118#discussion_r73134085 still stands.

Kegsay:
LGTM after the `Date` thing is fixed! :tada:

Kegsay:
LGTM

PR #120, opened 2016-08-01
Provisioning API

lukebarnard1:
A simple API has been provided with the bridge.

Required parameters to the /[un]link POST endpoints are validated so that they roughly resemble the following:

```JSON
{
    "matrix_room_id" : "!Abcdefgh:example.com[:8080]" , 
    "remote_room_server": "localhost", 
    "remote_room_channel" : "#room"
}
```

Tests have been done to make sure that messages are only sent to rooms when they are linked accordingly. However, the virtual users seem to hang around. Not sure if this is an issue.

I will move onto membership syncing tomorrow.

### DIFF
--- a/lib/DataStore.js
+++ b/lib/DataStore.js
@@ -88,18 +88,39 @@ DataStore.prototype.storeRoom = function(ircRoom, matrixRoom, origin) {
         matrixRoom.getId(), ircRoom.get("domain"), ircRoom.channel, origin);
 
     let mappingId = createMappingId(matrixRoom.getId(), ircRoom.get("domain"), ircRoom.channel);
+    return this._roomStore.linkRooms(matrixRoom, ircRoom, {
+        origin: origin
+    }, mappingId);
+};
 
-    let self = this;
+/**
+ * Get an IRC <--> Matrix room mapping from the database.
+ * @param {string} roomId : The Matrix room ID.
+ * @param {string} ircDomain : The IRC server domain.
+ * @param {string} ircChannel : The IRC channel.
+ * @param {string} origin : (Optional) "config" if this mapping was from the config yaml,
+ * "provision" if this mapping was provisioned, "alias" if it was created via aliasing and
+ * "join" if it was created during a join.
+ * @return {Promise}
+ */
+DataStore.prototype.getRoom = function(roomId, ircDomain, ircChannel, origin) {
+    if (typeof origin !== 'undefined' && typeof origin !== 'string') {
+        throw new Error(`If defined, origin must be a string =
+            "config"|"provision"|"alias"|"join"`);
+    }
+    let mappingId = createMappingId(roomId, ircDomain, ircChannel);
 
-    return this._roomStore.selectOne({id : mappingId})
-        .then((doc) => {
-            if (doc) {
-                return Promise.reject(new Error("Room mapping already exists"));
-            }
-            return self._roomStore.linkRooms(matrixRoom, ircRoom, {
-                origin: origin
-            }, mappingId);
-        });
+    let query = {
+        id : mappingId
+    };
+
+    if (origin) {
+        query.data = {
+            origin : origin
+        };
+    }
+
+    return this._roomStore.selectOne(query);
 };
 
 /**
--- a/lib/DataStore.js
+++ b/lib/DataStore.js
@@ -112,7 +112,7 @@ DataStore.prototype.getRoom = function(roomId, ircDomain, ircChannel, origin) {
 
     return this._roomStore.getEntryById(mappingId).then(
         (entry) => {
-            if (origin && origin !== entry.data.origin) {
+            if (origin && entry && origin !== entry.data.origin) {
                 return null;
             }
             return entry;
--- a/lib/DataStore.js
+++ b/lib/DataStore.js
@@ -80,19 +80,13 @@ DataStore.prototype.setServerFromConfig = Promise.coroutine(function*(server, se
  * @return {Promise}
  */
 DataStore.prototype.storeRoom = function(ircRoom, matrixRoom, origin) {
-    if (typeof origin === 'boolean') {
-        log.warn('Origin must be a string = "config"|"provision"');
-        origin = origin?'config':'unknown';
-    }
-
     if (typeof origin !== 'string') {
-        throw new Error('Origin must be a string = "config"|"provision"');
+        throw new Error('Origin must be a string = "config"|"provision"|"alias"|"join"');
     }
 
     log.info("storeRoom (id=%s, addr=%s, chan=%s, origin=%s)",
         matrixRoom.getId(), ircRoom.get("domain"), ircRoom.channel, origin);
     return this._roomStore.linkRooms(matrixRoom, ircRoom, {
-        from_config: origin === 'config', // kept for backwards compatibility
         origin: origin
     }, createMappingId(matrixRoom.getId(), ircRoom.get("domain"), ircRoom.channel));
 };
--- a/lib/DataStore.js
+++ b/lib/DataStore.js
@@ -260,8 +260,11 @@ DataStore.prototype.getRoomIdsFromConfig = function() {
 DataStore.prototype.removeConfigMappings = function() {
     log.debug("removeConfigMappings ");
     return this._roomStore.removeEntriesByLinkData({
-        from_config: true, // for backwards compatibility
-        origin: 'config'
+        from_config: true // for backwards compatibility
+    }).then(() => {
+        return this._roomStore.removeEntriesByLinkData({
+            origin: 'config'
+        })
     });
 };
 
--- a/lib/DataStore.js
+++ b/lib/DataStore.js
@@ -74,8 +74,9 @@ DataStore.prototype.setServerFromConfig = Promise.coroutine(function*(server, se
  * Persists an IRC <--> Matrix room mapping in the database.
  * @param {IrcRoom} ircRoom : The IRC room to store.
  * @param {MatrixRoom} matrixRoom : The Matrix room to store.
- * @param {string} origin : "config" if this mapping is from the config yaml. 
- * "provision" if this mapping was provisioned.
+ * @param {string} origin : "config" if this mapping is from the config yaml,
+ * "provision" if this mapping was provisioned, "alias" if it was created via
+ * aliasing and "join" if it was created during a join.
  * @return {Promise}
  */
 DataStore.prototype.storeRoom = function(ircRoom, matrixRoom, origin) {
--- a/lib/DataStore.js
+++ b/lib/DataStore.js
@@ -101,7 +101,7 @@ DataStore.prototype.storeRoom = function(ircRoom, matrixRoom, origin) {
  * @param {string} origin : (Optional) "config" if this mapping was from the config yaml,
  * "provision" if this mapping was provisioned, "alias" if it was created via aliasing and
  * "join" if it was created during a join.
- * @return {Promise}
+ * @return {Promise} A promise which resolves to a room entry, or null if one is not found.
  */
 DataStore.prototype.getRoom = function(roomId, ircDomain, ircChannel, origin) {
     if (typeof origin !== 'undefined' && typeof origin !== 'string') {
@@ -110,17 +110,13 @@ DataStore.prototype.getRoom = function(roomId, ircDomain, ircChannel, origin) {
     }
     let mappingId = createMappingId(roomId, ircDomain, ircChannel);
 
-    let query = {
-        id : mappingId
-    };
-
-    if (origin) {
-        query.data = {
-            origin : origin
-        };
-    }
-
-    return this._roomStore.selectOne(query);
+    return this._roomStore.getEntryById(mappingId).then(
+        (entry) => {
+            if (origin && origin !== entry.data.origin) {
+                return null;
+            }
+            return entry;
+        });
 };
 
 /**
--- a/lib/DataStore.js
+++ b/lib/DataStore.js
@@ -74,15 +74,25 @@ DataStore.prototype.setServerFromConfig = Promise.coroutine(function*(server, se
  * Persists an IRC <--> Matrix room mapping in the database.
  * @param {IrcRoom} ircRoom : The IRC room to store.
  * @param {MatrixRoom} matrixRoom : The Matrix room to store.
- * @param {boolean} fromConfig : True if this mapping is from the config yaml.
+ * @param {string} origin : "config" if this mapping is from the config yaml. 
+ * "provision" if this mapping was provisioned.
  * @return {Promise}
  */
-DataStore.prototype.storeRoom = function(ircRoom, matrixRoom, fromConfig) {
-    fromConfig = Boolean(fromConfig);
-    log.info("storeRoom (id=%s, addr=%s, chan=%s, config=%s)",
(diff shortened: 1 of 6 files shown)


Kegsay:
ok to test

lib/DataStore.js
```diff
@@ -198,7 +203,7 @@ DataStore.prototype.getRoomIdsFromConfig = function() {
 DataStore.prototype.removeConfigMappings = function() {
     log.debug("removeConfigMappings ");
     return this._roomStore.removeEntriesByLinkData({
-        from_config: true
+        origin: 'config'
```
Kegsay: This breaks backwards compatibility.

Some existing databases will have `from_config: true` in the `.db` file. When the new code runs, the old config mappings will not be removed because of this.

Port across to the new style by also removing entries where `from_config: true`.

spec/util/app-service-mock.js
```diff
@@ -4,10 +4,47 @@ var util = require("util");
 var instance = null;
 
 function MockAppService() {
+    let self = this;
+
+    this.app = {
+        post: function(path, handler) {
+            // Assume that the first post handler will be for the /link path
```
Kegsay: I would not assume this, and instead just check if the `path === "/_matrix/whatever/the/link/path/is`.

lib/provisioning/Provisioner.js
```diff
@@ -0,0 +1,112 @@
...
+            (err) => { res.status(500).json({error: err.message}); return Promise.reject(err); }
+        );
+    });
+}
+
+var parameterValidation = {
+    matrix_room_id :
+        {regex : /^![0-9A-Za-z]+?:[^:]+(\:[0-9]+)?$/, example : '!Abcdefg:example.com[:8080]'},
+    remote_room_channel :
+        {regex : /^#.+$/, example : '#roomname'},
```
Kegsay: You can be more precise in your regex here. Refer to the [RFC](https://tools.ietf.org/html/rfc2812#section-1.3) - though you will want to allow more than what it says (e.g. a lot of IRCds support unicode and hence things like `#☕`). Things like spaces and non-printable characters should definitely be rejected though.

lib/provisioning/Provisioner.js
```diff
@@ -0,0 +1,112 @@
...
+    remote_room_server :
+        {regex :  /^[a-z\.]+$/, example : 'example.com or localhost'},
```
Kegsay: You need to be more lax than this. IP addresses for example won't work, nor will punycode domain names. Refer to the [RFC](https://tools.ietf.org/html/rfc3986#appendix-A) for the `host` BNF.

lib/provisioning/Provisioner.js
```diff
@@ -0,0 +1,112 @@
...
+        ).then(
+            () => { return this.unlink(req.body) },
+            (err) => { res.status(500).json({error: err.message}); return Promise.reject(err); }
+        );
+    });
+}
+
+var parameterValidation = {
+    matrix_room_id :
+        {regex : /^![0-9A-Za-z]+?:[^:]+(\:[0-9]+)?$/, example : '!Abcdefg:example.com[:8080]'},
```
Kegsay: We don't currently specify a grammar for room IDs in [the spec](http://matrix.org/docs/spec/intro.html#room-ids-and-event-ids) so I would probably just check that it passes something like `!.*:.*`

lib/provisioning/Provisioner.js
```diff
@@ -0,0 +1,112 @@
...
+
+Provisioner.prototype.validate = Promise.coroutine(function*(parameters) {
+    let parameterNames = ['matrix_room_id', 'remote_room_channel', 'remote_room_server'];
+    for (var i = 0; i < parameterNames.length; i++) {
+        let name = parameterNames[i];
+
+        let actual = parameters[name];
+        let valid = parameterValidation[name];
+
+        if (!actual.match(valid.regex)) {
```
Kegsay: NPE and type mismatch risks here. You don't validate that `actual` is not `undefined`, nor do you ensure it is ` string` before calling `match()`.

lib/provisioning/Provisioner.js
```diff
@@ -0,0 +1,112 @@
...
+    this._ircBridge = ircBridge;
+
+    let as = this._ircBridge.getAppServiceBridge().appService;
+
+    as.app.post("/_matrix/provision/link", (req, res) => {
+        return this.validate(req.body).then(
+            () => { res.json({}) },
+            (err) => { res.status(500).json({error: err.message}); return Promise.reject(err); }
+        ).then(
+            () => { return this.link(req.body) },
```
Kegsay: It looks like you're returning the response on `:16` before actually seeing if `this.link` works? It's difficult to see what this is doing overall.

lib/provisioning/Provisioner.js
```diff
@@ -0,0 +1,112 @@
...
+// Link an IRC channel to a matrix room ID
+Provisioner.prototype.link = Promise.coroutine(function*(options) {
+    let ircDomain = options.remote_room_server;
+    let ircChannel = options.remote_room_channel;
+    let roomId = options.matrix_room_id;
+
+    log.info(`Provisioning link for room ${roomId} <---> ${ircChannel}`);
+
+    if (!ircDomain || !ircChannel|| !roomId) {
+        throw new Error("Server domain, ircChannel and room ID are required.");
```
Kegsay: Double validation going on here. Either you trust the input here or you don't. If you don't, you should be calling `validate()`. If you do, then you should assume that the fields are there, and are of the right type.

lib/provisioning/Provisioner.js
```diff
@@ -0,0 +1,112 @@
...
+    }
+
+    // Try to find the domain requested for linking
+    let server = this._ircBridge.getServer(ircDomain);
```
Kegsay: This is probably fine for a first cut, but we should try to handle a few more cases rather than doing only literal matches. E.g. `irc://irc.freenode.net` won't work, since it isn't `== irc.freenode.net`. Pop a `TODO` marker on here so we don't forget.

lib/provisioning/Provisioner.js
```diff
@@ -0,0 +1,112 @@
...
+
+    if (!server) {
+        throw new Error("Server requested for linking not found");
```
Kegsay: Log the input which meant that it wasn't found to aid debugging. In this case, `ircDomain`.

lib/provisioning/Provisioner.js
```diff
@@ -0,0 +1,112 @@
...
+    yield this._ircBridge.getStore().storeRoom(ircRoom, mxRoom, 'provision');
+});
+
+// Unlink an IRC channel from a matrix room ID
+Provisioner.prototype.unlink = Promise.coroutine(function*(options) {
+    let ircDomain = options.remote_room_server;
+    let ircChannel = options.remote_room_channel;
+    let roomId = options.matrix_room_id;
+
+    log.info(`Provisioning unlink for room ${roomId} <-/-> ${ircChannel}`);
```
Kegsay: Love the arrows :smile: - we should also log the `ircDomain` as well though!

lib/provisioning/Provisioner.js
```diff
@@ -0,0 +1,112 @@
...
+    let server = this._ircBridge.getServer(ircDomain);
+
+    if (!server) {
+        throw new Error("Server requested for linking not found");
+    }
+
+    // Delete the room link, but don't update the bridge
+    yield this._ircBridge.getStore()._roomStore.delete({
+        id: roomId + " " + ircDomain + " " + ircChannel
+    });
```
Kegsay: This is very very hacky (accessing private variables, forming the delimited ID on your own rather than letting the `DataStore` do it via `createMappingId`, etc), and you aren't giving any justification as to why you're doing this.

If this is something the bridge needs, then do it properly and add a new method to the `DataStore` and call that, rather than gut-wrenching the `_roomStore` and making assumptions as to the mapping `id` format.

lib/provisioning/Provisioner.js
```diff
@@ -0,0 +1,112 @@
...
+    // Try to find the domain requested for unlinking
+    let server = this._ircBridge.getServer(ircDomain);
+
+    if (!server) {
+        throw new Error("Server requested for linking not found");
+    }
+
+    // Delete the room link, but don't update the bridge
+    yield this._ircBridge.getStore()._roomStore.delete({
+        id: roomId + " " + ircDomain + " " + ircChannel
```
Kegsay: Is this what you *really* want to do? How do you know this mapping wasn't set in the config?

lukebarnard1:
Following the code review, I'm going to write a test to make sure that provisioning doesn't override config-derived mappings.

lib/DataStore.js
```diff
@@ -74,16 +74,78 @@ DataStore.prototype.setServerFromConfig = Promise.coroutine(function*(server, se
...
+
+/**
+ * Get an IRC <--> Matrix room mapping from the database.
+ * @param {string} roomId : The Matrix room ID.
+ * @param {string} ircDomain : The IRC server domain.
+ * @param {string} ircChannel : The IRC channel.
+ * @param {string} origin : (Optional) "config" if this mapping was from the config yaml,
+ * "provision" if this mapping was provisioned, "alias" if it was created via aliasing and
+ * "join" if it was created during a join.
+ * @return {Promise}
```
Kegsay: What does this resolve to? A room entry?

lib/provisioning/Provisioner.js
```diff
@@ -0,0 +1,141 @@
...
+        throw new Error(`Server requested for linking not found ('${ircDomain}')`);
+    }
+
+    // Create rooms for the link
+    let ircRoom = new IrcRoom(server, ircChannel);
+    let mxRoom = new MatrixRoom(roomId);
+
+    yield this._ircBridge.getStore()
+        .getRoom(roomId, ircDomain, ircChannel)
+        .then((doc) => {
```
Kegsay: Weird mixing of coroutines and promises. Just use coroutines :)

```javascript
let doc = yield this._ircBridge.getStore().getRoom(roomId, ircDomain, ircChannel);
// ...
```

lib/provisioning/Provisioner.js
```diff
@@ -0,0 +1,141 @@
...
+    let server = this._ircBridge.getServer(ircDomain);
+
+    if (!server) {
+        throw new Error("Server requested for linking not found");
+    }
+
+    // Delete the room link
+    yield this._ircBridge.getStore()
+        .getRoom(roomId, ircDomain, ircChannel, 'provision')
+        .then((doc) => {
```
Kegsay: https://github.com/matrix-org/matrix-appservice-irc/pull/120/files#r73664764

spec/util/app-service-mock.js
```diff
@@ -4,10 +4,52 @@ var util = require("util");
...
+    this.app = {
+        post: function(path, handler) {
+            if (path === '/_matrix/provision/link') {
+                self.link = handler;
+            }
+            else if (path === '/_matrix/provision/unlink') {
+                self.unlink = handler;
+            }
+            else {
+                throw new Error(`Unrecognised path for mock provisioning endpoint \"${path}\"`);
```
Kegsay: You shouldn't need to escape `"` here, given you're using `.

spec/util/app-service-mock.js
```diff
@@ -4,10 +4,52 @@ var util = require("util");
...
+
     EventEmitter.call(this);
 }
 util.inherits(MockAppService, EventEmitter);
 
+// Simulate a request to the link provisioning API
+//  parameters {object} - the API parameters
+//  statusCallback {function} - Called when the server returns a HTTP response code.
+//  jsonCallback {function} - Called when the server returns a JSON object.
+//  link {boolean} - true if this is a link request (false if unlink).
```
Kegsay: It is more readable in tests if you can see the word "link" and "unlink", rather than `true` and `false` and having to remember which is which. I would just have `_link` and `_unlink` functions on the `MockAppService`.

spec/util/app-service-mock.js
```diff
@@ -4,10 +4,52 @@ var util = require("util");
...
+            }
+        }
+    };
+
     EventEmitter.call(this);
 }
 util.inherits(MockAppService, EventEmitter);
 
+// Simulate a request to the link provisioning API
+//  parameters {object} - the API parameters
```
Kegsay: Mention the word "request" somewhere please 0:)

spec/integ/provisioning.spec.js
```diff
@@ -0,0 +1,368 @@
...
+            }
+
+            // When the _link promise resolves
+            let resolve = shouldSucceed ?
+                // success is indicated with empty object
+                () => { expect(json.calls[0].args[0]).toEqual({}); }:
+                // failure with 500 and JSON error message
+                () => {
+                    expect(json).toHaveBeenCalled();
+                    expect(status).toHaveBeenCalled();
```
Kegsay: Look into [toHaveBeenCalledWith](http://jasmine.github.io/2.0/introduction.html#section-Spies) to reduce the number of separate assertions here.

Kegsay:
LGTM aside from comments - nearly there!

lib/DataStore.js
```diff
@@ -74,16 +74,74 @@ DataStore.prototype.setServerFromConfig = Promise.coroutine(function*(server, se
...
+DataStore.prototype.getRoom = function(roomId, ircDomain, ircChannel, origin) {
+    if (typeof origin !== 'undefined' && typeof origin !== 'string') {
+        throw new Error(`If defined, origin must be a string =
+            "config"|"provision"|"alias"|"join"`);
+    }
+    let mappingId = createMappingId(roomId, ircDomain, ircChannel);
+
+    return this._roomStore.getEntryById(mappingId).then(
+        (entry) => {
+            if (origin && origin !== entry.data.origin) {
```
Kegsay: `entry` can be `null` id no match is found, which will then NPE here.

lib/provisioning/Provisioner.js
```diff
@@ -0,0 +1,136 @@
...
+        throw new Error(`Server requested for linking not found ('${ircDomain}')`);
+    }
+
+    // Create rooms for the link
+    let ircRoom = new IrcRoom(server, ircChannel);
+    let mxRoom = new MatrixRoom(roomId);
+
+    let entry = yield this._ircBridge.getStore().getRoom(roomId, ircDomain, ircChannel);
+    if (!entry) {
+        this._ircBridge.getStore().storeRoom(ircRoom, mxRoom, 'provision');
```
Kegsay: You're dropping the `Promise` returned here to the floor. If `storeRoom` fails for whatever reason, we should fail the `link` Promise but we aren't.

lib/provisioning/Provisioner.js
```diff
@@ -0,0 +1,136 @@
...
+    if (!server) {
+        throw new Error("Server requested for linking not found");
+    }
+
+    // Delete the room link
+    let entry = yield this._ircBridge.getStore()
+        .getRoom(roomId, ircDomain, ircChannel, 'provision');
+
+    if (entry) {
+        this._ircBridge.getStore().removeRoom(roomId, ircDomain, ircChannel, 'provision');
```
Kegsay: You're dropping the `Promise` returned here to the floor. If `storeRoom` fails for whatever reason, we should fail the `link` Promise but we aren't.

Issue #49, opened 2016-06-02
Support sending arbitrary IRC commands through


Kegsay:
https://matrix.org/jira/browse/BOTS-27

The use case here is to send explicit NICK and PASS irc commands.

We need to think about how this will affect the bridge semantics (doing something via raw IRC commands vs via the bridge).

Kegsay:
Ideally this would literally be sending a message with the line-delimited command like:
```
NICK foo
```

But the bridge can support multiple networks, so it'd need to know which network you're on about. I **guess** we could augment it with the name of the domain first (and omit it if there is only 1), e.g:

```
irc.freenode.net
NICK foo
JOIN #bar
irc.somewhere.else
NICK baz
```
- Change nick to 'foo' on freenode
- Join channel #bar on freenode
- Change nick to 'baz' on somewhere

eternaleye:
There's also that some commands ([`UHNAMES`](https://en.wikipedia.org/wiki/List_of_Internet_Relay_Chat_commands#UHNAMES) and [`NAMESX`](https://en.wikipedia.org/wiki/List_of_Internet_Relay_Chat_commands#NAMESX), both of which are invoked via `PROTOCTL`) alter the format the server uses for its messages.

Kegsay:
@eternaleye - interesting to know, thanks. I assume this is a per connection option rather than a per server option, so:
 -  the current bridge will never have seen the different forms before 
 -  implementing this feature would mean the bridge may encounter this format.

Kegsay:
Punting this from the next release, it will be in the next next release.

Kegsay:
The next next release cycle is now upon us... reassigning.

PR #125, opened 2016-08-04
Raw throughput (M->I) IRC commands in admin room

lukebarnard1:
All raw commands (apart from ```PROTOCTL```) can be sent via the admin room on the bridged HS. 

Addresses #49

### DIFF
--- a/lib/bridge/MatrixHandler.js
+++ b/lib/bridge/MatrixHandler.js
@@ -386,7 +386,7 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
         let blacklist = ['PROTOCTL'];
 
         try {
-            let lines = commandBlocks.replace(/\r/, "").split("\n");
+            let lines = commandBlocks.replace(/\r/g, "").split("\n");
             for (var ix = 0; ix < lines.length; ix++) {
                 let lineNumber = ix + 1;
                 let line = lines[ix].trim();
--- a/lib/bridge/MatrixHandler.js
+++ b/lib/bridge/MatrixHandler.js
@@ -386,7 +386,7 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
         let blacklist = ['PROTOCTL'];
 
         try {
-            let lines = commandBlocks.replace("\r").split("\n");
+            let lines = commandBlocks.replace(/\r/, "").split("\n");
             for (var ix = 0; ix < lines.length; ix++) {
                 let lineNumber = ix + 1;
                 let line = lines[ix].trim();
--- a/lib/bridge/MatrixHandler.js
+++ b/lib/bridge/MatrixHandler.js
@@ -11,7 +11,7 @@ var IrcClientConfig = require("../models/IrcClientConfig");
 var MatrixUser = require("matrix-appservice-bridge").MatrixUser;
 var BridgeRequest = require("../models/BridgeRequest");
 var toIrcLowerCase = require("../irc/formatting").toIrcLowerCase;
-
+var Client = require("irc").Client;
 
 function MatrixHandler(ircBridge) {
     this.ircBridge = ircBridge;
@@ -369,7 +369,71 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
         return;
     }
     else {
-        req.log.info("No valid admin command: %s", event.content.body);
+        let fmt = '(COMMAND [arg0 [arg1 [...]]]) | irc.server';
+        req.log.info(`No valid (old form) admin command, will try new format: ${fmt}`);
+
+        // Assumes commands have the form "(COMMAND [arg0 [arg1 [...]]]) | irc.server"
+        let command = event.content.body;
+
+        let matches = command.match(/^([a-z0-9\.:-]+)|(?:([A-Z]+)(?: (.*))?)$/);
+        try {
+            if (!matches) {
+                throw new Error(`Badly formed command`);
+            }
+            let ircDomain = matches[1];
+            let keyword = matches[2];
+            let argsJoined = matches[3];
+
+            if (!argsJoined) {
+                argsJoined = '';
+            }
+
+            if (ircDomain) {
+                // set domain of command to ircDomain
+                let ircServer = this.ircBridge.getServer(ircDomain);
+
+                if (!ircServer) {
+                    throw new Error(`IRC domain not found (${ircDomain})`);
+                }
+
+                this._adminCommandServer = ircServer;
+            }
+            else if (keyword) {
+                if (!this._adminCommandServer) {
+                    throw new Error('Please provide domain before invoking commands');
+                }
+
+                let acceptedKeywords = ['JOIN', 'PART', 'NOTICE', 'TOPIC', 'INVITE'];
+
+                if (acceptedKeywords.indexOf(keyword) == -1) {
+                    throw new Error(`Keyword unrecognised (${keyword})\n`+
+                        'Accepted keywords: ' + acceptedKeywords.join(', '));
+                }
+
+                let args = argsJoined.split(' ');
+
+                let bridgedClient = yield this.ircBridge.getBridgedClient(ircServer, event.user_id);
+
+                let sendArgs = args;
+                args.unshift(keyword);
+
+                if (!bridgedClient.unsafeClient) {
+                    throw new Error('Possibly disconnected');
+                }
+
+                Client.prototype.send.apply(bridgedClient.unsafeClient, sendArgs);
+            }
+            else {
+                throw new Error(`Keyword and arguments required`);
+            }
+        }
+        catch (err) {
+            let notice = new MatrixAction("notice",
+                `${err}\n`+
+               `Usage: ${fmt}` );
+            yield this.ircBridge.sendMatrixAction(adminRoom, botUser, notice, req);
+            return;
+        }
     }
 });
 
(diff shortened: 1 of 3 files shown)


lib/bridge/MatrixHandler.js
```diff
@@ -369,7 +369,101 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
...
+                    sendArgs.unshift(keyword);
+
+                    let bridgedClient = yield this.ircBridge
+                        .getBridgedClient(server, event.user_id);
+
+                    if (!bridgedClient.unsafeClient) {
+                        throw new Error('Possibly disconnected');
+                    }
+
+                    Client.prototype.send.apply(bridgedClient.unsafeClient, sendArgs);
```
Kegsay: uhhh what? Why not just `bridgedClient.unsafeClient.send.apply(bridgedClient.unsafeClient, sendArgs)`?

lib/bridge/MatrixHandler.js
```diff
@@ -369,7 +369,101 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
...
+            // ...
+
+            if (!matchBlocks) {
+                throw new Error(`Badly formed command block(s)`);
+            }
+
+            for (let i = 0; i < matchBlocks.length; i++) {
+                let block = matchBlocks[i];
+
+                let ircDomain = block.match(/^(?:\r?\n)?([a-z0-9\.:-]+)/)[1];
```
Kegsay: This looks to be the same as the regex in `matchBlocks` for the first line. Likewise on `:423`. I would try to keep this simple. Regular expressions are hard to read, and what you're doing doesn't require complicated matches. Why not just something like:
```javascript
var lines = commands.replace("\r", "").split("\n"); // remove \r and split on the remaining \n's
var currentDomain = null;
lines.forEach((line) => {
  // regex for matching server
  // else regex for matching command
  // else throw
});
```

ara4n:
what is the use case for passing through JOIN and PART here?

Kegsay:
It's any command, not just JOIN, PART.

ara4n:
ah, sorry, was going off the original PR description

lukebarnard1:
(PR desc. updated now!)

lib/bridge/MatrixHandler.js
```diff
@@ -369,7 +368,90 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
...
+        // (COMMAND [arg0 [arg1 [...]]])
+        // irc.server2
+        // ...
+        let commandBlocks = event.content.body;
+
+        let currentServer = ircServer;
+        let blacklist = ['PROTOCTL'];
+
+        try {
+            let lines = commandBlocks.replace("\r").split("\n");
```
Kegsay:
```javascript
> s = "foodfoodfood"
'foodfoodfood'
> s.replace("f")
'undefinedoodfoodfood'
```

Should be:
```javascript
> s = "foodfoodfood"
'foodfoodfood'
> s.replace(/f/g, "")
'oodoodood'
```

lib/bridge/MatrixHandler.js
```diff
@@ -369,7 +368,90 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
...
+                    if (blacklist.indexOf(keyword) != -1) {
+                        throw new Error(`Command blacklisted (${keyword})`);
+                    }
+
+                    let argsJoined = commandMatch[2];
+                    let sendArgs = [];
+
+                    // Arguments are optional
+                    if (!argsJoined) {
+                        argsJoined = '';
```
Kegsay: You never actually use `argsJoined` again after this, so why bother setting this to the empty string? Just do:

```javascript
let argsJoined = commandMatch[2];
let sendArgs = [];
if (argsJoined) {
  sendArgs = argsJoined.split(' ');
}
```

lib/bridge/MatrixHandler.js
```diff
@@ -369,7 +368,90 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
...
+                    }
+                    else {
+                        sendArgs = argsJoined.split(' ');
+                    }
+
+                    sendArgs.unshift(keyword);
+
+                    let bridgedClient = yield this.ircBridge
+                        .getBridgedClient(currentServer, event.user_id);
```
Kegsay: Style: Please break on `{` or `(`, not `.` - e.g:

```javascript
let bridgedClient = yield this.ircBridge.getBridgedClient(
    currentServer, event.user_id
);
```

lib/bridge/MatrixHandler.js
```diff
@@ -369,7 +368,90 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
...
+                    let server = this.ircBridge.getServer(domainMatch[1]);
+
+                    if (!server) {
+                        throw new Error(`IRC domain not found (${ircDomain})`);
+                    }
+
+                    currentServer = server;
+                }
+                else {
+                    throw new Error(`Please provide a valid domain or command block`);
```
Kegsay: We process things line by line, so it's possible we've half-executed the `content.body` before failing. We should state which line we failed to parse here so the user stands a chance at knowing what was/was not executed.

lib/bridge/MatrixHandler.js
```diff
@@ -369,7 +368,90 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
...
+        // ...
+        let commandBlocks = event.content.body;
+
+        let currentServer = ircServer;
+        let blacklist = ['PROTOCTL'];
+
+        try {
+            let lines = commandBlocks.replace("\r").split("\n");
+            for (var ix = 0; ix < lines.length; ix++) {
+                let line = lines[ix];
```
Kegsay: I would almost certainly call `.trim()` on this so mobile clients don't curse your name as often software keyboards automatically insert spaces after words.

lib/bridge/MatrixHandler.js
```diff
@@ -369,7 +368,90 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
         return;
     }
     else {
-        req.log.info("No valid admin command: %s", event.content.body);
+        let fmt = '\nirc.server\n'+
+                    '(COMMAND [arg0 [arg1 [...]]])\n'+
+                    '(COMMAND [arg0 [arg1 [...]]])\n'+
+                    'irc.server2';
```
Kegsay: You may find using ` easier here since then you can put actual new lines rather than escaping them:

```javascript
> s = `foo
... bar
... tar`
'foo\nbar\ntar'
```

lukebarnard1: I find this has the tendency to insert the indentations within the file as well, am I doing something wrong?

Kegsay: No, it will do that. Fair enough if you would rather not because of it.

lib/bridge/MatrixHandler.js
```diff
@@ -369,7 +368,92 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
...
+        // (COMMAND [arg0 [arg1 [...]]])
+        // irc.server2
+        // ...
+        let commandBlocks = event.content.body;
+
+        let currentServer = ircServer;
+        let blacklist = ['PROTOCTL'];
+
+        try {
+            let lines = commandBlocks.replace(/\r/, "").split("\n");
```
Kegsay: If you looked carefully at my comment, you would have see the little `g` at the end there. That is the [global flag](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/RegExp) - you need it in this case:

```javascript
> s = "foodfoodfood"
'foodfoodfood'
> s.replace(/f/, "")
'oodfoodfood'
> s.replace(/f/g, "")
'oodoodood'
```

lukebarnard1: Sorry, my bad!

spec/integ/admin-rooms.spec.js
```diff
@@ -543,4 +543,113 @@ describe("Admin rooms", function() {
...
+        var cmdCount = 0;
+        env.ircMock._whenClient(roomMapping.server, userIdNick, "send",
+        function(client) {
+            cmdCount++;
+        });
+
+        // 3 commands should be executed
+        let command = `JOIN ${newChannel}\n` +
+                        `TOPIC ${newChannel} :some new fancy topic\n` +
+                        `PART ${newChannel}`;
```
Kegsay: Ordering here is really important, so I would verify the ordering on `:589` rather than just incrementing the count.

spec/integ/admin-rooms.spec.js
```diff
@@ -543,4 +543,113 @@ describe("Admin rooms", function() {
...
+        // rubbishserver should throw and prevent more commands
+        let command = `${roomMapping.server}\n` +
+                        `JOIN ${newChannel}\n` +
+                        `TOPIC ${newChannel} :some new fancy topic\n` +
+                        `PART ${newChannel}\n` +
+                        `STUPID COMMANDS\n` +
+                        `${roomMapping.server}\n` +
+                        `SOME COMMAND\n` +
+                        `rubbishserver\n` +
+                        `PART ${newChannel}`;
```
Kegsay: You should strip out the \n and do it as a big \` block (it'll test that `trim()` function is working *and* look nicer :smile:)

Kegsay:
LGTM aside from comments!

PR #127, opened 2016-08-08
Better messages displayed with pastebin actions

lukebarnard1:
When a file is successfully uploaded in light of a text that is too long, the IRC side is shown 'Posted a long message' instead of 'Posted a File' and when the upload fails, the truncated text is sent through with '...(truncated)' on the end.


Kegsay:
LGTM

PR #128, opened 2016-08-08
Provisioning

lukebarnard1:
- Fixed cross-origin problems with provisioning endpoints
- ```/listlinks``` GET endpoint added for retrieving a list of currently provisioned mappings


lib/provisioning/Provisioner.js
```diff
@@ -12,9 +12,26 @@ function Provisioner(ircBridge) {
     let as = this._ircBridge.getAppServiceBridge().appService;
     let self = this;
 
+    if (!(as.app.use && as.app.get && as.app.post)) {
+        log.warn('Could not start provisioning API');
+        return;
```
Kegsay: I wonder if we should be continuing if this is the case. I guess it depends on whether the admin wants to run the provisioning API. If they do (is this in the config?), then we should definitely hard-fail. If it isn't, then failing like this feels fine to me.

lukebarnard1: It's not in the config, but it probably should be. This also begs the question, should the POST and GET endpoints not return something useful when provisioning hasn't been turned on? Something like, "Provisioning hasn't been turned on..."

lib/provisioning/Provisioner.js
```diff
@@ -12,9 +12,26 @@ function Provisioner(ircBridge) {
...
+    }
+
+    as.app.use(function(req, res, next) {
+        if (req.url === '/_matrix/provision/unlink' ||
+            req.url === '/_matrix/provision/link'||
+            req.url.match(/^\/_matrix\/provision\/listlinks/)) {
```
Kegsay: Double-indent please!

lib/provisioning/Provisioner.js
Kegsay: Good idea! Yeah, that would make sense.

lib/DataStore.js
```diff
@@ -120,6 +120,21 @@ DataStore.prototype.getRoom = function(roomId, ircDomain, ircChannel, origin) {
 };
 
 /**
+ * Get provisioned IRC <--> Matrix room mappings from the database where
+ * the matrix room ID is roomId.
+ * @param {string} roomId : The Matrix room ID.
+ * @return {Promise} A promise which resolves to a list
+ * of mappings.
```
Kegsay: List of entries surely?

lib/provisioning/Provisioner.js
```diff
@@ -46,36 +72,39 @@ var parameterValidation = {
...
-    for (var i = 0; i < parameterNames.length; i++) {
-        let name = parameterNames[i];
-
-        let actual = parameters[name];
-        let valid = parameterValidation[name];
+Provisioner.prototype._validate = function(actual, parameterName) {
+    let valid = parameterValidation[parameterName];
+    if (!actual) {
+        throw new Error(
+            `${parameterName} not provided (like '${valid.example}').`
```
Kegsay: Careful! What if `parameterName` is not one of the three you expect? This will `TypeError` rather than give you this error because of `valid.example`.

Kegsay: I would null check `valid` before this point then return something like "bad parameter name"

Kegsay:
LGTM aside from comments (please just merge once you've addressed them; no need for me to look at this again).

Issue #130, opened 2016-08-10
TypeError: (("IRC domain not found (" + (intermediate value)) + ")") is not a function


rubo77:
since two hours, we get this error message in vector, every time someone sais anything either in IRC or in matrix.
 the error message is always visible in matrix and sometimes in irc too:
```
@appservice-irc:matrix.org
TypeError: (("IRC domain not found (" + (intermediate value)) + ")") is not a function
Usage: 
irc.server
(COMMAND [arg0 [arg1 [...]]])
(COMMAND [arg0 [arg1 [...]]])
irc.server2
```

how can we fix this?

lukebarnard1:
Just to clarify, are you on the v0.4.0 branch?

rubo77:
I now checked out your branch `luke/hotfix0.4.0-1` on my server, but still this message, an I found out, that the error comes from the global `@appservice-irc:matrix.org`

should I block that user then? (There is no way of unblocking it in the current master)

rubo77:
You would need a quick fix on your main server, cause that spamming could be a big problem if it occurs on different servers.

The reason for the bug could have something to do, with DNS updates on our side for `service.ffnord.net` or Letsencrypt updates on our side on eclabs.de?

lukebarnard1:
The error text comes from the handler for raw IRC commands being sent from the matrix admin room, so it's quite strange that the message you're seeing would be printed in the bridged channel. Also, given that this new feature hasn't been merged into master yet, this suggests that the error wasn't coming from your bridge (and it looks like that it's the matrix bridge doing it somehow).

I shall investigate further this morning 👍🏻

rubo77:
The error message was also visible in irc, but only sometimes. I blocked your user now and our bridge works fine

Kegsay:
@rubo77 - Why did you invite that user into your room in the first place?

rubo77:
maybe the search for my own appbot suggested that bot. I guess this can happen some times if thousand users set up their own appbot.

nice, that you fixed it in 132

PR #131, opened 2016-08-10
Quick fix for #130

lukebarnard1:
This fix will make sure that the error is printed correctly, i.e:
```
IRC domain not found (somedomain) on line [some line number]
```

But it remains to be seend why @appservice-irc is sending out spam to _any_ matrix room it joins:
```
TypeError: (("IRC domain not found (" + (intermediate value)) + ")") is not a function
Usage: 
irc.server
(COMMAND [arg0 [arg1 [...]]])
(COMMAND [arg0 [arg1 [...]]])
irc.server2
```


Kegsay:
LGTM

PR #132, opened 2016-08-11
IRC bot will leave matrix room with > 2 members

lukebarnard1:
As seen with issue https://github.com/matrix-org/matrix-appservice-irc/issues/130, bots that are mistakenly invited to group chats will try to respond to every message sent (and prior to the new raw IRC commands feature (PR https://github.com/matrix-org/matrix-appservice-irc/pull/125), messages in the 'admin room' were ignored silently).

To deal with this, the bot will now leave a room if it determines that more than 2 members are in the Matrix room.

### DIFF
--- a/lib/bridge/MatrixHandler.js
+++ b/lib/bridge/MatrixHandler.js
@@ -152,7 +152,6 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
                 return m.content.membership && m.content.membership === "join";
             }
         );
-
     }
     else {
         req.log.warn('Member tracker not running');
@@ -502,11 +501,11 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
 /**
  * Called when the AS receives a new Matrix invite/join/leave event.
  * @param {Object} event : The Matrix member event.
- * @param {MatrixUser} inviter : The inviter (sender).
- * @param {MatrixUser} invitee : The invitee (receiver).
+ * @param {MatrixUser} sender : The sender.
+ * @param {MatrixUser} receiver : The receiver.
  * @return {Promise} which is resolved/rejected when the request finishes.
  */
-MatrixHandler.prototype._onMemberEvent = Promise.coroutine(function*(req, event, inviter, invitee) {
+MatrixHandler.prototype._onMemberEvent = Promise.coroutine(function*(req, event, sender, receiver) {
     if (!this._memberTracker) {
         let matrixClient = this.ircBridge.getAppServiceBridge().getClientFactory().getClientAs();
 
@@ -517,8 +516,6 @@ MatrixHandler.prototype._onMemberEvent = Promise.coroutine(function*(req, event,
     }
     else {
         this._memberTracker.onEvent(event);
-
-        yield Promise.resolve();
     }
 });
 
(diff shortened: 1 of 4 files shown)


lib/bridge/MatrixHandler.js
```diff
@@ -129,12 +130,47 @@ MatrixHandler.prototype._handleInviteFromUser = Promise.coroutine(function*(req,
...
+    // If an admin room has more than 2 people in it, kick the bot out
+
+    let matrixClient = this.ircBridge.getAppServiceBridge().getClientFactory().getClientAs();
+
+    let lookup = new StateLookup({
+        client : matrixClient,
+        eventTypes: ['m.room.member']
+    });
+
+    yield lookup.trackRoom(adminRoom.getId());
```
Kegsay: This method will make a `/state` HTTP request the first time this is called on an object. This can be a costly operation. It is being done for each and every admin message that is received. We cannot avoid doing at least one `/state` request per room, but we can make subsequent messages faster by remembering what the previous results were and feeding in updates as and when we receive them. That is the purpose of `onEvent`, which you should use to make this a bit nicer.

lib/bridge/MatrixHandler.js
```diff
@@ -129,12 +130,47 @@ MatrixHandler.prototype._handleInviteFromUser = Promise.coroutine(function*(req,
...
+    let members = lookup.getState(
+        adminRoom.getId(),
+        'm.room.member'
+    ).filter(
+        function (m) {
+            return m.membership && m.membership !== "leave";
```
Kegsay: Better to check for `=== "join"` as there is also a `ban` and `invite` state.

lib/bridge/MatrixHandler.js
```diff
@@ -129,12 +130,47 @@ MatrixHandler.prototype._handleInviteFromUser = Promise.coroutine(function*(req,
...
+        }
+    );
+
+    if (members.length > 2) {
+        req.log.error('_onAdminMessage: admin room has too many users; bot will leave');
```
Kegsay: Probably worth logging how many users

lib/bridge/MatrixHandler.js
```diff
@@ -458,6 +500,29 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
...
+
+        this._memberTracker = new StateLookup({
+            client : matrixClient,
+            eventTypes: ['m.room.member']
+        });
+    }
+    else {
+        this._memberTracker.onEvent(event);
+
+        yield Promise.resolve();
```
Kegsay: You don't need this because by default it will resolve the promise when the function ends.

lib/bridge/MatrixHandler.js
```diff
@@ -458,6 +500,29 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
 });
 
 /**
+ * Called when the AS receives a new Matrix invite/join/leave event.
+ * @param {Object} event : The Matrix member event.
+ * @param {MatrixUser} inviter : The inviter (sender).
+ * @param {MatrixUser} invitee : The invitee (receiver).
```
Kegsay: s/inviter/sender/
s/invitee/receiver/

Kegsay:
Test please and then LGTM!

PR #133, opened 2016-08-12
Changed RAW IRC command format

lukebarnard1:
To prevent commands being issued by mistake and to be less verbose in terms of errors
occuring, the command format and handling has changed. Commands issued now have this
form (which matches the previously available ones) :
```
!cmd [server.com] COMMAND [arg1 [arg2 [...]]
```
The usage message will now only be shown when !help is used, and not when an error
occurs.


Kegsay:
LGTM

PR #138, opened 2016-08-16
explanation for url corrected and port removed

rubo77:
The option `url` must represent the server url from external IRC channels (without port)

### DIFF
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -1,8 +1,8 @@
 # Configuration specific to AS registration. Unless other marked, all fields
 # are *REQUIRED*.
 homeserver:
-  # The URL to the home server for client-server API calls.
-  url: "http://localhost:8008"
+  # The URL to posted files on your home server as they are displayed in links in the bridged IRC channels:
+  url: "http://localhost"
   # The 'domain' part for user IDs on this home server. Usually (but not always)
   # is the "domain name" part of the HS URL.
   domain: "localhost"


rubo77:
@lukeb sais: "the port number is needed as part of the home server URL"

so maybe there is a problem here? and we need two config values? one for the internal url and one for postings in IRC?

lukeb:
I don't think you meant to tag me. Perhaps another Luke?

lukebarnard1:
Here! Sorry - my ID on GH is lukebarnard1

lukebarnard1:
For now, we're not supporting media servers with different URLs to the CS API. We may support this in the future. Also, the URL in the config.yaml _is_ used for CS API calls, so this PR is incorrect.

What we suggest is that you make a new PR which changes the comment in the config.yaml to say that the URL is _also_ used to form the media URLs (so that people are better informed).

rubo77:
I don't understand, what this has to do with "supporting media servers with different URLs to the CS API"

we only have one server and only bridge one channel to one ircnet (hackint). there we had the problem, that the lilnks to images had the wrong hostname in IRC.

How do you solve this?

Internal the port is supposed to be needed (although my server works without the port added here) and for klickable links in the bridged IRC channels, the url must be without port.

lukebarnard1:
The only conceivable scenarios that could result in your experience is either

1. You have two HSs, one running on port 80 and one on 8008. The bridge is registered with both and thus changing the HS URL in the bridge config.yaml is causing the links to appear differently in the IRC "posted a file message".
2. You have one HS but it is somehow configured to serve media from port 80 and do CS on port 8008.

With a typical setup, the HS will use the same URL for both the media serving and the CS API. The IRC bridge does not support (for now) the ability to use a different URL for media serving, so we suggest that for now, to make things clearer that you make a new PR which explains that the URL in the config is also used to form those post file messages.

rubo77:
I created a new PR here:
https://github.com/matrix-org/matrix-appservice-irc/pull/140

So lets discuss our problem here:

first: what stands "CS" for?

We only have on HS running on matrix.eclabs.de, how could we have defined two different ports? 80 and 8008? where is this configured?

rubo77:
I found out, that if I change the URL to just the domain without port, the bridge only works in one direction, but not from IRC->Matrix. so I changed it back to 

    url: server.org:8008

So how do I solve this now?

rubo77:
https://github.com/matrix-org/matrix-appservice-irc/issues/154

PR #140, opened 2016-08-17
explanation for url corrected

rubo77:
The option `url` also represents the server url from external IRC channels


Kegsay:
LGTM, thanks!

rubo77:
This will be better:

https://github.com/matrix-org/matrix-appservice-irc/issues/154

Issue #119, opened 2016-08-01
Freenode bridge got k-lined due to lack of throttling on reconnects


Kegsay:
```
{"command":"ERROR","rawCommand":"ERROR","commandType":"normal","args":["*** Banned: Banned due to too many connections in a short period, email kline@freenode.net when corrected. (2016/8/1 15.55)"]}
```

Due to a bunch of ping timeouts :/

Issue #134, opened 2016-08-12
Config idleTimeout is in seconds, whilst IrcServer exposes 'getIdleTimeoutMs' 


lukebarnard1:
IrcServer.js:~67
```js
IrcServer.prototype.getIdleTimeoutMs = function() {
    return this.config.ircClients.idleTimeout;
};
```
config.sample.yaml:~208 says seconds (and the code expects seconds):
```yaml
 # The maximum amount of time in seconds that the client can exist
        # without sending another message before being disconnected. Use 0 to
        # not apply an idle timeout. This value is ignored if this IRC server is
        # mirroring matrix membership lists to IRC. Default: 172800 (48 hours)
        idleTimeout: 10800
```

lukebarnard1:
Fixed
https://github.com/matrix-org/matrix-appservice-irc/commit/ae34cc1c7cb0b0639aae238336fda513772bac64

PR #141, opened 2016-08-17
Scheduled IRC Reconnects

lukebarnard1:
To prevent the IRC bridge from being k-lined in light of too many reconnections (https://github.com/matrix-org/matrix-appservice-irc/issues/119), reconnections are scheduled on a per-server basis. Only one reconnect can be currently executed per IRC server.

### DIFF
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -208,7 +208,7 @@ ircService:
         idleTimeout: 10800
         # The number of millseconds to wait between consecutive reconnections if a
         # client gets disconnected. Default: 5000 (5 seconds)
-        reconnectInterval: 5000
+        reconnectIntervalMs: 5000
         # The number of lines to allow being sent by the IRC client that has received
         # a large block of text to send from matrix. If the number of lines that would
         # be sent is > lineLimit, the text will instead be uploaded to matrix and the
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -208,12 +208,10 @@ ircService:
         idleTimeout: 10800
         # The number of millseconds to wait between consecutive reconnections if a
         # client gets disconnected. Setting to 0 will cause the scheduling to be
-        # disabled, i.e. it will be scheduled immediately. Otherwise, the scheduling
-        # interval will be used such that one client reconnect for this server will
-        # be handled every reconnectIntervalMs ms using a FIFO queue. Regardless of
-        # scheduling, a delay will be added to the reconnect process:
-        #   addedDelay = 5000 + (1000 * attempts) * Math.random()
-        #                + 20000 // if throttled
+        # disabled, i.e. it will be scheduled immediately (with jitter.
+        # Otherwise, the scheduling interval will be used such that one client
+        # reconnect for this server will be handled every reconnectIntervalMs ms using
+        # a FIFO queue.
         # Default: 5000 (5 seconds)
         reconnectIntervalMs: 5000
         # The number of lines to allow being sent by the IRC client that has received
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -207,7 +207,14 @@ ircService:
         # mirroring matrix membership lists to IRC. Default: 172800 (48 hours)
         idleTimeout: 10800
         # The number of millseconds to wait between consecutive reconnections if a
-        # client gets disconnected. Default: 5000 (5 seconds)
+        # client gets disconnected. Setting to 0 will cause the scheduling to be
+        # disabled, i.e. it will be scheduled immediately. Otherwise, the scheduling
+        # interval will be used such that one client reconnect for this server will
+        # be handled every reconnectIntervalMs ms using a FIFO queue. Regardless of
+        # scheduling, a delay will be added to the reconnect process:
+        #   addedDelay = 5000 + (1000 * attempts) * Math.random()
+        #                + 20000 // if throttled
+        # Default: 5000 (5 seconds)
         reconnectIntervalMs: 5000
         # The number of lines to allow being sent by the IRC client that has received
         # a large block of text to send from matrix. If the number of lines that would
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -206,6 +206,9 @@ ircService:
         # not apply an idle timeout. This value is ignored if this IRC server is
         # mirroring matrix membership lists to IRC. Default: 172800 (48 hours)
         idleTimeout: 10800
+        # The number of millseconds to wait between consecutive reconnections if a
+        # client gets disconnected. Default: 5000 (5 seconds)
+        reconnectInterval: 5000
         # The number of lines to allow being sent by the IRC client that has received
         # a large block of text to send from matrix. If the number of lines that would
         # be sent is > lineLimit, the text will instead be uploaded to matrix and the
--- a/lib/irc/ClientPool.js
+++ b/lib/irc/ClientPool.js
@@ -8,9 +8,26 @@ var stats = require("../config/stats");
 var log = require("../logging").get("ClientPool");
 var Promise = require("bluebird");
 
-const RECONNECT_TIME_MS = 10000;
-// Wait between recconections of clients
-const RECONNECT_INTERVAL_MS = 1000;
+const RECONNECT_TIME_MS = 1000;
+
+// Throttle
+function throttle (cp, func, delay) {
+    let queue = [];
+    let handleQueue = function () {
+        if (queue.length > 0) {
+            let args = queue.shift();
+            console.log('Doing call!');
+            func.apply(cp, args);
+        }
+    }
+
+    setInterval(handleQueue, delay);
+
+    return function () {
+        console.log('Queueing call!');
+        queue.push(Array.from(arguments));
+    }
+}
 
 function ClientPool(ircBridge) {
     this._ircBridge = ircBridge;
@@ -30,6 +47,8 @@ function ClientPool(ircBridge) {
         //    }
         // }
     };
+
+    this._throttledReconnects = {};
 }
 
 ClientPool.prototype.addBot = function(server, client) {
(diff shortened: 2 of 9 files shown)


config.sample.yaml
```diff
@@ -206,6 +206,9 @@ ircService:
         # not apply an idle timeout. This value is ignored if this IRC server is
         # mirroring matrix membership lists to IRC. Default: 172800 (48 hours)
         idleTimeout: 10800
+        # The number of millseconds to wait between consecutive reconnections if a
+        # client gets disconnected. Default: 5000 (5 seconds)
```
Kegsay: You don't mention how to disable this (setting 0? -1?).

I also thought we decided to have a "base" reconnect time to which this value will be added to it, yet this description makes no mention of this.

lib/irc/ClientPool.js
```diff
@@ -238,25 +238,31 @@ ClientPool.prototype._onClientDisconnected = function(bridgedClient) {
...
-                    cli.joinChannel(c);
-                });
-            }
-            self._sendConnectionMetric(cli.server);
-        }, function(e) {
-            log.error(
-                "<%s> Failed to reconnect %s@%s", cli._id, cli.nick, cli.server.domain
-            );
-        });
+        self._reconnect(cli, chanList);
```
Kegsay: I know this is leftover from when we previously did reconnects from the `ClientPool`, but this just adds needlessly to the diff. Can you revert this file (but keep the change to `RECONNECT_TIME_MS`)?

lib/irc/IrcServer.js
```diff
@@ -68,6 +68,10 @@ IrcServer.prototype.getIdleTimeoutMs = function() {
     return this.config.ircClients.idleTimeout;
 };
 
+IrcServer.prototype.getReconnectInterval = function() {
```
Kegsay: Units please, like with `getIdleTimeoutMs`.

lukebarnard1: Just so you know, I renamed that - https://github.com/matrix-org/matrix-appservice-irc/issues/134 
It was in seconds in the code and also, the code treated it as seconds.

lib/irc/Scheduler.js
```diff
@@ -0,0 +1,107 @@
...
+    finally {
+        this._processing = null;
+    }
+});
+
+ScheduledQueue.prototype._procFn = function (item) {
+    return item.fn();
+}
+
+ScheduledQueue.prototype.enqueue = function(item) {
```
Kegsay: Please take care with your spacing. This line is `function(item) {` whereas `:40` is `function (item) {`. Use the same form we use elsewhere please (no extra space).

lib/irc/Scheduler.js
```diff
@@ -0,0 +1,107 @@
...
+/**
+ * An IRC connection scheduler. Enables ConnectionInstance to reconnect
+ * in a way that queues reconnection requests and services the FIFO queue at a
+ * rate determined by ircServer.getReconnectInterval().
+ */
+
+var Scheduler = {
+    _queueServers: [],
+    _queues: {},
+    _getQueue: _getQueue,
```
Kegsay: I don't like the amount of indirection here. Why not just in-line this, or call the global function directly?

lib/irc/Scheduler.js
```diff
@@ -0,0 +1,107 @@
...
+var Promise = require("bluebird");
+var logging = require("../logging");
+var log = logging.get("scheduler");
+
+function ScheduledQueue (server) {
+    this._queue = [];
+    this._processing = null;
+    this._server = server;
+
+    setInterval(this._consume.bind(this), server.getReconnectInterval());
```
Kegsay: What if `getReconnectInterval` is `0`?

lib/irc/Scheduler.js
```diff
@@ -0,0 +1,107 @@
...
+}
+
+/**
+ * An IRC connection scheduler. Enables ConnectionInstance to reconnect
+ * in a way that queues reconnection requests and services the FIFO queue at a
+ * rate determined by ircServer.getReconnectInterval().
+ */
+
+var Scheduler = {
+    _queueServers: [],
```
Kegsay: We are exporting the entire `Scheduler` object. We have a bunch of private stuff that we don't want people to mess with (like this one). Why are we putting this with the exported object? Why not at the top-level (in which case it is entirely enclosed in a closure and physically cannot be accessed from an external file)?

lib/irc/Scheduler.js
```diff
@@ -0,0 +1,107 @@
...
+        q.enqueue({defer: d, fn: retryConnection});
+
+        return d.promise;
+    }),
+    killAll: killAll
+};
+
+function _newQueue (server) {
+    Scheduler._queues[server.domain] = new ScheduledQueue(server);
+    Scheduler._queueServers.push(server.domain);
```
Kegsay: It looks like the only purpose of `_queueServers` is to maintain a list of keys into `_queues`. We don't need to do this manually. `Object.keys(Schedule._queues)` will do that for you.

lib/irc/Scheduler.js
```diff
@@ -0,0 +1,107 @@
...
+ * in a way that queues reconnection requests and services the FIFO queue at a
+ * rate determined by ircServer.getReconnectInterval().
+ */
+
+var Scheduler = {
+    _queueServers: [],
+    _queues: {},
+    _getQueue: _getQueue,
+    _newQueue: _newQueue,
+    reschedule: Promise.coroutine(function*(server, addedDelay, retryConnection) {
```
Kegsay: Please add some comments to explain when this promise will be resolved/rejected.

lib/irc/ConnectionInstance.js
```diff
@@ -256,7 +257,7 @@ ConnectionInstance.create = function(server, opts, onCreatedCallback) {
 
     var connAttempts = 0;
     function retryForever() {
-        retryConnection().then(returnClient, function(err) {
+        let reschedule = function(err) {
```
Kegsay: Please call this something different. It's really confusing that this is called `reschedule` AND there is `Scheduler.reschedule` AND we're passing functions around.

lib/irc/Scheduler.js
```diff
@@ -0,0 +1,107 @@
...
+    Scheduler._queueServers.push(server.domain);
+
+    return Scheduler._queues[server.domain];
+}
+
+function _getQueue (server) {
+    let q = Scheduler._queues[server.domain];
+
+    if (!q) {
+        q = Scheduler._newQueue(server);
```
Kegsay: Please just inline this. What `_getQueue` is doing is a very common pattern (lazy initialisation) and `_newQueue` is only 3 lines and is called in only one place (here).

lib/irc/Scheduler.js
```diff
@@ -0,0 +1,107 @@
+/*eslint no-invalid-this: 0*/
+"use strict"
+var promiseutil = require("../promiseutil");
+var Promise = require("bluebird");
+var logging = require("../logging");
+var log = logging.get("scheduler");
+
+function ScheduledQueue (server) {
```
Kegsay: Overall this looks dangerously close to being the same as `Queue` in the util file, but with a set consume interval rather than consume-as-fast-as-possible. It would be better if you just added the extra config option (consume-at-intervals) to the `Queue` class rather than redo what the `Queue` is doing in the first place.

Kegsay: Oh and with a `killAll()` method, which can trivially be added to the existing `Queue`.

lib/irc/IrcServer.js
Kegsay: This is exactly why units are important :D Ideally the config file would've had that from the get-go rather than just `idleTimeout`.

lib/irc/Scheduler.js
lukebarnard1: But the important difference is with ```_consume``` using ```setTimeout```. If I use ```Queue```, we can't do any extra delay in the critical section (line 27) and we can't do a ```setTimeout``` to retry after the reconnection interval. I could inherit from ```Queue```, but I'd have to overload the consumption method anyway.

lukebarnard1: I totally didn't consider adding an option to ```Queue```, that makes a lot of sense.

lib/irc/ConnectionInstance.js
```diff
@@ -255,20 +263,43 @@ ConnectionInstance.create = function(server, opts, onCreatedCallback) {
...
+                Scheduler.reschedule(server, retryTimeMs, retryConnection, opts.nick).then(
+                    returnClient,
+                    retryFailed
+                );
+            }
+            else {
+                let delay = BASE_RETRY_TIME_MS + retryTimeMs +
+                           Math.round((connAttempts * 1000) * Math.random());
+
+                log.info(`Scheduling disabled, retrying with delay ${delay}ms`);
```
Kegsay: This is going to be hella spammy if scheduling is disabled. I wouldn't mind if we actually logged something more useful aside from the amount of time delay (e.g the nick and server being connected, the number of `connAttempts`)

lib/irc/Scheduler.js
```diff
@@ -0,0 +1,64 @@
...
+        var promise = q.enqueue(
+            `Scheduler.reschedule ${server.domain} ${nick}`,
+            {
+                fn: retryConnection,
+                addedDelayMs: addedDelayMs
+            }
+        );
+
+        log.info(
+            `Tryed to queue scheduled promise for ${server.domain} ${nick}` +
```
Kegsay: image

By this point you have already queued it, you aren't trying to do so.

lib/irc/Scheduler.js
```diff
@@ -0,0 +1,64 @@
...
+            (addedDelayMs > 0 ? ` with ${Math.round(addedDelayMs)}ms added delay`:'')
+        );
+
+        return promise;
+    }),
+    killAll: killAll
+};
+
+// Reject all queued promises
+function killAll() {
```
Kegsay: This function is lonely and wants to be with `Scheduler.reschedule` :) please in-line rather than indirect like this.

lib/irc/Scheduler.js
```diff
@@ -0,0 +1,64 @@
...
+var log = logging.get("scheduler");
+var Queue = require("../util/Queue.js");
+
+/**
+ * An IRC connection scheduler. Enables ConnectionInstance to reconnect
+ * in a way that queues reconnection requests and services the FIFO queue at a
+ * rate determined by ircServer.getReconnectIntervalMs().
+ */
+
+ var queues = {};
```
Kegsay: In-line comment explaining what this is would be super. e.g `// domain => Queue`

lib/util/Queue.js
```diff
@@ -9,11 +9,19 @@ var promiseutil = require("../promiseutil");
  * is in its critical section. Only 1 item at any one time will be calling this function.
  * The function should return a Promise which is resolved/rejected when the next item
  * can be taken from the queue.
+ * @param {integer} intervalMs Optional. If provided and > 0, this queue will be serviced
+ * at an interval of intervalMs. Otherwise, items will be processed imediately.
```
Kegsay: "immediately" is a bit of a white lie, since the entire point of the `Queue` is to ensure that no 2 items are processed concurrently. If they were processed "immediately" (that is, as soon as `enqueue` is called) then they will be concurrent and that's Bad. Bad bad bad.

lib/util/Queue.js
```diff
@@ -9,11 +9,19 @@ var promiseutil = require("../promiseutil");
...
  */
-function Queue(processFn) {
+function Queue(processFn, intervalMs) {
     this._queue = [];
     this._processing = null;
     this._procFn = processFn; // critical section Promise<result> = fn(item)
+    this._intervalMs = intervalMs;
+
+    if (this._intervalMs) {
```
Kegsay: You document what you expect this to be (an integer which will queue if > 0), but you don't check it. This is javascript remember: the language which will attempt to back stab you in a heartbeat:
```javascript
> s = -4
-4
> if (s) { console.log("HERRO") }
HERRO
undefined
> s = 0
0
> if (s) { console.log("HERRO") }
undefined
```

Needs more paranoia, it's out to get yoooooou! :imp:

lib/util/Queue.js
```diff
@@ -37,19 +45,28 @@ Queue.prototype.enqueue = function(id, thing) {
         item: thing,
         defer: defer
     });
-    // always process stuff asyncly, never syncly.
-    process.nextTick(() => {
-        this._consume();
-    });
+    if (!this._intervalMs) {
```
Kegsay: For the same reason as https://github.com/matrix-org/matrix-appservice-irc/pull/141/files#r76762550 - you cannot rely on truthiness of integers if you aren't asserting for negative numbers.

Kegsay:
I assume this works? 0:) Broadly speaking, it looks like it should work. It probably works, right?

lib/irc/Scheduler.js
lukebarnard1: The thing is, it may have already queued it and there is 0 indication as to whether it was actually queued or if it was already queued...

lukebarnard1:
I can confirm that I've tested it manually on my own locally setup many times. So yes, it probably works :white_check_mark:

Kegsay:
LGTM!

PR #143, opened 2016-08-19
Ask IRC chanop for Bridging

lukebarnard1:
Link provisioning now requires that the nick of a valid IRC operator is given with a requested link. The bridge sends a message to the op asking to respond 'yes' if they would like to authorise a given
bridging.

Determining whether someone is an op uses whois, which required tweaking on the ```BridgedClient``` to use the operator field that was so far unused.

```m.room.bridging``` room state is used in the room in the new mapping to signal whether the status of the bridging is 'pending', 'success', or 'failure'.

 ```JS
{
    content: {
        bridger: 'nick',
        status: 'pending'
    }
}

### DIFF
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -286,3 +286,8 @@ ircService:
   provisioning:
     # True to enable the provisioning HTTP endpoint. Default: true.
     enabled: true
+    # Number of seconds to wait before giving up on getting a response from
+    # an IRC operator. If the operator does not respond within the alloted
+    # time period, the provisioning request will fail.
+    # Default: 300 seconds (5 mins)
+    requestTimeoutSeconds: 300
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -286,8 +286,8 @@ ircService:
   provisioning:
     # True to enable the provisioning HTTP endpoint. Default: true.
     enabled: true
-    # Number of seconds to wait before giving up on getting a response from
-    # an IRC operator. If the operator does not respond within the alloted
-    # time period, the provisioning request will fail.
+    # The number of seconds to wait before giving up on getting a response from
+    # an IRC channel operator. If the channel operator does not respond within the
+    # allotted time period, the provisioning request will fail.
     # Default: 300 seconds (5 mins)
     requestTimeoutSeconds: 300
--- a/lib/irc/BridgedClient.js
+++ b/lib/irc/BridgedClient.js
@@ -339,7 +339,7 @@ BridgedClient.prototype.whois = function(nick) {
  * @param {string} channel : The channel to call /names on
  */
 BridgedClient.prototype.getOperators = function(channel) {
-    return this.getNicks(channel).then(function (nicksInfo) {
+    return this.getNicks(channel).then(function(nicksInfo) {
         let nicks = nicksInfo.nicks;
 
         nicksInfo.operatorNicks = nicks.filter(
--- a/lib/irc/BridgedClient.js
+++ b/lib/irc/BridgedClient.js
@@ -339,27 +339,39 @@ BridgedClient.prototype.whois = function(nick) {
  * @param {string} channel : The channel to call /names on
  */
 BridgedClient.prototype.getOperators = function(channel) {
+    let modify = function (nicksInfo) {
+        let nicks = nicksInfo.nicks;
+
+        nicksInfo.operatorNicks = nicks.filter(
+            (index) => {
+                return nicksInfo.names[index] === '@'
+            }
+        );
+        return nicksInfo;
+    }
+    return this.getNicks(channel).then(modify);
+};
+
+/**
+ * Get the nicks of the users in a channel
+ * @param {string} channel : The channel to call /names on
+ */
+BridgedClient.prototype.getNicks = function(channel) {
     if (this.disabled) {
         return Promise.resolve({
             server: this.server,
             channel: channel,
-            operatorNicks: []
+            names: {}
         });
     }
     var self = this;
     return new Promise(function(resolve, reject) {
         self.unsafeClient.names(channel, function(channelName, names) {
-
-            let operatorNicks = Object.keys(names).filter(
-                (index) => {
-                    return names[index] === '@'
-                }
-            );
-
             resolve({
                 server: self.server,
                 channel: channelName,
-                operatorNicks: operatorNicks
+                nicks: Object.keys(names),
+                names: names,
             });
         });
     }).timeout(5000);
--- a/lib/irc/BridgedClient.js
+++ b/lib/irc/BridgedClient.js
@@ -333,29 +333,6 @@ BridgedClient.prototype.whois = function(nick) {
     });
 };
 
-/**
- * Get the operator status of an IRC user
- * @param {string} mask : The mask to call /who on
- */
-BridgedClient.prototype.who = function(mask) {
-    if (this.disabled) {
-        return Promise.resolve({
-            server: this.server,
-            mask: mask
-        });
-    }
-    var self = this;
-    return new Promise(function(resolve, reject) {
-        self.unsafeClient.who(mask, function(who) {
-            resolve({
-                server: self.server,
-                mask: mask,
-                who: who
-            });
-        });
-    });
-};
-
 
 /**
  * Get the operators of a channel
--- a/lib/irc/BridgedClient.js
+++ b/lib/irc/BridgedClient.js
@@ -327,6 +327,7 @@ BridgedClient.prototype.whois = function(nick) {
             resolve({
                 server: self.server,
                 nick: nick,
+                isOp: whois.operator !== undefined,
                 msg: `Whois info for '${nick}': ${info}`
             });
         });
--- a/lib/irc/BridgedClient.js
+++ b/lib/irc/BridgedClient.js
@@ -343,8 +343,8 @@ BridgedClient.prototype.getOperators = function(channel) {
         let nicks = nicksInfo.nicks;
 
         nicksInfo.operatorNicks = nicks.filter(
-            (index) => {
-                return nicksInfo.names[index] === '@'
+            (nick) => {
+                return nicksInfo.names[nick] === '@'
             }
         );
         return nicksInfo;
--- a/lib/irc/BridgedClient.js
+++ b/lib/irc/BridgedClient.js
@@ -327,12 +327,67 @@ BridgedClient.prototype.whois = function(nick) {
             resolve({
                 server: self.server,
                 nick: nick,
-                isOp: whois.operator !== undefined,
                 msg: `Whois info for '${nick}': ${info}`
             });
         });
     });
 };
+
+/**
+ * Get the operator status of an IRC user
+ * @param {string} mask : The mask to call /who on
+ */
+BridgedClient.prototype.who = function(mask) {
+    if (this.disabled) {
+        return Promise.resolve({
+            server: this.server,
+            mask: mask
+        });
+    }
+    var self = this;
+    return new Promise(function(resolve, reject) {
+        self.unsafeClient.who(mask, function(who) {
+            resolve({
+                server: self.server,
+                mask: mask,
+                who: who
+            });
+        });
+    });
+};
+
+
+/**
+ * Get the operators of a channel
+ * @param {string} channel : The channel to call /names on
+ */
+BridgedClient.prototype.getOperators = function(channel) {
+    if (this.disabled) {
+        return Promise.resolve({
+            server: this.server,
+            channel: channel,
+            operatorNicks: []
+        });
+    }
+    var self = this;
+    return new Promise(function(resolve, reject) {
+        self.unsafeClient.names(channel, function(channelName, names) {
+
+            let operatorNicks = Object.keys(names).filter(
+                (index) => {
+                    return names[index] === '@'
+                }
+            );
+
+            resolve({
+                server: self.server,
+                channel: channelName,
+                operatorNicks: operatorNicks
+            });
+        });
+    });
+};
+
 
 /**
  * Convert the given nick into a valid nick. This involves length and character
--- a/lib/irc/BridgedClient.js
+++ b/lib/irc/BridgedClient.js
@@ -339,7 +339,7 @@ BridgedClient.prototype.whois = function(nick) {
  * @param {string} channel : The channel to call /names on
  */
 BridgedClient.prototype.getOperators = function(channel) {
-    let modify = function (nicksInfo) {
+    return this.getNicks(channel).then(function (nicksInfo) {
         let nicks = nicksInfo.nicks;
 
         nicksInfo.operatorNicks = nicks.filter(
@@ -348,8 +348,7 @@ BridgedClient.prototype.getOperators = function(channel) {
             }
         );
         return nicksInfo;
-    }
-    return this.getNicks(channel).then(modify);
+    });
 };
 
 /**
--- a/lib/irc/BridgedClient.js
+++ b/lib/irc/BridgedClient.js
@@ -362,7 +362,7 @@ BridgedClient.prototype.getOperators = function(channel) {
                 operatorNicks: operatorNicks
             });
         });
-    });
+    }).timeout(5000);
 };
 
 
--- a/lib/irc/BridgedClient.js
+++ b/lib/irc/BridgedClient.js
@@ -367,6 +367,8 @@ BridgedClient.prototype.getNicks = function(channel) {
     var self = this;
     return new Promise(function(resolve, reject) {
         self.unsafeClient.names(channel, function(channelName, names) {
+            // names maps nicks to chan op status, where '@' indicates chan op
+            // names = {'nick1' : '', 'nick2' : '@', ...}
             resolve({
                 server: self.server,
                 channel: channelName,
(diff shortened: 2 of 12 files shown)

### CODE REVIEW

**Comment by lukebarnard1** [comment]
Tests will follow :+1:

File: lib/irc/BridgedClient.js [hunk 1]
```diff
@@ -327,6 +327,7 @@ BridgedClient.prototype.whois = function(nick) {
             resolve({
                 server: self.server,
                 nick: nick,
+                isOp: whois.operator !== undefined,
```
Kegsay: I *think* this is set from `RPL_WHOISOPERATOR` which is the response command to `WHOIS`. This is whether they are an IRC operator rather than a chan op, which is what we really want to be checking here.

lib/provisioning/Provisioner.js
```diff
@@ -2,13 +2,17 @@
...
 var MatrixRoom = require("matrix-appservice-bridge").MatrixRoom;
+var StateLookup = require('matrix-appservice-bridge').StateLookup;
 
 var log = require("../logging").get("Provisioner");
+var promiseutil = require("../promiseutil.js");
 
 function Provisioner(ircBridge, enabled) {
     this._ircBridge = ircBridge;
     this._enabled = enabled;
+    this._expectations = [];
```
Kegsay: I'm confused what this is supposed to be representing?

lib/provisioning/Provisioner.js
```diff
@@ -129,24 +143,170 @@ Provisioner.prototype._validate = function(actual, parameterName) {
 };
 
 // Validate parameters for use in linking/unlinking
-Provisioner.prototype._validateAll = function(parameters) {
-    let parameterNames = ['matrix_room_id', 'remote_room_channel', 'remote_room_server'];
+Provisioner.prototype._validateAll = function(parameters, parameterNames) {
```
Kegsay: This is getting sufficiently complex that I'm unhappy for it to continue in its current form. Look into using the `matrix-appservice-bridge` component [ConfigValidator](http://matrix-org.github.io/matrix-appservice-bridge/1.2.0/ConfigValidator.html) and write a schema to represent the `/link` input JSON. You can then use it like so:
```javascript
var validator = new ConfigValidator({
  // .. schema as a JSON object
});
try {
  validator.validate(req.body);
}
catch (err) {
  // handle validation failures
}
```

lib/provisioning/Provisioner.js
```diff
@@ -129,24 +143,170 @@ Provisioner.prototype._validate = function(actual, parameterName) {
...
-Provisioner.prototype.link = Promise.coroutine(function*(options) {
+Provisioner.prototype.requestLink = Promise.coroutine(function*(options) {
     this._validateAll(options);
 
     let ircDomain = options.remote_room_server;
     let ircChannel = options.remote_room_channel;
     let roomId = options.matrix_room_id;
+    let opNick = options.op_nick;
+    let key = options.key || undefined; // Optional key
+    let bridger = options.bridger;
```
Kegsay: Vague var name. Is this a nick? User ID?

lib/provisioning/Provisioner.js
```diff
@@ -129,24 +143,170 @@ Provisioner.prototype._validate = function(actual, parameterName) {
...
+        let ircDomain = server.domain;
+
+        // (Matrix) Check power level of user
+        let matrixClient = this._ircBridge.getAppServiceBridge().getClientFactory().getClientAs();
+        var lookup = new StateLookup({
+            client : matrixClient,
+            eventTypes: ['m.room.power_levels']
+        });
+
+        yield lookup.trackRoom(roomId);
```
Kegsay: You don't need to be tracking any updates to the `m.room.power_levels` event you receive, so just use [getStateEvent](http://matrix-org.github.io/matrix-js-sdk/0.5.5/module-client-MatrixClient.html#getStateEvent).

lib/provisioning/Provisioner.js
```diff
@@ -129,24 +143,170 @@ Provisioner.prototype._validate = function(actual, parameterName) {
...
+
+        let powerState = lookup.getState(
+            roomId,
+            'm.room.power_levels'
+        )[0];
+
+        let actualPower = powerState.content.users[bridger];
+        let requiredPower = powerState.content.events["m.room.power_levels"];
```
Kegsay: This value may not always be defined, in which case you should be basing this off the `state_default` which is always defined. See [the spec](http://matrix.org/docs/spec/client_server/r0.2.0.html#m-room-power-levels) for more info.

lib/provisioning/Provisioner.js
```diff
@@ -129,24 +143,170 @@ Provisioner.prototype._validate = function(actual, parameterName) {
...
+        // (IRC) Check that op's nick is actually op
+        let info = yield this._ircBridge.checkNickExists(server, opNick);
+
+        if (!info.isOp) {
+            throw new Error('Provided user is not an Op');
+        }
+
+        // (Matrix) update room state
+        // State key for m.room.bridging
+        let skey = `${ircDomain}${ircChannel}`;
```
Kegsay: This should be a URI as per whiteboard

lib/provisioning/Provisioner.js
```diff
@@ -129,24 +143,170 @@ Provisioner.prototype._validate = function(actual, parameterName) {
...
+
+        // Send pending m.room.bridging
+        this._updateBridgingState(roomId, bridger, 'pending', skey);
+
+        // (IRC) Ask operator for authorisation
+        // Time that operator has to respond before giving up
+        let timeout = 10;
```
Kegsay: Units please.

lib/provisioning/Provisioner.js
```diff
@@ -129,24 +143,170 @@ Provisioner.prototype._validate = function(actual, parameterName) {
...
+
+        // Send PM to Bob
+        let botClient = yield this._ircBridge.getBotClient(server);
+        let authRequestAction = new IrcAction("message",
+            `${bridger} has requested to bridge ${roomId} with ${ircChannel} on this IRC ` +
+            `network. Respond with 'yes' to allow, or simply ignore this message to ` +
+            `dissallow. You have ${timeout} seconds from when this message was sent.`);
+        let pmRoom = yield botClient.joinChannel(opNick, key);
```
Kegsay: On IRC you don't technically "join" PM rooms, so is this actually necessary?

lib/provisioning/Provisioner.js
```diff
@@ -129,24 +143,170 @@ Provisioner.prototype._validate = function(actual, parameterName) {
...
+    let existingExpectation = this._expectations.find(
+        (expectation) => {
+            return expectation.server.domain === server.domain &&
+                   expectation.fromNick === fromNick;
+        }
+    );
+
+    if (existingExpectation) {
+        throw new Error(`Bridging request already sent to this op`+
+                        ` (${fromNick} on ${server.domain})`);
```
Kegsay: Saying by who would be useful.

lib/irc/BridgedClient.js
lukebarnard1: Fair enough, apparently the /WHO command might be the way to get a more detailed response.

lib/provisioning/Provisioner.js
```diff
@@ -129,24 +143,170 @@ Provisioner.prototype._validate = function(actual, parameterName) {
...
+);
+
+Provisioner.prototype._removeExpectation = function (expectation) {
+    expectation.defer = null;
+    this._expectations = this._expectations.filter((exp) => {
+        return exp.defer !== null;
+    });
+}
+
+Provisioner.prototype.expectPm = function(server, fromNick, timeout) {
```
Kegsay: I would prefer if this were done as a coroutine.

lib/provisioning/Provisioner.js
lukebarnard1: The array of expected authorisation resolutions. I could rename to ```pendingResponses```?

lib/provisioning/Provisioner.js
```diff
@@ -129,24 +143,170 @@ Provisioner.prototype._validate = function(actual, parameterName) {
...
+            `network. Respond with 'yes' to allow, or simply ignore this message to ` +
+            `dissallow. You have ${timeout} seconds from when this message was sent.`);
+        let pmRoom = yield botClient.joinChannel(opNick, key);
+
+        // Promise to receive PM from Bob or timeout
+        let opReplyPromise = this.expectPm(server, opNick, timeout);
+
+        this._ircBridge.sendIrcAction(pmRoom, botClient, authRequestAction);
+
+        opReplyPromise.then(
```
Kegsay: But you're in a coroutine already :( :( :( just say no to `then` and accept `yield` into your heart.

```javascript
try {
  yield opReplyPromise;
  // fulfilled function code here
}
catch (err) {
  // rejected function code here
}
```

lib/provisioning/Provisioner.js
```diff
@@ -129,24 +143,170 @@ Provisioner.prototype._validate = function(actual, parameterName) {
...
+    this._expectations.push(expectation);
+
+    return d.promise.timeout(timeout * 1000).then(
+        () => {this._removeExpectation(expectation)},
+        (err) => {this._removeExpectation(expectation); return Promise.reject(err);}
+    );
+}
+
+Provisioner.prototype.handlePm = function(server, fromUser, text) {
+    if (text.trim() !== 'yes') {
```
Kegsay: You should probably accept a case-insensitive match as well as simply "y".

Kegsay:
Broadly speaking, this is what I was hoping for, great! :D

lib/provisioning/Provisioner.js
Kegsay: If feels like the wrong data structure here. What you want is surely:
```javascript
{
  $domain: {
    $nick: {
      defer: Deferred
    }
  }
}
```
AKA a map of IRC users to Deferreds, where the Deferreds representing a pending auth request... so maybe `nicksToAuthRequests` or something?

lukebarnard1:
***This now depends on https://github.com/matrix-org/node-irc/pull/8***

lukebarnard1:
I've found a bug where if a non-public room is queried for power levels, a mapping gets created without the op being asked anything.

lukebarnard1:
- [ ] timeout should be variable

lukebarnard1:
@Kegsay PTAL when you get a chance :+1:

config.sample.yaml
```diff
@@ -286,3 +286,8 @@ ircService:
   provisioning:
     # True to enable the provisioning HTTP endpoint. Default: true.
     enabled: true
+    # Number of seconds to wait before giving up on getting a response from
+    # an IRC operator. If the operator does not respond within the alloted
```
Kegsay: Be specific when you say "operator": you mean channel operator but it's ambiguous as it stands.

Kegsay: Also, typo: `alloted`

lib/irc/BridgedClient.js
```diff
@@ -333,6 +333,51 @@ BridgedClient.prototype.whois = function(nick) {
...
+    if (this.disabled) {
+        return Promise.resolve({
+            server: this.server,
+            channel: channel,
+            names: {}
+        });
+    }
+    var self = this;
+    return new Promise(function(resolve, reject) {
+        self.unsafeClient.names(channel, function(channelName, names) {
```
Kegsay: Can you in-line comment what `names` looks like please.

lib/irc/BridgedClient.js
```diff
@@ -333,6 +333,51 @@ BridgedClient.prototype.whois = function(nick) {
...
+/**
+ * Get the operators of a channel
+ * @param {string} channel : The channel to call /names on
+ */
+BridgedClient.prototype.getOperators = function(channel) {
+    let modify = function (nicksInfo) {
+        let nicks = nicksInfo.nicks;
+
+        nicksInfo.operatorNicks = nicks.filter(
+            (index) => {
```
Kegsay: `filter()` calls the function for each element in the array, not the index, so this variable name is misleading. Please rename it to something less misleading e.g. `nick`.

lib/irc/BridgedClient.js
```diff
@@ -333,6 +333,51 @@ BridgedClient.prototype.whois = function(nick) {
...
+                return nicksInfo.names[index] === '@'
+            }
+        );
+        return nicksInfo;
+    }
+    return this.getNicks(channel).then(modify);
```
Kegsay: Any reason for not in-lining this?

spec/util/irc-client-mock.js
```diff
@@ -182,3 +182,9 @@ module.exports._autoConnectNetworks = function(addr, nick, networks) {
         }
     });
 };
+
+// For imitating events received by the client (such as an IRC op replying
+//  to a request for plumbing)
+module.exports.getClientEmitter = function () {
```
Kegsay: I don't think this is used anywhere? Remove it if so please.

lib/provisioning/Provisioner.js
```diff
@@ -91,62 +159,195 @@ Provisioner.prototype.isProvisionRequest = function(req) {
...
+            yield this._updateBridgingState(roomId, userId, 'failure', skey);
+            this._removeRequest(server, opNick);
+            return;
+        }
+        yield this._updateBridgingState(roomId, userId, 'success', skey);
     }
+);
+
+Provisioner.prototype._removeRequest = function (server, opNick) {
+    delete this._pendingRequests[server.domain][opNick];
```
Kegsay: Need a null guard like with get/set request.

lib/provisioning/Provisioner.js
```diff
@@ -91,62 +159,195 @@ Provisioner.prototype.isProvisionRequest = function(req) {
...
+            }, skey);
+        }
+        catch (err) {
+            console.error(err);
+            throw new Error(`Could not update m.room.bridging state in this room`);
+        }
+    }
+);
+
+Provisioner.prototype._authoriseProvisioning = Promise.coroutine(
```
Kegsay: A small summary of what this function does would be lovely.

lib/provisioning/Provisioner.js
```diff
@@ -91,62 +159,195 @@ Provisioner.prototype.isProvisionRequest = function(req) {
...
+
+        let existing = this._getRequest(server, opNick);
+        if (existing) {
+            let from = existing.userId;
+            throw new Error(`Bridging request already sent to `+
+                            `${opNick} on ${server.domain} from ${from}`);
+        }
+
+        // (Matrix) Check power level of user
+        log.info(`Check power level of user`);
```
Kegsay: Useless logging. Either add more info (which user, which room, etc) or remove this please.

lib/provisioning/Provisioner.js
```diff
@@ -91,62 +159,195 @@ Provisioner.prototype.isProvisionRequest = function(req) {
...
+        log.info(`Bot joined channel ${ircChannel}`);
+        let info = yield botClient.getOperators(ircChannel);
+
+        if (info.nicks.indexOf(opNick) == -1) {
+            let knownOps = info.operatorNicks.join(', ');
+            throw new Error(`Provided user is not in channel ${ircChannel}. ` +
+                            `Known ops in this channel: ${knownOps}`);
+        }
+
+        if (info.operatorNicks.indexOf(opNick) == -1) {
```
Kegsay: Nit: `===` should be the default. Only use `==` when you have a reason to do so.

lib/provisioning/Provisioner.js
```diff
@@ -91,62 +159,195 @@ Provisioner.prototype.isProvisionRequest = function(req) {
...
+            let knownOps = info.operatorNicks.join(', ');
+            throw new Error(`Provided user is not an op of ${ircChannel}. ` +
+                            `Known ops in this channel: ${knownOps}`);
+        }
+
+        // (Matrix) update room state
+        log.info(`Sending m.room.bridging`);
```
Kegsay: Useless logging. Add more info to this please.

lib/provisioning/Provisioner.js
```diff
@@ -91,62 +159,195 @@ Provisioner.prototype.isProvisionRequest = function(req) {
...
-        );
+        // Deliberately not yielding on this so that 200 OK is returned
+        log.info(`Contacting operator`);
+        this._contactOperator(
+            botClient, server, opNick, ircChannel,
+            roomId, userId, skey, timeoutSeconds);
     }
+);
+
+Provisioner.prototype._contactOperator = Promise.coroutine(
```
Kegsay: Misleading function name. It doesn't *just* contact the operator, it also then updates room state and more importantly, **create the link** if the contact was succesful.

lib/provisioning/Provisioner.js
```diff
@@ -91,62 +159,195 @@ Provisioner.prototype.isProvisionRequest = function(req) {
...
+    function*(botClient, server, opNick, ircChannel, roomId, userId, skey, timeoutSeconds) {
+        let d = promiseutil.defer();
+
+        // Send PM to operator
+        let authRequestAction = new IrcAction("message",
+            `${userId} has requested to bridge ${roomId} with ${ircChannel} on this IRC ` +
+            `network. Respond with 'yes' or 'y' to allow, or simply ignore this message to ` +
+            `dissallow. You have ${timeoutSeconds} seconds from when this message was sent.`);
```
Kegsay: Typo: `dissallow`

lib/irc/BridgedClient.js
lukebarnard1: Not really, I'll in-line it :+1:

Kegsay:
Aside from really cosmetic things, LGTM!

lib/provisioning/Provisioner.js
lukebarnard1: But deleting something that isn't there does nothing, which is expected in this case?

PR #146, opened 2016-08-24
Implement GET /_matrix/provision/queryLink

lukebarnard1:
This can be used to do a 'dummy' link request to check for errors with parameters. Currently, no further detail on the validation of parameters is given better than 'Parameters malformed'. This is an issue with ConfigValidator component of matrix-appservice-bridge. If the validation is a success, more interesting checks are done, possibly resulting in the following numerous errors:

- Server not found
- Server is configured to exclude channel
- Mapping already exists
- Bridging request already sent to this op
- Could not retrieve power levels
- You do not possess enough power to bridge this room
- Bot cannot join this channel
- Provided user is not a chan op in this channel (see example, below)
- Provided user is not in this channel

Results from this endpoint are given as a JSON blob, for example :

```JavaScript
{
    parameter_info: {
        remote_room_server: {},
        remote_room_channel: {},
        matrix_room_id: {},
        op_nick: {error : 'Provided user is not a chan op in this channel'},
        user_id: {}
    },
    // General info and errors
    infos: [],
    errors: [],
    // List of operators for the provided channel. Only defined if retrieved.
    operators: []
}
```

Any errors that occur in the querying process will be given in the normal way:
```JavaScript
{
    error: 'There was a problem querying for this link'
}
````

### DIFF
--- a/lib/provisioning/Provisioner.js
+++ b/lib/provisioning/Provisioner.js
@@ -148,6 +148,17 @@ function Provisioner(ircBridge, enabled, requestTimeoutSeconds) {
         }
     }));
 
+    as.app.get("/_matrix/provision/querylink", Promise.coroutine(function*(req, res) {
+        try {
+            let result = yield self.queryLink(req.body);
+            res.json(result);
+        }
+        catch (err) {
+            res.status(500).json({error: err.message});
+            throw err;
+        }
+    }));
+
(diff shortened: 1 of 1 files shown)


Kegsay:
Can you rebase this off `develop` please (which will require you to fix conflicts).

PR #150, opened 2016-08-30
(Rebase on develop) Implement /_matrix/provision/queryLink

lukebarnard1:
See https://github.com/matrix-org/matrix-appservice-irc/pull/146

### DIFF
--- a/lib/provisioning/Provisioner.js
+++ b/lib/provisioning/Provisioner.js
@@ -148,6 +148,17 @@ function Provisioner(ircBridge, enabled, requestTimeoutSeconds) {
         }
     }));
 
+    as.app.get("/_matrix/provision/querylink", Promise.coroutine(function*(req, res) {
+        try {
+            let result = yield self.queryLink(req.body);
+            res.json(result);
+        }
+        catch (err) {
+            res.status(500).json({error: err.message});
+            throw err;
+        }
+    }));
+
     if (enabled) {
         log.info("Provisioning started");
     }
@@ -175,26 +186,9 @@ Provisioner.prototype._updateBridgingState = Promise.coroutine(
     }
 );
 
-// Do a series of checks before contacting an operator for permission to create
-//  a provisioned mapping. If the operator responds with 'yes' or 'y', the mapping
-//  is created.
-// The checks done are the following:
-//  - (Matrix) Check power level of user is high enough
-//  - (IRC) Check that op's nick is actually a channel op
-//  - (Matrix) update room state m.room.brdiging
-Provisioner.prototype._authoriseProvisioning = Promise.coroutine(
-    function*(server, userId, ircChannel, roomId, opNick, key) {
-        let ircDomain = server.domain;
-
-        let existing = this._getRequest(server, opNick);
-        if (existing) {
-            let from = existing.userId;
-            throw new Error(`Bridging request already sent to `+
-                            `${opNick} on ${server.domain} from ${from}`);
-        }
-
-        // (Matrix) Check power level of user
-        log.info(`Check power level of ${userId} in ${roomId}`);
+Provisioner.prototype._userHasProvisioningPower = Promise.coroutine(
+    function*(userId, roomId) {
+        log.info(`Check power level of user`);
         let matrixClient = this._ircBridge.getAppServiceBridge().getClientFactory().getClientAs();
 
         let powerState = null;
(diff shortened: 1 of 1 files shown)


lukebarnard1:
Can I merge this, @Kegsay ?

lib/provisioning/Provisioner.js
```diff
@@ -341,6 +359,123 @@ Provisioner.prototype.handlePm = function(server, fromUser, text) {
...
+                 .error = 'Server is configured to exclude channel';
+    }
+
+    let entry = yield this._ircBridge.getStore().getRoom(roomId, ircDomain, ircChannel);
+    if (entry) {
+        queryInfo.errors.push('Mapping already exists');
+    }
+
+    // Errors with nick prevent further querying
+    if (queryInfo.parameter_info['op_nick'].error) {
```
Kegsay: When would this *ever* be true?

Kegsay: You probably meant to check for `opNick` instead, which is optional?

lib/provisioning/Provisioner.js
```diff
@@ -341,6 +359,123 @@ Provisioner.prototype.handlePm = function(server, fromUser, text) {
...
+        log.error(err.stack);
+        queryInfo.errors.push('Could not retrieve power levels');
+    }
+
+    if (!hasPower) {
+        queryInfo.errors.push('You do not possess enough power to bridge this room');
+    }
+
+    // There must be no error with the channel to continue
+    if (queryInfo.parameter_info['remote_room_channel'].error) {
```
Kegsay: I really dislike the indirection here. Having to check this entire monolithic function to see that `queryInfo.parameter_info['remote_room_channel'].error` is only ever set when `server.isExcludedChannel(ircChannel)` is true is really error-prone because these are done so far apart.

lib/provisioning/Provisioner.js
```diff
@@ -341,6 +359,123 @@ Provisioner.prototype.handlePm = function(server, fromUser, text) {
...
+    try {
+        yield botClient.joinChannel(ircChannel, key);
+    }
+    catch (err) {
+        log.error(err.stack);
+        queryInfo.parameter_info['remote_room_channel'].error = 'Bot cannot join this channel';
+        return queryInfo;
+    }
+
+    let opsInfo = yield botClient.getOperators(ircChannel);
```
Kegsay: Can this fail? Does it need to be try/catch'd?

lib/provisioning/Provisioner.js
```diff
@@ -341,6 +359,123 @@ Provisioner.prototype.handlePm = function(server, fromUser, text) {
...
+
+    queryInfo.operators = opsInfo.operatorNicks;
+
+    if (opsInfo.operatorNicks.indexOf(opNick) == -1) {
+        queryInfo.parameter_info['op_nick'].error = 'Provided user is not a chan op in this '+
+                                                    'channel';
```
Kegsay: In which channel? You can be more specific with this error message.

lib/provisioning/Provisioner.js
```diff
@@ -175,26 +186,9 @@ Provisioner.prototype._updateBridgingState = Promise.coroutine(
...
-            let from = existing.userId;
-            throw new Error(`Bridging request already sent to `+
-                            `${opNick} on ${server.domain} from ${from}`);
-        }
-
-        // (Matrix) Check power level of user
-        log.info(`Check power level of ${userId} in ${roomId}`);
+Provisioner.prototype._userHasProvisioningPower = Promise.coroutine(
+    function*(userId, roomId) {
+        log.info(`Check power level of user`);
```
Kegsay: Of which user? In which room?

lib/provisioning/Provisioner.js
```diff
@@ -341,6 +359,123 @@ Provisioner.prototype.handlePm = function(server, fromUser, text) {
...
+    let opNick = options.op_nick;
+    let key = options.key || undefined; // Optional key
+    let userId = options.user_id;
+
+    let queryInfo = {
+        // Map between options and info pertaining to them
+        // $parameter {
+        //      info: 'some information' ,
+        //      error: 'an error'
+        // }
```
Kegsay: You say this, but never seem to use the `info` property, only `error`.

It seems that this function is mainly doing validation, rather than giving any kind of useful information back to the caller (such as a list of operator nicks of whom you can contact, which iirc was the main reason for wanting this API in the first place).

lib/provisioning/Provisioner.js
```diff
@@ -341,6 +359,123 @@ Provisioner.prototype.handlePm = function(server, fromUser, text) {
...
+    }
+
+    if (server.isExcludedChannel(ircChannel)) {
+        queryInfo.parameter_info['remote_room_channel']
+                 .error = 'Server is configured to exclude channel';
+    }
+
+    let entry = yield this._ircBridge.getStore().getRoom(roomId, ircDomain, ircChannel);
+    if (entry) {
+        queryInfo.errors.push('Mapping already exists');
```
Kegsay: It feels weird that this is an error. We're just *querying* the link, we're not actively trying to make it.

lib/provisioning/Provisioner.js
lukebarnard1: But the theory goes that a channel is always provided with the API call (or validation will fail), so they already know the channel, which is why all info and errors returned don't reference the room, channel, server, etc.

lukebarnard1:
So IRL we decided that if an error occurs upon using ```queryLink``` should just stop the process of querying the link for operators. It also only expect the things it needs to get operators, instead of everything that ```/link``` expects. 

We also need ```/queryNetworks``` to get the list of available networks on the bridge.

- [ ] queryLink is fixed
- [ ] ```/querNetworks```

lib/provisioning/Provisioner.js
```diff
@@ -341,6 +378,81 @@ Provisioner.prototype.handlePm = function(server, fromUser, text) {
...
+//  }
+Provisioner.prototype.queryLink = Promise.coroutine(function*(options) {
+    let ircDomain = options.remote_room_server;
+    let ircChannel = options.remote_room_channel;
+    let key = options.key || undefined; // Optional key
+
+    let queryInfo = {
+        // Array of operator nicks
+        operators: []
+    }
```
Kegsay: Semicolon

Kegsay:
LGTM

PR #155, opened 2016-09-05
Use Travis for building pushes and PRs 

lukebarnard1:


Kegsay:
Does this actually fail if there are failing tests or lint?

I suspect it won't, because the `ci` script goes to great lengths to **not** return a non-zero exit code to make Jenkins run all the commands and not stop at the first one that fails. Can you please double-check that this is returning a non-zero exit code correctly on failures (use `$?` tto grab the last command's exit code):
```
$ npm run ci
$ echo $?
0
```

Kegsay:
LGTM

PR #157, opened 2016-09-05
Prevent route loops when provisioning new mappings

lukebarnard1:
The provisioner will fail to request for a mapping if the m.room.bridging status in a room for the same channel (i.e. with the same state key) is not = 'failure'. Meaning, if it is either pending or a success, the request will fail (unless the bridging requester is the same as the ```user_id``` in the ```m.room.bridging```).

Maybe the check for ```user_id``` should really be with the sender, meaning the bridge checks that its own matrix bot was the one to send the bridging. But currently, when you get an event, you don't get it's original sender, so we may have to add extra state to the content of the ```m.room.bridging``` indicating the matrix id of the bridge that set it.

### DIFF
--- a/lib/provisioning/Provisioner.js
+++ b/lib/provisioning/Provisioner.js
@@ -181,6 +181,8 @@ Provisioner.prototype._updateBridgingState = Promise.coroutine(
 // The checks done are the following:
 //  - (Matrix) Check power level of user is high enough
 //  - (IRC) Check that op's nick is actually a channel op
+//  - (Matrix) check room state to prevent route looping: don't bridge the same
+//    room-channel pair
 //  - (Matrix) update room state m.room.brdiging
 Provisioner.prototype._authoriseProvisioning = Promise.coroutine(
     function*(server, userId, ircChannel, roomId, opNick, key) {
@@ -248,11 +250,46 @@ Provisioner.prototype._authoriseProvisioning = Promise.coroutine(
                             `Known ops in this channel: ${knownOps}`);
         }
 
-        // (Matrix) update room state
         // State key for m.room.bridging
         let skey = `irc://${ircDomain}/${ircChannel}`;
+
+        // (Matrix) check room state to prevent route looping
+        try {
+            let bridgingState = yield matrixClient.getStateEvent(roomId, 'm.room.bridging', skey);
+            // Bridging state exists and is either success or pending (ignore failures)
+            if (bridgingState && bridgingState.status !== 'failure') {
+                // If bridging state sender is the bridge matrix user
+                if (bridgingState.user_id === userId) {
+                    // Success, already pending/success
+                    log.info(`Bridging state already exists in room ${roomId} ` +
+                             `(status = ${bridgingState.status},` +
+                             ` bridger = ${bridgingState.user_id}. Ignoring request.`);
+                    return;
+                }// If it is from a different sender, fail
+                throw new Error(`A request to create this mapping has already been sent ` +
+                         `(status = ${bridgingState.status},` +
+                         ` bridger = ${bridgingState.user_id}. Ignoring request.`);
+            }
+        }
+        catch (err) {
+            // The request to discover bridging state has failed
+
+            // http-api error indicated by errcode
+            if (err.errcode) {
+                //  ignore M_NOT_FOUND: this bridging does not exist
+                console.error(err);
+                if (err.errcode !== 'M_NOT_FOUND') {
+                    throw new Error(err.data.error);
+                }
+            }
+            else {
+                throw err;
+            }
+        }
+
         log.info(`Sending m.room.bridging to ${roomId}, state key = ${skey}`);
 
+        // (Matrix) update room state
         // Send pending m.room.bridging
         yield this._updateBridgingState(roomId, userId, 'pending', skey);
 
--- a/lib/provisioning/Provisioner.js
+++ b/lib/provisioning/Provisioner.js
@@ -255,10 +255,13 @@ Provisioner.prototype._authoriseProvisioning = Promise.coroutine(
 
         // (Matrix) check room state to prevent route looping
         try {
-            let bridgingState = yield matrixClient.getStateEvent(roomId, 'm.room.bridging', skey);
+            let wholeBridgingState = yield matrixClient.getWholeStateEvent(
+                roomId, 'm.room.bridging', skey
+            );
+            let bridgingState = wholeBridgingState.content;
             // Bridging state exists and is either success or pending (ignore failures)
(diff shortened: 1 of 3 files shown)


lib/provisioning/Provisioner.js
```diff
@@ -248,11 +250,46 @@ Provisioner.prototype._authoriseProvisioning = Promise.coroutine(
...
-        // (Matrix) update room state
         // State key for m.room.bridging
         let skey = `irc://${ircDomain}/${ircChannel}`;
+
+        // (Matrix) check room state to prevent route looping
+        try {
+            let bridgingState = yield matrixClient.getStateEvent(roomId, 'm.room.bridging', skey);
+            // Bridging state exists and is either success or pending (ignore failures)
+            if (bridgingState && bridgingState.status !== 'failure') {
+                // If bridging state sender is the bridge matrix user
```
Kegsay: s/bridge matrix user/provisioner/

Kegsay:
LGTM with 1 teeny test :)

lukebarnard1:
@Kegsay as per IRL, PTAL

lib/provisioning/Provisioner.js
```diff
@@ -287,11 +302,84 @@ Provisioner.prototype._authoriseProvisioning = Promise.coroutine(
...
+                (e) => {
+                    return e.type === 'm.room.bridging' && e.state_key === skey
+                }
+            );
+            // Bridging state exists and is either success or pending (ignore failures)
+            if (wholeBridgingState &&
+                wholeBridgingState.content) {
+                let bridgingState = wholeBridgingState.content;
+
+            if (bridgingState.status !== 'failure') {
```
Kegsay: Indentation is weird.

lib/provisioning/Provisioner.js
```diff
@@ -287,11 +302,84 @@ Provisioner.prototype._authoriseProvisioning = Promise.coroutine(
...
+        try {
+            let roomState = yield matrixClient.roomState(roomId);
+            let wholeBridgingState = roomState.find(
+                (e) => {
+                    return e.type === 'm.room.bridging' && e.state_key === skey
+                }
+            );
+            // Bridging state exists and is either success or pending (ignore failures)
+            if (wholeBridgingState &&
+                wholeBridgingState.content) {
```
Kegsay: Double indent please to make it clear where the if body begins.

lib/provisioning/Provisioner.js
```diff
@@ -287,11 +302,84 @@ Provisioner.prototype._authoriseProvisioning = Promise.coroutine(
...
+
+            if (bridgingState.status !== 'failure') {
+                // If bridging state sender is this bot
+                if (wholeBridgingState.sender === matrixClient.credentials.userId) {
+                    // Success, already pending/success
+                    log.info(`Bridging state already exists in room ${roomId} ` +
+                             `(status = ${bridgingState.status},` +
+                             ` bridger = ${bridgingState.user_id}.)`);
+
+                    // If pending, resend the message to the op as if it were the original
```
Kegsay: This comment should be with the if pending block and not the if success block.

lib/provisioning/Provisioner.js
```diff
@@ -287,11 +302,84 @@ Provisioner.prototype._authoriseProvisioning = Promise.coroutine(
...
+                else {// If it is from a different sender, fail
+                    throw new Error(`A request to create this mapping has already been sent ` +
+                             `(status = ${bridgingState.status},` +
+                             ` bridger = ${bridgingState.user_id}. Ignoring request.`);
+                }
+            }
+            }
+        }
+        catch (err) {
+            // The request to discover bridging state has failed
```
Kegsay: Lies. This try block covers more than just getting the bridge state event.

lib/provisioning/Provisioner.js
```diff
@@ -365,20 +461,35 @@ Provisioner.prototype._setRequest = function (server, opNick, request) {
     this._pendingRequests[server.domain][opNick] = request;
 }
 
-Provisioner.prototype.handlePm = function(server, fromUser, text) {
+Provisioner.prototype.handlePm = Promise.coroutine(function*(server, fromUser, text) {
     if (['y', 'yes'].indexOf(text.trim().toLowerCase()) == -1) {
         log.warn(`Provisioner only handles text 'yes'/'y' ` +
                  `(from ${fromUser.nick} on ${server.domain})`);
+
+        this._sendToUser(
```
Kegsay: We should probably yield here so handlePm only returns after we have sent this message.

lib/provisioning/Provisioner.js
```diff
@@ -365,20 +461,35 @@ Provisioner.prototype._setRequest = function (server, opNick, request) {
...
+        yield this._sendToUser(
+            fromUser.nick, server,
+            'Thanks for your response, bridge request authorised.'
+        );
+
         return;
     }
     log.warn(`Provisioner was not expecting PM from ${fromUser.nick} on ${server.domain}`);
-}
+    this._sendToUser(
```
Kegsay: Yield

lib/provisioning/Provisioner.js
```diff
@@ -513,13 +617,41 @@ Provisioner.prototype._doLink = Promise.coroutine(
...
+        if (entry) {
             throw new Error(`Room mapping already exists (${mappingLogId},` +
                             `origin = ${entry.data.origin})`);
         }
+        yield this._ircBridge.getStore().storeRoom(ircRoom, mxRoom, 'provision');
+
+        // Cause the bot to join the new plumbed channel (if it is enabled (see joinChannel))
+        // TODO: key not persisted on restart
+        let botClient = yield this._getBotClientForServer(server);
+        botClient.joinChannel(ircChannel, key);
```
Kegsay: Yield?

lib/provisioning/Provisioner.js
```diff
@@ -551,12 +683,14 @@ Provisioner.prototype.unlink = Promise.coroutine(function*(options) {
...
-    }
-    else {
+    if (!entry) {
         throw new Error(`Provisioned room mapping does not exist (${mappingLogId})`);
     }
+    yield this._ircBridge.getStore().removeRoom(roomId, ircDomain, ircChannel, 'provision');
+
+    // Cause the bot to part the channel (if it is enabled (see leaveChannel))
+    let botClient = yield this._getBotClientForServer(server);
+    botClient.leaveChannel(ircChannel);
```
Kegsay: Yield?

Kegsay:
LGTM aside from comments. Not sure in some places if you should be yielding, but otherwise we're just dropping those promises to the floor without logging errors which is bad.

PR #158, opened 2016-09-08
Pass an explicit object of controller methods to new Bridge()

leonerd:
Rather than implying via `this` directly, because now it's much easier to extend with `thirdPartyLookup`. Also this way is easier to grep for and for casual readers to understand.


leonerd:
I'm not quite sure why one of the travises (travisii? travum?) appears unhappy. The nature of the errors doesn't look very related to my change. The other is OK.

Kegsay:
LGTM

PR #159, opened 2016-09-08
Better Provisioning 

lukebarnard1:
- MatrixBot will now join and then part an IRC channel to get the operator nicks in a room.
- MatrixBot will now join and stay in a provisioned channel
- The provisioning matrix user will be 'fake joined' into the IRC channel
- The provisioner will respond to 'yes' or 'no' or any message the IRC chanop sends it with something a bit more useful, like confirmation of the provisioned mapping.
- Unlinking actually unlinks

TODO:
-  ```m.room.bridging```: Recall pending states so long as the ```sender``` is the same as the bridge bot **see #157**
- [x] Make the tests pass - tests really don't like injecting a fake event to join the provisioner to the bridged room


Kegsay:
Can we close this PR please since it is entirely obsolete.

PR #164, opened 2016-09-12
Improved Provisioning

lukebarnard1:
- Bot only leaves when it the bridge no longer maps that IRC channel
- Better error messages given when linking
- Cache operator nicks per channel for 10s when using ```queryLink```

### DIFF
--- a/lib/irc/BridgedClient.js
+++ b/lib/irc/BridgedClient.js
@@ -349,7 +349,7 @@ BridgedClient.prototype.whois = function(nick) {
  * @param {integer} cacheDurationMs : Optional. The duration of time to keep a
  * list of operator nicks cached. If > 0, the operator nicks will be returned
  * whilst the cache is still valid and it will become invalid after cacheDurationMs
- * milliseconds.
+ * milliseconds. Cache will not be used if left undefined.
  */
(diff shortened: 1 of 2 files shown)


lib/irc/BridgedClient.js
```diff
@@ -342,8 +346,35 @@ BridgedClient.prototype.whois = function(nick) {
  * Get the operators of a channel
  * @param {string} channel : The channel to call /names on
  * @param {string} key : Optional. The key to use to join the channel.
+ * @param {integer} cacheDurationMs : Optional. The duration of time to keep a
+ * list of operator nicks cached. If > 0, the operator nicks will be returned
+ * whilst the cache is still valid and it will become invalid after cacheDurationMs
+ * milliseconds. Cache will not be used if left undefined.
  */
-BridgedClient.prototype.getOperators = function(channel, key) {
+BridgedClient.prototype.getOperators = function(channel, key, cacheDurationMs) {
```
leonerd: Rather than these two positional arguments (and the complications of detecting which by type) it might be nicer to pass in `opts` in the usual named-argument style. Also makes callsites neater with a `{timeout: 10000}` rather than just a trailing number.

PR #165, opened 2016-09-12
Only respond to room alias requests that embed a valid IRC channel name

### LINKED ISSUE #161: Bridge accepts alias requests that lack hash signs
To join an e.g. freenode channel of `#foo`, requires the alias `#freenode_#foo:matrix.org`; i.e. that the hash symbol must be repeated within the alias name.

The bridge will accept as valid a lookup request for `#freenode_foo:matrix.org`, and thus try to join a channel on Freenode called `foo`, lacking the hash symbol, which fails in a variety of nonobvious ways which often confuse newbie users (including myself) who forget that you need to type that second `#` symbol.

The bridge ought to outright reject alias lookup requests that lack this second (or third in the case of `##channels`) hash symbol.

leonerd:
I.e. channel names that start with a # (fixes #161)

### DIFF
--- a/lib/irc/IrcServer.js
+++ b/lib/irc/IrcServer.js
@@ -237,7 +237,7 @@ IrcServer.prototype.claimsAlias = function(alias) {
             "$SERVER": this.domain
         },
         {
-            "$CHANNEL": "(.*)"
+            "$CHANNEL": "#(.*)"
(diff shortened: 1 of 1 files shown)


lukebarnard1:
LGTM

PR #168, opened 2016-09-13
Use 3PL to get provisioning queryNetworks info

lukebarnard1:
Also, pass ```bot_user_id``` when acquiring 3PL

### DIFF
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -428,11 +428,11 @@ IrcBridge.prototype.getThirdPartyProtocol = function(protocol) {
     var servers = this.getServers();
 
     return Promise.resolve({
-        "bot_user_id": this.getAppServiceUserId(),
         "user_fields": ["domain", "nick"],
         "location_fields": ["domain", "channel"],
         "instances": servers.map(function (server) {
             return {
+                bot_user_id: this.getAppServiceUserId(),
                 desc: server.config.description || server.domain,
                 icon: server.config.icon,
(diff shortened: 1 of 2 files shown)


lib/bridge/IrcBridge.js
```diff
@@ -428,6 +428,7 @@ IrcBridge.prototype.getThirdPartyProtocol = function(protocol) {
     var servers = this.getServers();
 
     return Promise.resolve({
+        "bot_user_id": this.getAppServiceUserId(),
```
leonerd: If it's up here, it'll be merged wrongly by the synapse. It wants to live in the `instances` list, so that when synapse merges multiple instances lists across multiple AS programs, they all preserve the right value.

lib/provisioning/Provisioner.js
```diff
@@ -565,10 +565,13 @@ Provisioner.prototype.queryLink = Promise.coroutine(function*(options) {
...
 });
 
-// Get the list of currently configured networks (from the config)
+// Get the list of currently network instances
 Provisioner.prototype.queryNetworks = Promise.coroutine(function*() {
+    let thirdParty = yield this._ircBridge.getThirdPartyProtocol();
+
     return {
-        servers: this._ircBridge.ircServers.map((server) => {return server.domain})
+        bot_user_id: thirdParty.bot_user_id,
```
leonerd: Having done [theabove] you can still do this here, by

    bot_user_id: thirdParty.instances[0].bot_user_id,

PR #169, opened 2016-09-13
Provisioner leaves channels/rooms upon unlink

lukebarnard1:


leonerd:
LGTM*

*: By which I mean: well it looks like valid JavaScript code and it's calling methods that appear to exist and probably do the right thing and it is observed to function correctly on manual testing, but I'm not sure I'm truely best placed to qualify if it's the most appropriate way.

PR #170, opened 2016-09-15
Prevent admin room from being created when plumbing

lukebarnard1:
The bridge will not create an admin room if there is plumbing state in the room of the form:

```JS
m.room.plumbing: {
    content: {
        status: "enabled"
    },
    ...
}
```


leonerd:
LGTM I guess? :) Not being too familiar with how the IRC bridge works...

lukebarnard1:
I will write tests for this later

PR #181, opened 2016-09-20
Message to op changed

lukebarnard1:
Alias, but no room name:

```
@person:matrix.org has requested to bridge "!m42jsksfgloi9ueot:matrix.org" 
(#somealias:matrix.org) with #somechannel on this IRC network. Respond 
with 'yes' or 'y' to allow, or simply ignore this message to disallow. You have
 300 seconds from when this message was sent.
```
Room name and alias:
```
@person:matrix.org has requested to bridge "Matrix Room" 
(#somealias:matrix.org) with #somechannel on this IRC network. 
Respond with 'yes' or 'y' to allow, or simply ignore this message to 
disallow. You have 300 seconds from when this message was sent.
```

Room name, no alias
```
@person:matrix.org has requested to bridge "Matrix Room" with 
#somechannel on this IRC network. Respond with 'yes' or 'y' to 
allow, or simply ignore this message to disallow. You have 300 
seconds from when this message was sent.
```

### DIFF
--- a/lib/provisioning/Provisioner.js
+++ b/lib/provisioning/Provisioner.js
@@ -453,14 +453,23 @@ Provisioner.prototype._createAuthorisedLink = Promise.coroutine(
             log.error(`Error retrieving room alias (${err.data.error})`);
         }
 
-        let roomDesc = `"${roomId}"`;
-
-        if (nameState && nameState.name) {
-            roomDesc = `"${nameState.name}"`;
-        }
+        let roomDesc = null;
+        let matrixToLink = `https://matrix.to/#/${roomId}`;
 
         if (aliasState && aliasState.alias) {
-            roomDesc = `${roomDesc} (${aliasState.alias})`;
+            roomDesc = aliasState.alias;
+            matrixToLink = `https://matrix.to/#/${aliasState.alias}`;
+        }
+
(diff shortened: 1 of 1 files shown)


Kegsay:
I think @ara4n wants to never show the room id under any circumstances, but not sure.

Kegsay:
Otherwise LGTM!

ara4n:
I suggest we just put a matrix.to link in there.  So for the three scenarios:

1. User wants to bridge #foo:matrix.org (https://matrix.to/#/#foo:matrix.org)
2. User wants to bridge 'Matrix HQ' (https://matrix.to/#/#matrix:matrix.org)
3. User wants to bridge 'Matthew and Luke' (https://matrix.to/#/!sgjhc2tuy3:matrix.org)

N.B. that we can always synthesize human readable names if the room hasn't got an explicit m.room.name - the spec has a section on the rules to follow.

Kegsay:
So the format is `https://matrix.to/#/$ROOM_ID_OR_ALIAS`?

Kegsay:
Also, you probably need to unbreak tests given you're now doing unexpected HTTP requests.

Kegsay:
LGTM

PR #183, opened 2016-09-21
Prevent race occuring whilst spamming 'yes'->MBot

lukebarnard1:
When replying to a bot in response to a provisioning request, spamming 'yes' can cause 'bridge request authorised' to be sent back multiple times, indicating that the same mapping is being created over and over again, or at least that the Provisioner is trying to do that.

The fix is to only return pending requests that have pending promises, i.e. those that have not been resolved/cancelled/rejected.


Kegsay:
LGTM

Issue #60, opened 2016-06-02
IRC servers on IPv6 addresses need to be entered directly


Kegsay:
https://matrix.org/jira/browse/BOTS-136

If you type a domain, it'll do the AAAA-lookup but then use v4, obviously. If it find a valid v6, it should really be using it. This is on Node v0.10.x

PR #196, opened 2016-09-27
Specifiy family = 6 when prefix is set

lukebarnard1:
Otherwise, it is assumed that the IRC server ONLY does IPv6

**NOTE: This depends on some uncommitted node-irc code: https://github.com/matrix-org/node-irc/pull/11**

### DIFF
--- a/lib/irc/ConnectionInstance.js
+++ b/lib/irc/ConnectionInstance.js
@@ -245,7 +245,7 @@ ConnectionInstance.create = function(server, opts, onCreatedCallback) {
         secure: server.useSsl(),
         selfSigned: server.useSslSelfSigned(),
         retryCount: 0,
-        family: server.getIpv6Prefix() === undefined ? 4 : 6,
+        family: server.getIpv6Prefix() ? 6 : 4,
     };
     // TODO : coroutine this
     var d = promiseutil.defer();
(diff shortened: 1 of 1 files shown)


lib/irc/ConnectionInstance.js
```diff
@@ -244,7 +244,8 @@ ConnectionInstance.create = function(server, opts, onCreatedCallback) {
         port: server.getPort(),
         secure: server.useSsl(),
         selfSigned: server.useSslSelfSigned(),
-        retryCount: 0
+        retryCount: 0,
+        family: server.getIpv6Prefix() === undefined ? 4 : 6,
```
Kegsay:
```js
family: server.getIpv6Prefix() ? 6 : 4,
```
looks clearer to me.

Kegsay:
> Otherwise, it is assumed that the IRC server ONLY does IPv6

OOI, why is this a problem?

lukebarnard1:
Because, with freenode and OFTC, the servers allow both, instead of having a domain for IPv6 only, like Snoonet

Kegsay:
Oh? I'd have thought specifying a v6 source address is enough for the IRCd to say "hey look some random v6 address is talking to me, I'll use v6"? Orrrr?

lukebarnard1:
It's the DNS lookup on the client that will happen to find the A result before the AAAA result

Kegsay:
Ohhhhh *that*. I'm sure I bumped into this...

Kegsay:
LGTM! (Please close #60 then too!)

Issue #188, opened 2016-09-26
Log how long the Scheduler queue gets


Kegsay:
Over the weekend, Freenode had a huge netsplit. `Scheduler` logic correctly kicked in, making a humongous queue for reconnections. It looks like this queue was taking forever to clear (probably due to said netsplit). Some numbers: Between 18:00~19:00 on 2016-09-24, there were 1724 reconnections queued up, but only 74 successful connections (that is, the connection got `rpl_welcome`). This led to massive head-of-line blocking, effectively stopping M->I traffic

Kegsay:
We're kinda stuck between a rock and a hard place with this. By queueing sensibly, we avoid being k-lined from reconnect speed, but it means that things like this can happen.

Kegsay:
Here's a fun graph illustrating the severity of that netsplit: (Red is Freenode)
![screencapture-matrix-org-prometheus-graph-1474880459627](https://cloud.githubusercontent.com/assets/7190048/18828552/4e04bf98-83d0-11e6-8532-321bb118087e.png)

When the red line hits 0 that is when it was restarted.

Kegsay:
@leonerd suggests that we can mitigate this a little by effectively allowing a certain number of "simultaneous connections" - that is, we allow N connections to be up-in-the-air at any one time. This helps against 1 nasty TCP connection slowing everything down. We would need to experiment with what to set N to.

Of course, the long term solution is to just use IPv6....

Kegsay:
@ara4n - thoughts?

Kegsay:
tl;dr - Use IPv6 on Freenode.

Kegsay:
- Tagged as support to move Freenode to IPv6 (we need to speak to them first).
- Tagged as bug because we really do need that logging in general.

lukebarnard1:
Freenode bridge is now using IPv6 and scheduled queue length is being logged.

PR #200, opened 2016-09-28
Fix log scheduled queue length

lukebarnard1:
Fix for #188

Log line looks like:

```
2016-09-28 09:58:48 INFO:scheduler Queue for localhost length = 1
```


Kegsay:
LGTM

PR #201, opened 2016-09-28
Add optional media_url to homeserver config

### LINKED ISSUE #154: have a separate setting for internal/external URL
If you upload an image, it uses the homeserver url for the IRC side of it, which makes sense until that is "localhost:8008" :/

lukebarnard1:
This is used souly for the purposes of generating http URLs from mxcs.

Fixes #154


lukebarnard1:
We seem to be getting the old 'Trying to join #coffee' again. Seems that ```Created new bot client for irc.example``` is being called multiple times, and the bot clients are not destroyed after tests.

lukebarnard1:
FTR The tests failing are probably due to flaky tests, not this feature causing issues - it's a very simple feature.

lib/bridge/MatrixHandler.js
```diff
@@ -993,9 +996,13 @@ MatrixHandler.prototype._sendIrcAction = Promise.coroutine(
...
         event.content.body = fileName;
 
+        let mediaUrl = this.mediaUrl;
+
         // Create a file event to reflect the recent upload
         let mAction = MatrixAction.fromEvent(
-            this.ircBridge.getAppServiceBridge().getClientFactory().getClientAs(), event
+            this.ircBridge.getAppServiceBridge().getClientFactory().getClientAs(),
+            event,
+            mediaUrl
```
Kegsay: Inline this please.

lib/models/MatrixAction.js
```diff
@@ -45,7 +45,13 @@ MatrixAction.fromEvent = function(client, event) {
...
                     "KB)";
             }
-            text = client.mxcUrlToHttp(event.content.url) +
+
+            // By default assume that the media server = client homeserver
+            if (!mediaUrl) {
+                mediaUrl = client.getHomeserverUrl();
+            }
+
+            text = client.mxcUrlToAlternateHttp(mediaUrl, event.content.url) +
```
Kegsay: Where is `mxcUrlToAlternateHttp`? What happened to doing https://github.com/matrix-org/matrix-appservice-irc/issues/154#issuecomment-248430575 ?

lib/bridge/MatrixHandler.js
lukebarnard1: Done.

lib/models/MatrixAction.js
Kegsay: So the `ContentRepo` class is currently being exported from the js-sdk @ https://github.com/matrix-org/matrix-js-sdk/blob/master/lib/matrix.js#L53 - I would much rather we exposed the `js-sdk` from `matrix-appservice-bridge` much like we do with `AppServiceRegistration` at https://github.com/matrix-org/matrix-appservice-bridge/blob/master/lib/exports.js#L29 than add this function to the client.

lib/bridge/MatrixHandler.js
Kegsay: Not pushed though.

lib/models/MatrixAction.js
lukebarnard1: Done, but how shall we include https://github.com/matrix-org/matrix-appservice-bridge/pull/31 if mx-as-irc is pinned at v1.3 of mx-as-bridge?

Kegsay:
Need tests to pass (merge develop in) then can you add a test for this please :angel:

lukebarnard1:
FTR, the tests now fail on not having https://github.com/matrix-org/matrix-appservice-bridge/commit/e2dec16854f364f7a07f71815540138a93ded0ab from mx-as-bridge

Kegsay:
LGTM - as soon as we've bumped the bridge lib and updated package.json to 1.3.1 then this can be merged.

PR #202, opened 2016-09-28
Do not block the connection queue for 1h

lukebarnard1:
Prevent banned users from head-of-line blocking the queue for one hour. Scheduling delays effect the entire queue, as intended, but a single banned user should not prevent others from connecting.


Kegsay:
LGTM

Issue #166, opened 2016-09-12
Tests are flaky


leonerd:
The tests that run under travis appear to be flaky. Commits can fail that subsequently pass after forcing a retest by pushing an empty commit.

An issue I have observed is the test:

```
  1) Matrix-to-IRC message bridging should bridge matrix topics as IRC topics
   Message:
     timeout: timed out after 5000 msec waiting for spec to complete
   Stacktrace:
     undefined
```

But I believe @lukebarnard1 has some others too.

leonerd:
The detailed log from my failing test:

```
Matrix-to-IRC message bridging : should bridge matrix topics as IRC topics
==========================================================================
Add listener(irc.example_M-flibble) for fn=connect
Add listener(irc.example_M-flibble) for fn=join
Add listener(irc.example_ro_bot_nick) for fn=connect
Add listener(irc.example_ro_bot_nick) for fn=join
2016-09-12 13:29:36 ERROR:BridgedClient <ro_bot_nick@irc.example#hbxhxrimee0cs> Timed out trying to join #coffee - trying again.
2016-09-12 13:29:36 DEBUG:BridgedClient <ro_bot_nick@irc.example#hbxhxrimee0cs> Joining channel #coffee
2016-09-12 13:29:36 ERROR:BridgedClient <ro_bot_nick@irc.example#vfspbq83fvkg> Timed out trying to join #coffee - trying again.
2016-09-12 13:29:36 DEBUG:BridgedClient <ro_bot_nick@irc.example#vfspbq83fvkg> Joining channel #coffee
2016-09-12 13:29:36 ERROR:BridgedClient <ro_bot_nick@irc.example#5g1owfudkzk0w> Timed out trying to join #coffee - trying again.
2016-09-12 13:29:36 DEBUG:BridgedClient <ro_bot_nick@irc.example#5g1owfudkzk0w> Joining channel #coffee
2016-09-12 13:29:36 ERROR:BridgedClient <ro_bot_nick@irc.example#igmgjfjbhw8c8> Timed out trying to join #coffee - trying again.
...
```

lukebarnard1:
And on another unrelated test:

```
IRC-to-Matrix mode bridging : should set join_rules to 'invite' on +i.
======================================================================
Add listener(irc.example_ro_bot_nick) for fn=join
Add listener(irc.example_ro_bot_nick) for fn=connect
[...]

Issue #74, opened 2016-06-17
Add killswitch to IRC clients


Kegsay:
Currently, if an IRC client is created and told to connect to a network and some channels then it will do just that. It will periodically retry if any failures occur and there is no way to actually stop it from trying.

This sucks for tests because:
 - You get log spam leaking out from other tests (which are still trying to reconnect on a NOP mock)
 - Any fatal errors can result in sporadic, racey test failures - see https://github.com/matrix-org/matrix-appservice-irc/commit/b621a40e31bb975e0121f452d972a7e6986d26a3 for an example of this.

We need some way of communicating to the internal client logic to say "no really, please just stop.". Unfortunately, there is no real error code we can piggyback on for this from the IRC-side, which makes me think this should be an external-to-the-protocol killswitch. This makes me feel a bit sad though since it's additional logic simply for the benefit of testing, but I feel it is necessary for sanity.

lukebarnard1:
Flakiness is now gone: Fixed by #205

Mikaela:
So what is the command to kill the IRC connection?

lukebarnard1:
@Mikaela: that's actually https://github.com/matrix-org/matrix-appservice-irc/pull/253 TL;DR it's `!quit server.name`

PR #205, opened 2016-09-29
Attempt to fix problem with bot clients (and remove flaky tests)

lukebarnard1:
They were joining forever and never dying after tests.

### DIFF
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -303,10 +303,10 @@ IrcBridge.prototype.run = Promise.coroutine(function*(port) {
     this._startedUp = true;
 });
 
-IrcBridge.prototype.kill = Promise.coroutine(function*() {
-    this._bridge.getBot().unsafeClient = null;
-    log.info("Bot client killed");
-});
+IrcBridge.prototype.kill = function() {
+    log.info("Killing bot client");
+    return this._bridge.getBot().kill();
+}
 
 IrcBridge.prototype.isStartedUp = function() {
     return this._startedUp;
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -303,6 +303,13 @@ IrcBridge.prototype.run = Promise.coroutine(function*(port) {
     this._startedUp = true;
 });
 
+// Kill the bridge by killing all IRC clients in memory.
+//  Killing a client means that it will disconnect forever
+//  and never do anything useful again.
+//  There is no guarentee that the bridge will do anything
+//  usefull once this has been called.
+//
+//  See (BridgedClient.prototype.kill)
 IrcBridge.prototype.kill = function() {
     if (!this._bridge.getBot()) {
         log.info("No bot client to kill");
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -304,8 +304,12 @@ IrcBridge.prototype.run = Promise.coroutine(function*(port) {
 });
 
 IrcBridge.prototype.kill = function() {
-    log.info("Killing bot client");
-    return this._bridge.getBot().kill();
+    if (!this._bridge.getBot()) {
+        log.info("No bot client to kill");
+        return Promise.resolve({});
+    }
+    log.info("Killing all clients");
+    return this._clientPool.killAllClients();
 }
 
 IrcBridge.prototype.isStartedUp = function() {
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -311,10 +311,6 @@ IrcBridge.prototype.run = Promise.coroutine(function*(port) {
 //
 //  See (BridgedClient.prototype.kill)
 IrcBridge.prototype.kill = function() {
-    if (!this._bridge.getBot()) {
-        log.info("No bot client to kill");
-        return Promise.resolve({});
-    }
     log.info("Killing all clients");
     return this._clientPool.killAllClients();
 }
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -303,6 +303,10 @@ IrcBridge.prototype.run = Promise.coroutine(function*(port) {
     this._startedUp = true;
 });
 
+IrcBridge.prototype.kill = Promise.coroutine(function*() {
+    this._bridge.getBot().unsafeClient = null;
+});
+
 IrcBridge.prototype.isStartedUp = function() {
     return this._startedUp;
 };
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -305,6 +305,7 @@ IrcBridge.prototype.run = Promise.coroutine(function*(port) {
 
 IrcBridge.prototype.kill = Promise.coroutine(function*() {
     this._bridge.getBot().unsafeClient = null;
+    log.info("Bot client killed");
 });
 
 IrcBridge.prototype.isStartedUp = function() {
--- a/lib/irc/BridgedClient.js
+++ b/lib/irc/BridgedClient.js
@@ -178,11 +178,7 @@ BridgedClient.prototype.disconnect = function(reason) {
     if (!this.inst || this.inst.dead) {
         return Promise.resolve();
     }
-    var d = promiseutil.defer();
-    this.inst.disconnect(reason, function() {
-        d.resolve();
-    });
-    return d.promise;
+    return this.inst.disconnect(reason);
 };
 
 /**
--- a/lib/irc/BridgedClient.js
+++ b/lib/irc/BridgedClient.js
@@ -81,6 +81,7 @@ BridgedClient.prototype.kill = function() {
     // nullify references to connection instance client
     this.unsafeClient = null;
     // kill connection instance
+    log.info('Killing client ', this.nick);
     return this.disconnect("Bridged client killed");
 }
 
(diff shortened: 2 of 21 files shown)


lukebarnard1:
Actually, maybe I should make this a thing that BridgedClient does... and then use that here.

lukebarnard1:
Conclusion:

- Clients have a way of being killed (#74)
- The bridge now has a way to kill all clients
- The tests now kill the bridge if it has been created, prior to creating a new one, removing flaky tests (#166)

lib/bridge/IrcBridge.js
```diff
@@ -303,6 +303,15 @@ IrcBridge.prototype.run = Promise.coroutine(function*(port) {
     this._startedUp = true;
 });
 
+IrcBridge.prototype.kill = function() {
```
Kegsay: Docs please.

lib/bridge/IrcBridge.js
```diff
@@ -303,6 +303,15 @@ IrcBridge.prototype.run = Promise.coroutine(function*(port) {
...
+    if (!this._bridge.getBot()) {
+        log.info("No bot client to kill");
+        return Promise.resolve({});
```
Kegsay: This seems weird. Why should there being no bot affect the `ClientPool`? Why aren't we just directly calling `killAllClients()`?

lib/irc/ClientPool.js
```diff
@@ -30,6 +30,33 @@ function ClientPool(ircBridge) {
...
+ClientPool.prototype.killAllClients = function() {
+    var domainList = Object.keys(this._virtualClients);
+    var killList = domainList.map((domain) => {
+        let clients = Object.keys(this._virtualClients[domain].nicks).map(
+            (nick) => this._virtualClients[domain].nicks[nick]
+        );
+
+        clients = clients.concat(
+            Object.keys(this._virtualClients[domain].userIds).map(
+                (userId) => this._virtualClients[domain].userIds[userId]
```
Kegsay: Is this `concat` necessary? The nick => Client lookup and user_id => Client lookup should contain the same set of Clients.

lib/irc/ClientPool.js
```diff
@@ -30,6 +30,33 @@ function ClientPool(ircBridge) {
...
+            )
+        );
+
+        clients.push(this._botClients[domain]);
+
+        return clients.filter((c) => c !== undefined);
+    }).reduce((a, b) => {return a.concat(b)});
+
+    // console.log('Clients to kill ', killList.map((c)=>c.nick));
```
Kegsay: Debugging comment :)

lib/main.js
```diff
@@ -120,6 +121,15 @@ module.exports.runBridge = Promise.coroutine(function*(port, config, reg) {
...
     // run the bridge
-    var ircBridge = new IrcBridge(config, reg);
+    ircBridge = new IrcBridge(config, reg);
     yield ircBridge.run(port);
 });
+
+module.exports.killBridge = function() {
+    if (!ircBridge) {
+        log.info('killBridge(): No bridge running');
+        return Promise.resolve({});
```
Kegsay: `Promise.resolve()` would be clearer, else it looks like something is going to look at the result of this.

lib/main.js
```diff
@@ -120,6 +121,15 @@ module.exports.runBridge = Promise.coroutine(function*(port, config, reg) {
     }
 
     // run the bridge
-    var ircBridge = new IrcBridge(config, reg);
+    ircBridge = new IrcBridge(config, reg);
```
Kegsay: We should absolutely be guarding against clobbering the `ircBridge` global. Either immediately die if one already exists, or whine loudly in the logs, up to you.

spec/util/test.js
```diff
@@ -72,15 +73,25 @@ module.exports.beforeEach = function(testCase, env) {
...
+                    "@global": true
+                },
+                "matrix-js-sdk": env.clientMock,
+                "irc": env.ircMock
+            });
+            env.mockAppService = MockAppService.instance();
+        };
+
+        if (env.main) {
+            env.main.killBridge().then(done);
```
Kegsay: How are you not racing when doing this? `beforeEach` isn't returning a `Promise`, so how does a test know when it can continue setting up its environment?

lib/bridge/IrcBridge.js
lukebarnard1: Yep, that's not supposed to be there!

lib/irc/BridgedClient.js
```diff
@@ -77,6 +77,14 @@ BridgedClient.prototype.getClientConfig = function() {
     return this._clientConfig;
 };
 
+BridgedClient.prototype.kill = function() {
+    // nullify references to connection instance client
```
lukebarnard1: Option number 2. I didn't consider GCing because at the moment, the only reason for a client to be killed is if the entire bridge is killed. I'll change the comment :+1: 

Also, ```unsafeClient``` is mostly null guarded but sometimes not explicitly, with some functions relying on calling another and it succeeding and I think that may be a separate problem.

lib/irc/ClientPool.js
lukebarnard1: I feel like there should be a better data structure here then if there is a genuine redundancy. Also killing a killed client is a nop.

lib/irc/ClientPool.js
```diff
@@ -30,6 +30,33 @@ function ClientPool(ircBridge) {
...
+
+        clients = clients.concat(
+            Object.keys(this._virtualClients[domain].userIds).map(
+                (userId) => this._virtualClients[domain].userIds[userId]
+            )
+        );
+
+        clients.push(this._botClients[domain]);
+
+        return clients.filter((c) => c !== undefined);
```
lukebarnard1: Fair enough

lib/main.js
lukebarnard1: Yep, that's a mistake

spec/util/test.js
lukebarnard1: It seems to me that ```beforeEach``` should return a deferred promise, but this will mean changing all the tests to yield on it. Or I could put an optional callback in ```killBridge``` and give it ```done```?

Kegsay: The race isn't in `killBridge`, the race is `beforeEach`. Look at how the tests call it:
```js
    beforeEach(function(done) {
        test.beforeEach(this, env); // eslint-disable-line no-invalid-this

        sdk = env.clientMock._client(tUserId);
        // add registration mock impl:
        // registering should be for the irc user
        sdk._onHttpRegister({
            expectLocalpart: roomMapping.server + "_" + tFromNick,
            returnUserId: tUserId
        });

        env.ircMock._autoJoinChannels(
            roomMapping.server, roomMapping.botNick, roomMapping.server
        );
        env.ircMock._autoConnectNetworks(
            roomMapping.server, roomMapping.botNick, roomMapping.server
        );

        // do the init
        test.initEnv(env).done(function() {
            done();
        });
    });
```

They rely on `test.beforeEach(this, env);` setting up `env.main` and `env.mockAppService` so that `test.initEnv(env)` can [use them](https://github.com/matrix-org/matrix-appservice-irc/pull/205/files/3547516021a5d20e96ed9784cd79d7eeeebd3ea5#diff-593f55536362a50eb5956ca38a68c936R41). We aren't ever waiting for this to be set up in this code as it stands.

lib/bridge/IrcBridge.js
lukebarnard1: Done.

lib/bridge/IrcBridge.js
lukebarnard1: Done.

lib/irc/ClientPool.js
lukebarnard1: Done.

lib/irc/ConnectionInstance.js
```diff
@@ -74,8 +74,11 @@ ConnectionInstance.prototype.connect = function() {
  * @param {string} reason - Reason to reject with. One of:
  * throttled|irc_error|net_error|timeout|raw_error
  */
-ConnectionInstance.prototype.disconnect = function(reason) {
+ConnectionInstance.prototype.disconnect = function(reason, cb) {
```
lukebarnard1: Done.

spec/util/test.js
lukebarnard1: Done. That was painful, but now all tests yield on `beforeEach`and `test.initEnv`.

lib/irc/BridgedClient.js
```diff
@@ -77,6 +77,17 @@ BridgedClient.prototype.getClientConfig = function() {
     return this._clientConfig;
 };
 
+BridgedClient.prototype.kill = function() {
+    // Nullify so that no further commands can be issued
+    //  via unsafeClient, which should be null checked
+    //  anyway as it not instantiated until a connection
```
Kegsay: ` as it not` => ` as it is not`

lib/irc/BridgedClient.js
```diff
@@ -77,6 +77,17 @@ BridgedClient.prototype.getClientConfig = function() {
...
+    //  has occured.
```
Kegsay: `occured`

lib/irc/ConnectionInstance.js
```diff
@@ -95,6 +96,7 @@ ConnectionInstance.prototype.disconnect = function(reason) {
         // separate objects.
         this.onDisconnect();
     }
+    return Promise.resolve();
```
Kegsay: Maybe you meant to add a callback to `this.client.disconnect();`?

lib/main.js
```diff
@@ -119,7 +120,20 @@ module.exports.runBridge = Promise.coroutine(function*(port, config, reg) {
         config.homeserver = config.appService.homeserver;
     }
 
+    if (ircBridge) {
+        log.warn('Bridge already running, destroying reference to existing bridge!');
```
Kegsay: :heart:

lib/irc/BridgedClient.js
lukebarnard1: Done.

lib/irc/BridgedClient.js
lukebarnard1: Done.

lib/irc/ConnectionInstance.js
lukebarnard1: Done.

PR #207, opened 2016-09-30
Admin room name change and added topic

### LINKED ISSUE #198: IRC Admin Room -> Network-specific Name
Ideas:
- Need to have the first two words to distinguish between admin rooms.
- Probably shouldn't have the word "room" in the name, it's already a room.
- Maybe a topic should describe what it does!

Candidates:
- ~~Freenode Admin Room~~ includes word room
- ~~Freenode IRC Admin Room~~ includes word room
- ~~IRC Admin (Freenode)~~ can't distinguish with first two words
- ~~irc.freenode.net Connection Admin~~ too long, use topic instead
- **irc.freenode.net Admin, /w topic**

lukebarnard1:
Example:
 - Room name: 'irc.freenode.net Admin'
 - Room topic: 'This room shows any errors or status messages from irc.freenode.net, as well as letting you control the connection.'

Fixes #198

### DIFF
--- a/lib/bridge/IrcHandler.js
+++ b/lib/bridge/IrcHandler.js
@@ -444,8 +444,8 @@ IrcHandler.prototype.onMetadata = Promise.coroutine(function*(req, client, msg) 
             createAsClient: false,
             options: {
                 name: `${client.server.domain} Admin`,
-                topic:  "This room shows any errors or status messages from " +
-                        "${client.server.domain}, as well as letting you control " +
+                topic:  `This room shows any errors or status messages from ` +
+                        `${client.server.domain}, as well as letting you control ` +
                         "the connection. ",
                 preset: "trusted_private_chat",
                 visibility: "private",
--- a/lib/bridge/IrcHandler.js
+++ b/lib/bridge/IrcHandler.js
@@ -443,7 +443,10 @@ IrcHandler.prototype.onMetadata = Promise.coroutine(function*(req, client, msg) 
         let response = yield this.ircBridge.getAppServiceBridge().getIntent().createRoom({
             createAsClient: false,
             options: {
-                name: "IRC Admin Room",
+                name: `${client.server.domain} Admin`,
+                topic:  "This room shows any errors or status messages from " +
+                        "${client.server.domain}, as well as letting you control " +
+                        "the connection. ",
                 preset: "trusted_private_chat",
                 visibility: "private",
                 invite: [client.userId]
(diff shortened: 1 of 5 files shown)


lib/bridge/IrcHandler.js
```diff
@@ -443,7 +443,10 @@ IrcHandler.prototype.onMetadata = Promise.coroutine(function*(req, client, msg)
         let response = yield this.ircBridge.getAppServiceBridge().getIntent().createRoom({
             createAsClient: false,
             options: {
-                name: "IRC Admin Room",
+                name: `${client.server.domain} Admin`,
```
Kegsay: I like this. The only problem I see is that the domains may end up being ugly (think `ipv6.chat.freenode.net` or equiv). Should we instead prioritise the "description" field in the config or whatever it was? That would then be "Freenode".

Kegsay: Though "Freenode Admin" is also pretty confusing. Ugh, naming is hard.

lib/bridge/IrcHandler.js
```diff
@@ -443,7 +443,10 @@ IrcHandler.prototype.onMetadata = Promise.coroutine(function*(req, client, msg)
...
+                topic:  "This room shows any errors or status messages from " +
+                        "${client.server.domain}, as well as letting you control " +
```
lukebarnard1: oh.. I thought I had, I'll test again

lib/bridge/IrcHandler.js
lukebarnard1: "Admin" sounds like you are the administrator for the network... Or at least in a room that connects you to an admin user.

I had other suggestions in #198

Kegsay: How about keeping it more informal perhaps, and something like "Your Freenode Connection"?

Kegsay: Really need @ara4n to chime in on this.

Kegsay:
LGTM aside from naming fun times!

lib/irc/IrcServer.js
```diff
@@ -17,6 +17,25 @@ function IrcServer(domain, serverConfig) {
     this.config = serverConfig;
 }
 
+/**
+ * Get a string that represents the human-readable name for a server.
+ * @returns config.name if available, otherwise it will attempt to
```
Kegsay: If you're going to use JSDoc then please make it valid JSDoc:
```js
/**
 * @return {string} config.name blah blah
 */
```

Kegsay:
Can we deprecate `description` and return `name` in the 3P lookup responses as well then please?

Kegsay:
Code-wise, LGTM

lukebarnard1:
Done. Unfortunately, this will mean changing the descriptions of all the bridges unless you want admin rooms with names like `Freenode (irc.freenode.net) IRC Bridge status` /o\

PR #216, opened 2016-10-03
Map +s to room visibility

lukebarnard1:
[-|+]s maps to setRoomVisibility ['private'|'public']

### DIFF
--- a/lib/bridge/IrcHandler.js
+++ b/lib/bridge/IrcHandler.js
@@ -390,49 +390,51 @@ IrcHandler.prototype.onMode = Promise.coroutine(function*(req, server, channel, 
         channel, by, arg
     );
 
-    // redundant if statement currently but eases burden when adding
-    // support for more modes
-    if (privateModes.indexOf(mode) !== -1) {
-        // 'k' = Channel requires 'keyword' to join.
-        // 'i' = Channel is invite-only.
-        // 's' = Channel is secret
+    // 'k' = Channel requires 'keyword' to join.
+    // 'i' = Channel is invite-only.
+    // 's' = Channel is secret
 
-        // For k and i, we currently want to flip the join_rules to be
-        // 'invite' to prevent new people who are not in the room from
-        // joining.
+    // For k and i, we currently want to flip the join_rules to be
+    // 'invite' to prevent new people who are not in the room from
+    // joining.
 
-        // For s, we just want to control the room directory visibility
-        // accordingly. (+s = 'private', -s = 'public')
+    // For s, we just want to control the room directory visibility
+    // accordingly. (+s = 'private', -s = 'public')
 
-        // TODO: Add support for specifying the correct 'keyword' and
-        // support for sending INVITEs for virtual IRC users.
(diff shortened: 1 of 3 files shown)


lib/bridge/IrcHandler.js
```diff
@@ -390,12 +392,18 @@ IrcHandler.prototype.onMode = Promise.coroutine(function*(req, server, channel,
 
     // redundant if statement currently but eases burden when adding
     // support for more modes
-    if (["k", "i"].indexOf(mode) !== -1) {
+    if (privateModes.indexOf(mode) !== -1) {
```
leonerd: I imagined the point of this line was to be able to add other lines later on. So rather than putting the different `+s` handling logic inside this if block, add a new one later?

lib/bridge/IrcHandler.js
```diff
@@ -379,47 +379,85 @@ IrcHandler.prototype.onPart = Promise.coroutine(function*(req, server, leavingUs
...
+ * @param {Request} req The metadata request
+ * @param {IrcServer} server : The sending IRC server.
+ * @param {string} channel The channel that has the given mode.
+ * @param {string} mode The mode that the channel is in, e.g. +sabcdef
+ * @return {Promise} which is resolved/rejected when the request finishes.
+ */
+IrcHandler.prototype.onModeIs = Promise.coroutine(function*(req, server, channel, mode) {
+    req.log.info(`onModeIs for ${channel} = ${mode}.`);
+
+    mode = mode.slice(1); // Remove '+'
```
leonerd: For robustness I'd consider instead trimming a leading `+` if one exists, using a regexp. I could imagine some silly ircd somewhere not bothering with a leading `+` and so we'll ignore the first mode flag.

Maybe

    if (mode.match(/^\+/)) mode = mode.slice(1);

leonerd:
LGTM aside one remaining comment.
Though you probably want to update the description in the PR itself.

lib/bridge/IrcHandler.js
lukebarnard1: I'd prefer to just leave the string as is and ignore '+' as a mode

lukebarnard1: Done

lib/bridge/IrcHandler.js
```diff
@@ -448,11 +448,12 @@ IrcHandler.prototype.onMode = Promise.coroutine(function*(req, server, channel,
 IrcHandler.prototype.onModeIs = Promise.coroutine(function*(req, server, channel, mode) {
     req.log.info(`onModeIs for ${channel} = ${mode}.`);
 
-    mode = mode.slice(1); // Remove '+'
-
     // Delegate to this.onMode
     let promises = mode.split('').map(
         (modeChar) => {
+            if (modeChar !== '+') {
```
leonerd: You almost certainly mean `===` here. You want to ignore `+` and fallthrough to handling everything else.

lukebarnard1: duh, oops

lukebarnard1: ok, so I'll merge once the tests pass

PR #219, opened 2016-10-03
Request channel mode on ANY join

lukebarnard1:
This is so that +s is synced whenever the bridge has access to the channel mode. This is probably overkill and should only be done when the frontier user has joined the channel (because the bot might be disabled).


lukebarnard1:
Redundant after #225

PR #221, opened 2016-10-04
Cache the visibilty of a room last sent to mx

lukebarnard1:
This is so that the HS is not hammered on startup when a lot of RPL_CHANMODEIS _could_ be received as clients join channels.

The problem with this is, what happens when two channels are mapped to one room and just one of the rooms is +s or +i/k? See issue #223


Kegsay:
LGTM

Kegsay:
LGTM

PR #225, opened 2016-10-05
WIP of +s conflict checking

lukebarnard1:
It is currently possible still for +s channels to leak as public matrix rooms. So this fix goes to the extreme of full graph traversal of all mappings between channels and rooms and corrects discrepancies.

When any change to the cache of '+s' for channels or private/public for rooms occurs, the conflicts are resolved by keeping the matrix side as private as possible (by set[ing]RoomDirectoryVisibility()). This resolution only occurs when 10 seconds has elapsed where no changes have occurred. On top of this, it may also be useful to have a minimum interval where it is always updated (provided there have been changes).

TODO:

- [ ] Listen for room visibility from matrix somehow

### DIFF
--- a/lib/DataStore.js
+++ b/lib/DataStore.js
@@ -120,6 +120,37 @@ DataStore.prototype.getRoom = function(roomId, ircDomain, ircChannel, origin) {
 };
 
 /**
+ * Get all Matrix <--> IRC room mappings from the database.
+ * @return {Promise} A promise which resolves to a map:
+ *      $roomId => ['server #channel1', 'server #channel2',...]
+ */
+DataStore.prototype.getAllMappings = Promise.coroutine(function*() {
+    let entries = yield this._roomStore.select(
+        {
+            matrix_id : {$exists: true},
+            remote_id: {$exists: true},
+        }
+    );
+
+    let uniqueFilter = (value, index, self) => {
+        // Returns true if this is the first occurance of value
+        return self.indexOf(value) === index;
+    };
+
+    let roomIds = entries.map((e) => e.matrix_id).filter(uniqueFilter);
+
+    let mappings = {};
+
+    roomIds.forEach((roomId) => {
+        mappings[roomId] = entries.filter(
+            (e) => e.matrix_id === roomId
+        ).map((e) => e.remote.domain + ' ' + e.remote.channel);
+    });
+
+    return mappings;
+});
+
+/**
  * Get provisioned IRC <--> Matrix room mappings from the database where
  * the matrix room ID is roomId.
  * @param {string} roomId : The Matrix room ID.
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -102,6 +102,21 @@ function IrcBridge(config, registration) {
     );
 
     this._provisioner = null;
+
+    // Cache the mode of each channel, the visibility of each room and the
+    // known mappings between them. When any of these change, any inconsistencies
+    // should be resolved by keeping the matrix side as private as necessary
+    this._visibilityMap = {
+        mappings: {
+            //room_id: ['server #channel1', 'server channel2',...]
+        },
+        channelModes: {
+            // '$server $channel': true | false
+        },
+        roomVisibilities: {
+            // room_id: "private" | "public"
+        }
+    }
 }
 
 IrcBridge.prototype.getAppServiceUserId = function() {
@@ -753,6 +768,128 @@ IrcBridge.prototype.getBotClient = function(server) {
         return this._clientPool.getBot(server);
     });
 }
+
+
+var _solveTimeout = null;
+
+IrcBridge.prototype._updateVisibilityMap = function(isMode, key, value) {
+    let hasChanged = false;
+    if (isMode) {
+        if (typeof value !== 'boolean') {
+            throw new Error('+s state must be indicated with a boolean');
+        }
+        if (this._visibilityMap.channelModes[key] !== value) {
+            this._visibilityMap.channelModes[key] = value;
+            hasChanged = true;
+        }
+    }
+    else {
+        if (typeof value !== 'string' || (value !== "private" && value !== "public")) {
+            throw new Error('Room visibility must = "private" | "public"');
+        }
+
+        if (this._visibilityMap.roomVisibilities[key] !== value) {
+            this._visibilityMap.roomVisibilities[key] = value;
+            hasChanged = true;
+        }
+    }
+
+    if (hasChanged) {
+        clearTimeout(_solveTimeout);
+
+        _solveTimeout = setTimeout(() => {
+            this._solveVisibility()
+        }, 10000);
+        return Promise.resolve();
+    }
+
+    return Promise.resolve();
+}
+
+IrcBridge.prototype._solveVisibility = Promise.coroutine(function*() {
+    // For each room, do a big OR on all of the channels that are linked in any way
+
+    this._visibilityMap.mappings = yield this.getStore().getAllMappings();
+
+    let roomIds = Object.keys(this._visibilityMap.roomVisibilities);
+
+    let checkedChannels = [];
+
+    let shouldBePrivate = (roomId) => {
+        // If any channel connected to this room is +s, stop early and call it private
+
+        // List first connected
+        let channels = this._visibilityMap.mappings[roomId];
+        //      = ['localhost #channel1', 'localhost #channel2', ... ]
+
+        // Filter out already checked channels
+        channels = channels.filter((c) => checkedChannels.indexOf(c) === -1);
+
+        // No channels mapped to this roomId
+        if (!channels) {
+            return false;
+        }
+
+        let anyAreSecret = channels.any((channel) => {
+            let channelIsSecret = this._visibilityMap.channelModes[channel];
+
+            // If a channel mode is unknown, assume it is secret
+            if (typeof channelIsSecret === 'undefined') {
+                channelIsSecret = true;
+            }
+
+            return channelIsSecret;
+        });
+        if (anyAreSecret) {
+            return true;
+        }
+
+        checkedChannels = checkedChannels.concat(channels);
+
+        // Otherwise, recurse with the rooms connected to each channel
+
+        // So get all the roomIds that this channel is mapped to
+        let connectedRooms = channels.map((channel) => {
+            return Object.keys(this._visibilityMap.mappings).filter((roomId2) => {
+                return this._visibilityMap.mappings[roomId2].indexOf(channel) !== -1;
+            });
+        });
+
+        return connectedRooms.any(shouldBePrivate);
+    }
+
+    let roomCorrectVisibilities = roomIds.map(
+        shouldBePrivate
+    ).map((b) => b ? 'private' : 'public');
+
+    let roomsShouldBePrivate = roomIds.filter((roomId, index) => {
+        let isCurrentlyPrivate = this._visibilityMap.roomVisibilities[roomId] === 'private';
+        let roomShouldBePrivate = roomCorrectVisibilities[index] === 'private';
+
+        return !isCurrentlyPrivate && roomShouldBePrivate;
+    });
+
+    let roomsShouldBePublic = roomIds.filter((roomId, index) => {
+        let isCurrentlyPrivate = this._visibilityMap.roomVisibilities[roomId] === 'private';
+        let roomShouldBePublic = roomCorrectVisibilities[index] === 'public';
+
+        return isCurrentlyPrivate && roomShouldBePublic;
+    });
+
+    // Update rooms to correct visibilities
+    let cli = this.getAppServiceBridge().getClientFactory().getClientAs();
+
+    let promises = roomsShouldBePrivate.map((roomId) => {
+        return cli.setRoomDirectoryVisibility(roomId, "private");
+    });
+
+    promises = promises.concat(roomsShouldBePublic.map((roomId) => {
+        return cli.setRoomDirectoryVisibility(roomId, "public");
+    }));
+
+    return Promise.all(promises);
+});
+
 
 IrcBridge.DEFAULT_LOCALPART = "appservice-irc";
 
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -813,9 +813,8 @@ IrcBridge.prototype._solveVisibility = Promise.coroutine(function*() {
 
     let roomIds = Object.keys(this._visibilityMap.roomVisibilities);
 
-    let checkedChannels = [];
-
-    let shouldBePrivate = (roomId) => {
+    // level is the current depth of recursion
+    let shouldBePrivate = (roomId, checkedChannels) => {
         // If any channel connected to this room is +s, stop early and call it private
 
         // List first connected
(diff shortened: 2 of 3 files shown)


lib/bridge/IrcBridge.js
```diff
@@ -102,6 +102,21 @@ function IrcBridge(config, registration) {
     );
 
     this._provisioner = null;
+
+    // Cache the mode of each channel, the visibility of each room and the
+    // known mappings between them. When any of these change, any inconsistencies
+    // should be resolved by keeping the matrix side as private as necessary
+    this._visibilityMap = {
```
Kegsay: So I've gone through all the dependencies you're using in order to do this, and it's just:
 - A `DataStore`
 - A bot `MatrixClient`

Can we please yank all of this into its own file and dependency inject, so I don't have to constantly check to see if you're using other fun aspects of the `IrcBridge`? This would make it a lot clearer to follow, and avoids polluting `IrcBridge` more than it already is (poor thing).

lib/bridge/IrcBridge.js
```diff
@@ -102,6 +102,21 @@ function IrcBridge(config, registration) {
...
+        mappings: {
+            //room_id: ['server #channel1', 'server channel2',...]
+        },
+        channelModes: {
+            // '$server $channel': true | false
```
Kegsay: I haven't gone through the bulk of the code yet, I'm just reading the core data structure currently, and this jumps out at me. In order to see if a channel is `+s`, I bet we're going to do:
```js
if (this._visibilityMap.channelModes["irc.freenode.net #chan"]) {
  // stuff
}
```
I don't like that because it isn't clear that this being `true` means the channel is `+s`. Can we rename this to `channelIsSecret` or something instead of the really generic `channelModes` (of which it doesn't actually represent the set of channel modes, just a very specific one, `+s`). This would mean the equiv is:
```js
if (this._visibilityMap.channelIsSecret["irc.freenode.net #chan"]) {
  // stuff
}
```
which is then immediately clear what is going on.

lib/bridge/IrcBridge.js
```diff
@@ -754,6 +769,128 @@ IrcBridge.prototype.getBotClient = function(server) {
     });
 }
 
+
+var _solveTimeout = null;
```
Kegsay: What is this? I'm guessing a timeout ID? Or is it a time to wait? Or maybe a Promise? Please comment and/or give a clearer variable name. (quick grep at the code here looks like this is a [timeout ID](https://developer.mozilla.org/en-US/docs/Web/API/WindowTimers/clearTimeout))

lib/bridge/IrcHandler.js
```diff
@@ -409,11 +409,13 @@ IrcHandler.prototype.onMode = Promise.coroutine(function*(req, server, channel,
         return;
     }
 
+    if (mode === "s") {
+        // Update the visibility for all rooms connected to this channel
+        return this.ircBridge.updateVisibilityMap(true, server.domain + ' ' + channel, enabled);
```
Kegsay: How does this not crash at runtime? The method is `_updateVisibilityMap` not `updateVisibilityMap`.

lib/bridge/IrcBridge.js
lukebarnard1: Sure np

lib/bridge/IrcHandler.js
lukebarnard1: It does crash at run-time, hence WIP

lukebarnard1: (I've fixed it now though)

lib/bridge/IrcBridge.js
```diff
@@ -754,6 +769,128 @@ IrcBridge.prototype.getBotClient = function(server) {
...
+            hasChanged = true;
+        }
+    }
+
+    if (hasChanged) {
+        clearTimeout(_solveTimeout);
+
+        _solveTimeout = setTimeout(() => {
+            this._solveVisibility()
+        }, 10000);
```
Kegsay: What is the purpose of maintaining a timer here? I'm guessing this is to avoid the risk of the bridge wedging entirely on large graphs (or maybe we aren't doing cycle detection so could loop forevaaaar?). A comment to state as much would be good, and given you're returning a Promise in this function, I would be tempted to use `Promise.timeout()` rather than wrangling timeout IDs and `setTimeout` yourself.

Kegsay: I'm also not sure how much we want to *always* assign-then-solve here. When the bridge starts up, it could be getting thousands of these guys, it may be better to separate the calls here (so `_updateVisibilityMap` returns `hasChanged`), and let the caller decide if they want to solve (which they may only do after startup, after they have processed a block of users, whatever).

lukebarnard1: We are doing cycle detection (by keeping a list of all channels checked in the stack, I'll push this in a bit). The reason this is used is so that if it is called again, the timeout won't fire. It resets the timer, meaning if we get 100 RPL_CHANMODEIS from 100 IRC clients joining, it will reset the timer 100 times, and then once 10 seconds of silence has elapsed, it will finally run.

If you think this behaviour is ok then I shall document it.

lib/bridge/IrcBridge.js
```diff
@@ -754,6 +769,128 @@ IrcBridge.prototype.getBotClient = function(server) {
...
+        return Promise.resolve();
+    }
+
+    return Promise.resolve();
+}
+
+IrcBridge.prototype._solveVisibility = Promise.coroutine(function*() {
+    // For each room, do a big OR on all of the channels that are linked in any way
```
Kegsay: Obviously we both know what we mean by "solve visibility" but the rest of the team won't. I would go in to a decent amount of detail explaining what the problem is, why you need to solve it, and how this function solves it. As a very terse example:
```js
// Walk the graph of mapped rooms to work out if a Matrix room should be on the published room list or not.
//
// We need to do this in order to ensure that messages sent on a +s (secret) IRC channel do not leak out into rooms/channels which ARE on the published room list.
// If a Matrix room is connected to other IRC channels, if ANY single channel is `+s` then remove this Matrix room from the published room list.
// If ALL channels are NOT `+s`, then publish this Matrix room on the published room list.
IrcBridge.prototype._solveVisibility = Promise.coroutine(function*() {
  // ...
}
```

lib/bridge/IrcBridge.js
```diff
@@ -754,6 +769,128 @@ IrcBridge.prototype.getBotClient = function(server) {
...
+        // No channels mapped to this roomId
+        if (!channels) {
+            return false;
+        }
+
+        let anyAreSecret = channels.any((channel) => {
+            let channelIsSecret = this._visibilityMap.channelModes[channel];
+
+            // If a channel mode is unknown, assume it is secret
+            if (typeof channelIsSecret === 'undefined') {
```
Kegsay: This is an interesting one: should we be doing this? It's the safe bet, but most likely this will be `undefined` on startup as the modes slowly trickle in. Should we be bailing from modifying the visibility at all if this is not set?

lib/bridge/IrcBridge.js
```diff
@@ -754,6 +769,128 @@ IrcBridge.prototype.getBotClient = function(server) {
...
+
+        // Filter out already checked channels
+        channels = channels.filter((c) => checkedChannels.indexOf(c) === -1);
+
+        // No channels mapped to this roomId
+        if (!channels) {
+            return false;
+        }
+
+        let anyAreSecret = channels.any((channel) => {
```
Kegsay: I don't see `any()` on MDN, and Node v4 also thinks this does not exist:
```js
> s = []
[]
> s.any
undefined
```

lib/bridge/IrcBridge.js
```diff
@@ -754,6 +769,128 @@ IrcBridge.prototype.getBotClient = function(server) {
...
+        let roomShouldBePublic = roomCorrectVisibilities[index] === 'public';
+
+        return isCurrentlyPrivate && roomShouldBePublic;
+    });
+
+    // Update rooms to correct visibilities
+    let cli = this.getAppServiceBridge().getClientFactory().getClientAs();
+
+    let promises = roomsShouldBePrivate.map((roomId) => {
+        return cli.setRoomDirectoryVisibility(roomId, "private");
```
Kegsay: The fun bit is that this will race a lot because we're rescanning the entire set of mappings every single time `_solveVisibility` is called, and this is currently triggered on *any* mode change. This will result in spamming Matrix a lot, even *with* the cache (because there is a window from determining "I should change the directory visibility" to "I have received the new directory visibility and updated the cache").

lib/bridge/IrcBridge.js
lukebarnard1: I think it's best to keep everything private unless we know it is definitely not private, otherwise we're going to get more issues later on with race conditions causing privacy leaks.

lib/bridge/IrcBridge.js
```diff
@@ -754,6 +769,128 @@ IrcBridge.prototype.getBotClient = function(server) {
...
+        }).any((roomId2) => {
+            return shouldBePrivate(roomId2, checkedChannels.concat(channels));
+        });
+    }
+
+    let roomCorrectVisibilities = roomIds.map(
+        (roomId) => {
+            return shouldBePrivate(roomId, []);
+        }
+    ).map((b) => b ? 'private' : 'public');
```
lukebarnard1: Fair enough!

lib/bridge/IrcBridge.js
lukebarnard1: I could add an extra similar mechanism that can send visibility if they don't change in the next min or something?

lib/DataStore.js
```diff
@@ -120,6 +120,37 @@ DataStore.prototype.getRoom = function(roomId, ircDomain, ircChannel, origin) {
...
+ * Get all Matrix <--> IRC room mappings from the database.
+ * @return {Promise} A promise which resolves to a map:
+ *      $roomId => ['server #channel1', 'server #channel2',...]
+ */
+DataStore.prototype.getAllMappings = Promise.coroutine(function*() {
+    let entries = yield this._roomStore.select(
+        {
+            matrix_id : {$exists: true},
+            remote_id: {$exists: true},
+        }
```
Kegsay: Whilst this does what it says on the tin, this is almost certainly not what you want. This will pick up PM rooms which look like:
```json
{
"id":"PM_@kegan:matrix.org @mozilla_ffwywvmervtdliqphiabshsmibvfns:matrix.org",
"remote_id":"irc.mozilla.org ffwywvmervtdliqphiabshsmibvfns",
"matrix_id":"!gDHoKrNsohmsFydWRb:matrix.org",
"remote":{"domain":"irc.mozilla.org","type":"pm","channel":"ffwywvmervtdliqphiabshsmibvfns"},
"matrix":{"extras":{}},
"data":{"virtual_user_id":"@mozilla_ffwywvmervtdliqphiabshsmibvfns:matrix.org","real_user_id":"@kegan:matrix.org"},
"_id":"3z5eJBIRz6f8d8FK"
}
```

What you really want is to pick out all *channels*:
```json
{
"id":"!PcCidzjUKXHKImhTrB:matrix.org irc.mozilla.org #rust",
"remote_id":"irc.mozilla.org #rust",
"matrix_id":"!PcCidzjUKXHKImhTrB:matrix.org",
"remote":{"domain":"irc.mozilla.org","channel":"#rust","type":"channel"},
"matrix":{"extras":{}},
"data":{"origin":"config"},
"_id":"cmQ0M7lHHAZg76nQ"
}
```

AKA `remote.type === "channel"`

lib/DataStore.js
```diff
@@ -120,6 +120,37 @@ DataStore.prototype.getRoom = function(roomId, ircDomain, ircChannel, origin) {
...
+    );
+
+    let uniqueFilter = (value, index, self) => {
+        // Returns true if this is the first occurance of value
+        return self.indexOf(value) === index;
+    };
+
+    let roomIds = entries.map((e) => e.matrix_id).filter(uniqueFilter);
```
lukebarnard1: Done.

lib/bridge/IrcBridge.js
lukebarnard1: Yep, done.

lib/DataStore.js
lukebarnard1: Done

lib/bridge/IrcBridge.js
lukebarnard1: Damn, forgot to do this one /me corrects...

lib/bridge/IrcBridge.js
lukebarnard1: Done

lib/bridge/IrcBridge.js
Kegsay: The 10s delay on recalcualting visibilities will be Good Enough for now, given time constraints.

lib/bridge/PublicitySyncer.js
```diff
@@ -0,0 +1,157 @@
...
+        if (this._visibilityMap.roomVisibilities[key] !== value) {
+            this._visibilityMap.roomVisibilities[key] = value;
+            hasChanged = true;
+        }
+    }
+
+    if (hasChanged) {
+        clearTimeout(solveVisibilityTimeoutId);
+
+        log.info('Solving in 10s');
```
Kegsay: Remove log line please.

lib/bridge/PublicitySyncer.js
```diff
@@ -0,0 +1,157 @@
...
+   and rooms 'visibility' states. This does full graph traversal to prevent any +s
+   channels ever escaping into a 'public' room. This function errs on the side of
+   caution by assuming an unknown channel state is '+s'. This just means that if the
+   modes of a channel are not received yet (e.g when no virtual user is in said channel)
+   then the room is assumed secret (+s).
+
+   The bare minimum is done to make sure no private channels are leaked into public
+   matrix channels. If ANY +s channel is somehow being bridged into a room, that room
+   is updated to private. If ALL channels somehow being bridged into a room, that
+   room is allowed to be public.
```
Kegsay: Missing a few words? "If ALL channels being bridged into a room are NOT +s, that room is allowed to be public."

lib/bridge/PublicitySyncer.js
```diff
@@ -0,0 +1,157 @@
...
+                return shouldBePrivate(roomId2, checkedChannels.concat(channels));
+            });
+        });
+    }
+
+    let roomCorrectVisibilities = roomIds.map(
+        (roomId) => {
+            return shouldBePrivate(roomId, []);
+        }
+    ).map((b) => b ? 'private' : 'public');
```
Kegsay: I still don't like how we index into this array on `:142`. Can we not just put:
```js
let correctState = shouldBePrivate(roomId, []) ? "private": "public";
```
on `:142`? One less loop and it's clearer.

Kegsay:
Otherwise, this is looking good! Does it work?! LGTM after comments.

lukebarnard1:
It works in the slightly contrived case and the typical case, so maybe?..

lib/bridge/PublicitySyncer.js
```diff
@@ -0,0 +1,157 @@
+/*eslint no-invalid-this: 0*/ // eslint doesn't understand Promise.coroutine wrapping
+"use strict";
+var Promise = require("bluebird");
+var log = require("../logging").get("PublicitySyncer");
+
+function PublicitySyncer(ircBridge) {
```
lukebarnard1: Done.

lib/bridge/PublicitySyncer.js
lukebarnard1: Done

lib/bridge/PublicitySyncer.js
```diff
@@ -0,0 +1,157 @@
...
+        }
+    }
+
+    if (hasChanged) {
+        clearTimeout(solveVisibilityTimeoutId);
+
+        log.info('Solving in 10s');
+        solveVisibilityTimeoutId = setTimeout(() => {
+            this._solveVisibility().catch((err) => {
+                log.error(err.message);
```
lukebarnard1: Done.

lib/bridge/PublicitySyncer.js
lukebarnard1: Done.

lib/bridge/PublicitySyncer.js
```diff
@@ -0,0 +1,157 @@
...
+        let currentState = this._visibilityMap.roomVisibilities[roomId];
+        let correctState = roomCorrectVisibilities[index];
+
+        if (currentState !== correctState) {
+            return cli.setRoomDirectoryVisibility(roomId, correctState).then(
+                () => {
+                    // Update cache
+                    this._visibilityMap.roomVisibilities[roomId] = correctState;
+                }
+            ).catch((e) => console.err);
```
lukebarnard1: Done.

Kegsay:
What can possibly go wrong? LGTM!

lukebarnard1:
Oh god flaky tests

Kegsay:
We should also be checking the value of https://github.com/matrix-org/matrix-appservice-irc/blob/master/config.sample.yaml#L85 - if it is `false` then we don't need to do any of this (they will be initially not published and we don't ever need to change that).

Kegsay:
LGTM

PR #226, opened 2016-10-05
Include "field_types" metadata in 3PE protocol query result

leonerd:
If this can be in before Freenode's restart tomorrow that would be most helpful for @dbkr

### DIFF
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -450,6 +450,20 @@ IrcBridge.prototype.getThirdPartyProtocol = function(protocol) {
     return Promise.resolve({
         user_fields: ["domain", "nick"],
         location_fields: ["domain", "channel"],
+        field_types: {
+            domain: {
+                regexp: "[a-z0-9-_]+(\.[a-z0-9-_]+)*",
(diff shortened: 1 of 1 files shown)


lib/bridge/IrcBridge.js
```diff
@@ -448,9 +448,23 @@ IrcBridge.prototype.getThirdPartyProtocol = function(protocol) {
...
+            domain: {
+                regexp: "[a-z0-9-_]+(\.[a-z0-9-_]+)*",
+                placeholder: "irc.example.com",
+            },
+            nick: {
+                regexp: "[^#\\s]+",
+                placeholder: "SomeNick",
+            },
+            channel: {
+                regexp: "[#&+][^\\s]+",
```
Kegsay: (The bridge doesn't really *do* `&chan` and `+chan` rooms, so I would cheekily request you remove that from the `regexp` :angel: )

Kegsay:
Otherwise LGTM - merge when done ;3

lib/bridge/IrcBridge.js
leonerd: Done 99efdf7

PR #227, opened 2016-10-05
!join creates public rooms

### LINKED ISSUE #224: Joining channels via !join alters join rules
We've had a few cases where people are getting 403 Forbidden when trying to join `#freenode_` rooms. The `join_rules` state `invite` in this case, which is why they are getting 403d. They shouldn't ever be 403. I noticed this in one room today, and the person who joined the channel was doing so via `!join`. The `!join` command exists to make a user join a channel which is **invite-only** (since we nominally expect people to just join the room alias). I suspect the logic is a bit broken and is causing the room to be invite-only if the first(?) person to join it does so via `!join`.

lukebarnard1:
Should fix #224

### DIFF
--- a/lib/bridge/MatrixHandler.js
+++ b/lib/bridge/MatrixHandler.js
@@ -345,8 +345,28 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
             ).createRoom({
                 options: {
                     name: ircChannel,
-                    visibility: ircServer.getJoinRule()
-                    // Intent will automatically invite the user
+                    visibility: (
+                        ircServer.shouldPublishRooms() ? "public" : "private"
+                    ),
+                    creation_content: {
+                        "m.federate": ircServer.shouldFederate()
+                    },
+                    initial_state: [
+                        {
+                            type: "m.room.join_rules",
+                            state_key: "",
+                            content: {
+                                join_rule: ircServer.getJoinRule()
+                            }
+                        },
+                        {
+                            type: "m.room.history_visibility",
+                            state_key: "",
+                            content: {
+                                history_visibility: "joined"
+                            }
+                        }
+                    ]
                 }
             });
             let mxRoom = new MatrixRoom(response.room_id);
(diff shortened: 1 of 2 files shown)


lukebarnard1:
And the tests are making sure the created rooms are in-fact private...

Kegsay:
Yes because that is the desired behaviour when the rooms are not meant to be shared. We need to conditionally set them to public if the channel is allowed to be shared.

Kegsay:
The code *should* be inspecting https://github.com/matrix-org/matrix-appservice-irc/blob/master/config.sample.yaml#L91 but it isn't.

lukebarnard1:
That failed PR merge test seems to be caused by flaky tests... I'll try merging and seeing if I can make it a better merge and then push it here.

lib/bridge/MatrixHandler.js
```diff
@@ -345,7 +345,7 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
             ).createRoom({
                 options: {
                     name: ircChannel,
-                    visibility: "private"
+                    visibility: ircServer.getJoinRule()
```
Kegsay: Join rule != visibility. Copy what we do on [alias creation](https://github.com/matrix-org/matrix-appservice-irc/blob/bdeebc0f81636037016adf6bd16224293d10cf66/lib/bridge/MatrixHandler.js#L1072):
```js
            let res = yield botIntent.createRoom({
                options: {
                    name: channelInfo.channel,
                    visibility: (
                        channelInfo.server.shouldPublishRooms() ? "public" : "private"
                    ),
                    creation_content: {
                        "m.federate": channelInfo.server.shouldFederate()
                    },
                    initial_state: [
                        {
                            type: "m.room.join_rules",
                            state_key: "",
                            content: {
                                join_rule: channelInfo.server.getJoinRule()
                            }
                        },
                        {
                            type: "m.room.history_visibility",
                            state_key: "",
                            content: {
                                history_visibility: "joined"
                            }
                        }
                    ]
                }
            });
```

lukebarnard1: :+1:

Kegsay:
LGTM

PR #231, opened 2016-10-07
Allow setting a custom CA

Waldteufel:
In some configurations it's useful to trust a single CA only. This allows setting the 'ca' parameter from the configuration file.

### DIFF
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -30,6 +30,12 @@ ircService:
       ssl: true
       # Whether or not IRC server is using a self-signed cert or not providing CA Chain
       sslselfsign: false
+      # A specific CA to trust instead of the default CAs. Optional.
+      #ca: |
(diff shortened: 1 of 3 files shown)


config.sample.yaml
```diff
@@ -30,6 +30,12 @@ ircService:
       ssl: true
       # Whether or not IRC server is using a self-signed cert or not providing CA Chain
       sslselfsign: false
+      # A specific CA to trust instead of the default CAs. Optional.
+      ca: |
+        -----BEGIN CERTIFICATE-----
+        ...
+        -----END CERTIFICATE-----
```
Kegsay: Can you wrap these in `#` comments please? Otherwise, if people blindly `cp config.sample.yaml config.yaml` they will use this as their `ca`!

Kegsay:
LGTM aside from comment.

Kegsay:
LGTM! Thank you!

Issue #100, opened 2016-07-15
Bridge topics


Kegsay:
We don't really do this currently because of the semantics between "should we use the matrix topic" or "should we use the IRC topic".  I propose:
 - Add a new config section:

   ```yaml
   # Controls how topics should be set between Matrix/IRC
   topics:
      # Use the topics set in these rooms instead of any IRC one
      useRooms: ["!foo:bar", "!baz:quuz"]
      # Use the topics set in these channels instead of any IRC one
      useChannels: ["#chan", "#chan2"]
      # If neither room nor channel match the lists above, use the topic from irc|matrix
      default: "irc"
   ```

Edge cases:
 - If a room matches both `useChannels` and `useRooms` then the topic will not be bridged at all (current behaviour).
 - If syncing M->I topics and the virtual IRC client fails to set the topic (not ops), then [1]
 - If syncing I->M topics and the virtual Matrix user fails to set the topic (no power), **and** the bridge bot cannot grant the user power, then [1]
 - [1] = Do one of the following:
    * [PM] : Send a **PM** from the bridge bot to the originator of the topic to say `ERROR: Failed to bridge topic.`
    * [SRC] : Send a message into the **source room** as the bridge bot (who may not be joined) to say `ERROR: Failed to bridge topic`.
    * [DST] : Send a message into the **destination room** as the virtual user to say `ERROR: I can't set the topic`.
    * [NOP] : Silently fail.

My feeling is probably [PM] but this is horrible because often the topic is set by an individual who may not be in the room anymore or may not really contribute/follow the room much at all. In this case, PMs will be sent to them to say "waaah waaaah I can't bridge! SADPANDA!" which feels really horrible.

Rationale:
 - Not part of explicit `mappings` because you may want a specific dynamically bridged channel to use a different topic.
 - Whilst the bug is mainly to do with bridging IRC topics to Matrix, also including the inverse for symmetry.

eternaleye:
One thing I'm unclear on: Why "should" we use one or the other statically? To expand:

- On Matrix, the bridge can see the history of changes retrospectively
- On IRC, as long as the bridge is connected, it can observe changes as they happen
- The bridge, being a single node, is essentially free to presume linear time

As a result, unless connectivity with IRC is lost, the bridge can indeed safely mirror in both directions.

For dynamic bridging, I suspect the only sensible option is that on room instantiation, immediately copy the IRC-side topic, and then begin mirroring.

[...]

PR #243, opened 2016-10-17
Bridge topics

lukebarnard1:
### Add simple I->M topic bridging
When the IRC topic changes, any mappings that have an origin of 'join' or 'alias' will be used to update the related matrix rooms topics.

### Cache topics that have already been sent into a matrix room
Prevent duplicate topics being sent into a room, e.g. on startup when multiple TOPICs could be received for one room.

Aims to fix https://github.com/matrix-org/matrix-appservice-irc/issues/100


lukebarnard1:
Tests are failing because they expect to be able to bridge topics for mappings set in the config file, I shall update them to update current functionality.

lukebarnard1:
@Kegsay, FYI I think I might coroutinise that test (tomorrow)

lukebarnard1:
The test has been coroutinised, @Kegsay PTAL

lib/bridge/IrcHandler.js
```diff
@@ -16,6 +16,12 @@ function IrcHandler(ircBridge) {
     this._roomIdToPrivateMember = {
         // room_id: { user_id: $USER_ID, membership: "join|invite|leave|etc" }
     };
+
+    // Maintain a map of topics so as not to repeatidly send duplicate topics into
```
Kegsay: `repeatidly`

spec/integ/irc-to-matrix.spec.js
```diff
@@ -115,10 +115,23 @@ describe("IRC-to-Matrix message bridging", function() {
         });
     });
 
-    it("should bridge IRC topics as Matrix m.room.topic",
-    function(done) {
+    it("should bridge IRC topics as Matrix m.room.topic in aliased rooms",
+    test.coroutine(function*() {
```
Kegsay: This is now a coroutine, meaning the test ends when this function returns. This function returns too early, before `cli.sendStateEvent.andCallFake` fires (and it would be calling a non-existent function `done()`).

lib/bridge/IrcHandler.js
```diff
@@ -201,6 +207,55 @@ IrcHandler.prototype.onMessage = Promise.coroutine(function*(req, server, fromUs
         return;
     }
 
+    // For topics, set matrix topic as AS bot user
+    if (action.type === "topic") {
```
Kegsay: This PR makes it so that there is no "common code" shared between topic changes and incoming messages anymore. Please can you split out everything in this `if` block into its own function like `onTopic` and call it from the `IrcEventBroker` to make this clearer.

lib/bridge/IrcHandler.js
```diff
@@ -201,6 +207,55 @@ IrcHandler.prototype.onMessage = Promise.coroutine(function*(req, server, fromUs
...
+        // Only bridge topics for rooms created by the bridge, via !join or an alias
+        let origins = ["join", "alias"];
+        let mappings = yield this.ircBridge.getStore().getMappingsForChannelByOrigin(
+            server, channel, origins
+        );
+        let promises = mappings.map((entry) => {
+
```
Kegsay: Too many needless blank lines. Here, `:227`, `:233`, `:243`. I like blank lines when they are used to express "blocks" of code, much like how you write paragraphs. This isn't what is being done here though.

lib/bridge/IrcHandler.js
```diff
@@ -201,6 +207,55 @@ IrcHandler.prototype.onMessage = Promise.coroutine(function*(req, server, fromUs
...
+                return;
+            }
+
+            req.log.info(
+                "Message is a topic, using AS bot to set topic in room %s: %s",
+                entry.matrix.getId(),
+                action.text
+            );
+
+            return this.ircBridge.getAppServiceBridge().getIntent().setRoomTopic(
```
Kegsay: I tried to warn you about this yesterday, but maybe I wasn't clear enough. This code will race.

Even *with* the guard at `this._roomIdToTopic`, you can still set the **same** room topic multiple times. Do you see why? I've highlighted this line for a reason :)

lib/bridge/IrcHandler.js
```diff
@@ -201,6 +207,55 @@ IrcHandler.prototype.onMessage = Promise.coroutine(function*(req, server, fromUs
...
+                    entry.matrix.getId(),
+                    action.text
+                );
+
+                this._roomIdToTopic[entry.matrix.getId()] = action.text;
+            }, (err) => {
+                req.log.error(err.stack);
+            });
+        });
+        if (mappings.length === 0) {
```
Kegsay: Whilst logically this is completely valid, from a reader's perspective this feels out of order. You're in effect doing:
```
// pseudocode
mappings = getSomeMappings();
promises = mappings.map(() => { mapToPromise() })
if (mappings.length === 0) {
  log("no mappings!")
}
```

This feels out of order because typically you would **check** `getSomeMappings()` returned something sensible before operating on it, rather than operating on it **then** checking. Like I said, this is valid because it returns an empty array, but it "feels" out-of-order. This is clearer:

```
mappings = getSomeMappings();
if (mappings.length === 0) {
  log("no mappings!")
  return
}
promises = mappings.map(() => { mapToPromise() })

```

lib/bridge/IrcHandler.js
lukebarnard1: So the set should be moved up to be immediately after the if?

lib/bridge/IrcHandler.js
```diff
@@ -16,6 +16,12 @@ function IrcHandler(ircBridge) {
     this._roomIdToPrivateMember = {
         // room_id: { user_id: $USER_ID, membership: "join|invite|leave|etc" }
     };
+
+    // Maintain a map of topics so as not to repeatidly send duplicate topics into
+    // a room.
+    this._roomIdToTopic = {
```
lukebarnard1: Oh I see, I thought we were happy with passing topics through on startup, my mistake.

lib/bridge/IrcHandler.js
Kegsay: Yup. Of course, there is no guarantee that after we proclaim "the topic on room X is Y" that we will end up succeeding to set the topic... so a 3-state "it is X", "it is going to be Y", "it is Y" may be a better solution.

spec/integ/irc-to-matrix.spec.js
```diff
@@ -115,23 +115,38 @@ describe("IRC-to-Matrix message bridging", function() {
...
+                resolve();
+                return Promise.resolve();
+            });
         });
-    });
+
+        let client = yield env.ircMock._findClientAsync(roomMapping.server, roomMapping.botNick);
+        client.emit("topic", tChannel, testTopic, tFromNick);
+
+        return p;
```
Kegsay: Why aren't you `yield`ing here?

lib/bridge/IrcHandler.js
```diff
@@ -176,6 +176,66 @@ IrcHandler.prototype.onPrivateMessage = Promise.coroutine(function*(req, server,
...
+                action.text
+            );
+            return;
+        }
+        entry.matrix.topic = action.text;
+        this.ircBridge.getStore().upsertRoomStoreEntry(entry);
+        req.log.info(
+            "Message is a topic, using AS bot to set topic in room %s: %s",
+            entry.matrix.getId(),
+            action.text
```
Kegsay: The main use for these log lines is for debugging when things blow up. You want to minimise the amount of noise in the logs. Most of this line is redundant and not telling you anything: ""Message is a topic, using AS bot to set topic in room" has little information, and even the params given at the end are missing one of the most important pieces of information: which channel? I would say:
```js
`New topic in ${channel} - bot setting topic in ${entry.matrix.getId()}`
```

We don't care about the topic content.

lib/bridge/IrcHandler.js
```diff
@@ -176,6 +176,66 @@ IrcHandler.prototype.onPrivateMessage = Promise.coroutine(function*(req, server,
...
+        if (entry.matrix.topic === action.text) {
+            req.log.info(
+                "Topic of %s already set to '%s'",
+                entry.matrix.getId(),
+                action.text
+            );
+            return;
+        }
+        entry.matrix.topic = action.text;
+        this.ircBridge.getStore().upsertRoomStoreEntry(entry);
```
Kegsay: This isn't safe to be done here. There's a lot of race conditions, but unfortunately this one is a big one.

Assuming the IRC Event Broker is doing its job, we *shouldn't* get `onTopic` called multiple times for the *same* topic change (1 per connection), so `onTopic` will only be called once and only once when the topic changes. So far so good.

When it does get called, we'll try to the following *concurrently*:
 - Update the DB
 - Set the topic via HTTP

What if the HTTP request fails? The topic will be in the DB but will not be set on the room! When should we be updating the DB?

This gets exacerbated if a real IRC user changes the topic very quickly (e.g. `/topic 1`, `/topic 2`, `/topic 3`). Is there any guarantee that "3" will be the final topic set on Matrix?

Kegsay:
Aside from the race on quick topic changes and the more obvious race-if-fail-to-send-http, this is looking good.

lukebarnard1:
Merging in the fix for flaky tests seems to have fixed things :+1:

lib/bridge/IrcHandler.js
```diff
@@ -9,13 +9,22 @@ var MatrixRoom = require("matrix-appservice-bridge").MatrixRoom;
...
     // need to re-invite them if they bail.
     this._roomIdToPrivateMember = {
         // room_id: { user_id: $USER_ID, membership: "join|invite|leave|etc" }
     };
+
+    // Use a queue to keep the setting of topics in rooms atomic in order to
+    // prevent races involving several topics being received from IRC in quick
+    // succession. This queue will cause them to block unless the server, channel
+    // and topic are the same as an ongoing promise to set a topic, whihc will in
+    // that case be returned.
```
Kegsay: `whihc`

Also, last sentence is a bit hard to grok, you're describe a 3-tuple with words - I would just say:

> This queue will cause them to block unless (server, channel, topic) are the same [...]

lib/bridge/IrcHandler.js
```diff
@@ -176,6 +185,86 @@ IrcHandler.prototype.onPrivateMessage = Promise.coroutine(function*(req, server,
...
+            }
+        );
+    });
+    try {
+        yield Promise.all(promises);
+        item.req.log.info(
+            "Set topics: ",
+            item.entries.map(
+                (entry) => `id = ${entry.matrix.getId()}, topic = '${entry.matrix.topic}'`
+            ).join('\n')
```
Kegsay: The topic will always be the same, so no point logging it N times where N = the number of rooms. Also, the topic may be large, so I would cheat and substring it to death via `.substring(0, 20)` (first 20 chars) which will Do The Right Thing if the topic is < 20 chars too. So in total, something like this please:
```js
item.req.log.info(
    `Topic '${item.topic.substring(0, 20)}...' set in rooms: `, 
    item.entries.map((entry) => entry.matrix.getId()).join(",")
);
```

lib/bridge/IrcHandler.js
```diff
@@ -176,6 +185,86 @@ IrcHandler.prototype.onPrivateMessage = Promise.coroutine(function*(req, server,
...
+            );
+            return Promise.resolve();
+        }
+        console.log('setRoomTopic');
+        return this.ircBridge.getAppServiceBridge().getIntent().setRoomTopic(
+            entry.matrix.getId(),
+            item.topic
+        ).then(
+            () => {
+                console.log('upsertRoomStoreentry');
```
Kegsay: Remove please.

Kegsay:
Aside from the logging and the Queue ID, LGTM!

lib/bridge/IrcHandler.js
```diff
@@ -176,6 +185,86 @@ IrcHandler.prototype.onPrivateMessage = Promise.coroutine(function*(req, server,
...
+        return;
+    }
+
+    req.log.info(
+        "New topic in %s - bot queing to set topic in %s",
+        channel,
+        entries.map((e) => e.matrix.getId())
+    );
+    yield this.topicQueue.enqueue(
+        server.domain + " " + channel + " " + topic,
```
lukebarnard1: I'm not sure they would run concurrently, as the Queue only handles one item at a time...? Also, by removing the topic, if /topic 1, /topic 2 is received, then /topic 2 will result in the /topic 1 promise being returned if it is still executing, which is wrong.

Kegsay: Oh, you're right, I'm misreading this.

So we have a single global queue for every topic change in the entire bridge? Can we map this to be per-channel please to avoid head of line blocking?

lukebarnard1: Sure

lukebarnard1:
@Kegsay PTAL

Kegsay:
LGTM

PR #245, opened 2016-10-18
Enable simple presence syncing

### LINKED ISSUE #222: Presence-based membership list syncing
We really really really need to be syncing membership list from IRC to Matrix by default so that folks in Matrix don't end up speaking to long-since-gone users on the IRC side.  We've held off historically as the volume of join/part spam is too high and chews CPU & diskspace on synapse with all the state changes.

A relatively nice solution could be:
- When IRC users quit, the bridge changes their presence to `disconnected` (a new state)
- If they come back within N minutes (after bouncing their clients; a netsplit etc), their presence is restored to online.
- Otherwise they are parted from the room with an actual membership change.

Clients would treat 'disconnected' presence events very similarly to part events; they're just not persisted in the DAG (but are persisted otherwise, so actually perfectly useful).

Discussing with @erikjohnston there are concerns that this may be overkill given the biggest source of offensive join/part traffic are netsplits, and we can typically recognise and suppress those by the quit messages (although one might still want to bridge them into presence data).

Suggestion is to do some stats to find out what patterns there are in typical IRC join/part spam - specifically, how many parts are transient, and how many of those are netsplits, and so what the best mitigation is.

lukebarnard1:
When a user joins a channel, 'online' will be sent as their presence to the server. When a user parts and a 'quit' is indicated (i.e. the user quit), the bridge will set their presence to 'offline' and wait 15mins before causing the corresponding virtual matrix user to leave rooms. If the user joins a channel in that time, their presence will be set to 'online' and the corresponding virtual matrix user will not leave rooms.

@Kegsay I have tested this locally and have observed that if a user `/quit`s, and then reconnects and joins, the 'online' presence does get sent to the HS, but this is not always reflected in the observed member list state (say, in a Riot instance for example).

Intended to fix #222


lib/bridge/IrcHandler.js
```diff
@@ -9,13 +9,22 @@ var MatrixRoom = require("matrix-appservice-bridge").MatrixRoom;
...
     // need to re-invite them if they bail.
     this._roomIdToPrivateMember = {
         // room_id: { user_id: $USER_ID, membership: "join|invite|leave|etc" }
     };
+    this._rejoinPromises = {
+        // nick: Promise that resolves if the user joins a channel having quit
+        //       previously
+    };
+    this._rejoinResolvers = {
+        // nick: function to call when a user joins
```
Kegsay: Keying off the nick alone is not sufficient. It needs to be a `(server.domain, nick)` tuple.

lib/bridge/IrcHandler.js
```diff
@@ -367,6 +391,44 @@ IrcHandler.prototype.onPart = Promise.coroutine(function*(req, server, leavingUs
...
+        let presence = "offline";
+        yield Promise.all(matrixRooms.map((room) => {
+            req.log.info(
+                "Setting presence of %s in room %s to %s",
+                matrixUser.getId(), room.getId(), presence
+            );
+            return this.ircBridge.getAppServiceBridge().getIntent(
+                matrixUser.getId()
+            ).client.setPresence(presence);
+        }));
```
Kegsay: This doesn't make sense. Presence is user-scoped, not room-scoped, so why are we setting the presence N times, one for each mapped room?

lib/bridge/IrcHandler.js
```diff
@@ -367,6 +391,44 @@ IrcHandler.prototype.onPart = Promise.coroutine(function*(req, server, leavingUs
...
+
+        this._rejoinPromises[nick] = new Promise((resolve, reject) => {
+            this._rejoinResolvers[nick] = resolve;
+            setTimeout(() => {
+                reject(new Error("Rejoin timeout"));
+            }, DISCONNECT_PRESENCE_TIMEOUT);
```
Kegsay: This is a bit "messy" with resources - if we part/rejoin quickly, this timeout will still be lurking about for a full 15mins, holding a closure and preventing GC. Furthermore, if we `reject()` then subsequently get a join, we'll `resolve()` an already fulfilled Promise. Bluebird *will* handle that, but it's really messy. Can we not do:
```js
this._rejoinPromises[nick] = new Promise((resolve) => {
    this._rejoinResolvers[nick] = resolve;
}).timeout(DISCONNECT_PRESENCE_TIMEOUT);
```

Kegsay:
Broadly speaking, this looks like it could work. Presence won't always be represented on Riot AFAIK because there's weird interactions with `/sync` altering presence IIRC. We'll have to hunt that down.

This almost certainly wants to go behind a config option of sorts though, given the only reason we're doing this is because our HS is slow. Other people's won't be, and we shouldn't be forcing this 15min wait. Also needs netsplit handling.

lib/bridge/IrcHandler.js
lukebarnard1: Yes, and that seems so obvious now you've mentioned it! Bluebird is nice :+1:

lib/bridge/IrcHandler.js
```diff
@@ -330,15 +335,27 @@ IrcHandler.prototype.onJoin = Promise.coroutine(function*(req, server, joiningUs
...
     let matrixUser = yield this.ircBridge.getMatrixUser(joiningUser);
     req.log.info("Mapped nick %s to %s", nick, JSON.stringify(matrixUser));
     let matrixRooms = yield this.ircBridge.getStore().getMatrixRoomsForChannel(server, chan);
     let promises = matrixRooms.map((room) => {
-        req.log.info("Joining room %s", room.getId());
-        return this.ircBridge.getAppServiceBridge().getIntent(
-            matrixUser.getId()
-        ).join(room.getId());
+        req.log.info("Joining room %s and setting presence to online", room.getId());
+        let intents = [];
```
Kegsay: These are technically promises, not Intent objects.

lib/irc/IrcServer.js
```diff
@@ -31,6 +31,27 @@ IrcServer.prototype.getReadableName = function() {
     return '';
 }
 
+/**
+ * Returns whether the server is configured to wait some time
+ * (see this.config.getDisconnectPresenceTimeoutMs()) before parting
```
Kegsay: `getDisconnectPresenceTimeoutMs` doesn't exist?

lib/irc/IrcServer.js
```diff
@@ -31,6 +31,27 @@ IrcServer.prototype.getReadableName = function() {
...
+ */
+IrcServer.prototype.shouldDebouceParts = function() {
+    return this.config.debounceParts;
+}
+
+/**
+ * Get the number of ms to debounce before bridging a `/quit` to a leave
+ * during a detected net-split. If the user rejoins a channel before bridging
+ * the quit to a leave, the leave will not be sent.
+ * @return {integer}
```
Kegsay: `Number` in JS, not `integer`.

lib/irc/IrcServer.js
```diff
@@ -31,6 +31,27 @@ IrcServer.prototype.getReadableName = function() {
     return '';
 }
 
+/**
+ * Returns whether the server is configured to wait some time
+ * (see this.config.getDisconnectPresenceTimeoutMs()) before parting
+ * a user that has `/quit` and also wait getDeboucePartsMs before parting
+ * a user that has disconnected due to a net-split.
+ * @return {bool} this.config.debounceParts.
```
Kegsay: `Boolean` in JS, not `bool`.

lib/irc/IrcServer.js
```diff
@@ -31,6 +31,27 @@ IrcServer.prototype.getReadableName = function() {
...
+}
+
+/**
+ * Get the number of ms to debounce before bridging a `/quit` to a leave
+ * during a detected net-split. If the user rejoins a channel before bridging
+ * the quit to a leave, the leave will not be sent.
+ * @return {integer}
+ */
+IrcServer.prototype.getDeboucePartsMs = function() {
+    return this.config.debouncePartsMs;
```
Kegsay: Missing examples in `config.sample.yaml`. Missing defaults set in the default config. Should group together related variables `debouncePartsMs`, `debounceParts`, `getDisconnectPresenceTimeoutMs` in its own section like we do for other things.

lib/bridge/QuitDebouncer.js
```diff
@@ -0,0 +1,106 @@
...
+        Object.keys(this._recentQuits).forEach(
+            (server) => {
+                this._quitsPerSecond[server] = this._recentQuits[server];
+                this._recentQuits[server] = 0;
+
+                let threshold = 5;
+
+                // Possible net-split
+                if (this._quitsPerSecond[server] > threshold) {
+                    this._is_split_occuring = true;
```
Kegsay: This fails in multi-server scenarios (where an IRC bridge is bridging to Freenode and Moznet for example). If Freenode netsplits, we will incorrectly execute netsplitting logic for **all servers**, in this case including Moznet.

lib/bridge/QuitDebouncer.js
```diff
@@ -0,0 +1,106 @@
...
+            }
+        );
+    }, 1000);
+}
+
+
+/**
+ * Called when the IrcHandler receives a JOIN. This resolves any promises to join that were made
+ * when a quit was debounced during a split.
+ * @param {string} nick The nick of the sending IRC user.
```
Kegsay: > The nick of the sending IRC user.

"the sending IRC user"? This isn't clear. This function is called when the given nick *joins* any Matrix-bridged IRC channel, you should mention that somewhere.

> This resolves any promises to join

"to join"? Doesn't make sense?

lib/bridge/IrcHandler.js
```diff
@@ -456,6 +473,20 @@ IrcHandler.prototype.onPart = Promise.coroutine(function*(req, server, leavingUs
...
+    //  When an IRC user quits, debounce before leaving them from matrix rooms. In the meantime,
+    //  update presence to "offline". If the user rejoins a channel before timeout, do not part
+    //  user from the room. Otherwise timeout and leave rooms.
+    let shouldBridgePart = true;
+    if (kind === "quit" && server.shouldDebouceParts()) {
+        shouldBridgePart = yield this.quitDebouncer.debounceQuit(req, server, matrixUser);
+    }
+
+    if (!shouldBridgePart) {
+        return;
```
Kegsay: Leaky control flow. Why not just:
```js
if (kind === "quit" && server.shouldDebouceParts()) {
    let shouldBridgePart = yield this.quitDebouncer.debounceQuit(req, server, matrixUser);
    if (!shouldBridgePart) {
        return;
    }
}

File: lib/bridge/QuitDebouncer.js [hunk 12]
```diff
@@ -0,0 +1,106 @@
...
+QuitDebouncer.prototype.debounceQuit = Promise.coroutine(function*(req, server, matrixUser) {
+    // Initialise counters for quit statistics
+    if (this._quitsPerSecond[server.domain] === undefined) {
+        this._recentQuits[server.domain] = 0;
+        this._quitsPerSecond[server.domain] = 0;
+    }
+    this._recentQuits[server.domain] += 1;
+
+    // TODO: This should be replaced with "disconnected" as per matrix-appservice-irc#222
+    let presence = "offline";
```
Kegsay: Any reason why this isn't simply a string change? Can we do it now?

lib/bridge/QuitDebouncer.js
```diff
@@ -0,0 +1,106 @@
...
+    yield this.ircBridge.getAppServiceBridge().getIntent(
+        matrixUser.getId()
+    ).client.setPresence(presence);
```
Kegsay: So if this request fails, we not only fail to debounce, but we fail the entire `onLeave` request (no catches here or in `IrcHandler`!).

lib/bridge/QuitDebouncer.js
```diff
@@ -0,0 +1,106 @@
...
+
+    // Bridge parts if a split is not occuring. This is in the case where a QUIT is
```
Kegsay: `occuring`

lib/bridge/QuitDebouncer.js
```diff
@@ -0,0 +1,106 @@
...
+    // received for reasons such as ping timeout or IRC client (G)UI being killed.
+    if (!this._is_split_occuring) {
+        return true;
```
Kegsay: > Bridge parts

It's not the bridge parting though, is it? It's the user leaving? Also, the comment isn't really clear **why** we should part if a split is not occurring. Please clarify.

lib/bridge/QuitDebouncer.js
```diff
@@ -0,0 +1,106 @@
...
+    // This is to smooth incoming PART spam from IRC clients that suffer from a
+    // net-split (or other issues that lead to mass PART-ings)
+    this._recentQuits = {
+        // $server => {integer} : the number of recent parts with kind = quit
+    };
+    this._quitsPerSecond = {
+        // $server => {integer} : the number of quits in the last interval 1s
+    };
+
+    this._is_split_occuring = {
```
lukebarnard1: Fair point..

lib/bridge/QuitDebouncer.js
lukebarnard1: It's a promise to join a channel; a promise in the literal sense.

lib/bridge/IrcHandler.js
lukebarnard1: wfm

lib/bridge/QuitDebouncer.js
```diff
@@ -0,0 +1,106 @@
...
+var Promise = require("bluebird");
+
+function QuitDebouncer(ircBridge) {
+    this.ircBridge = ircBridge;
+
+    this._rejoinPromises = {
+        // '$nick $server.domain': Promise that resolves if the user joins a channel having quit
+        //       previously
+    };
+    this._rejoinResolvers = {
```
lukebarnard1: wfm

lib/bridge/QuitDebouncer.js
Kegsay: Then please add "to join a channel" else it's confusing.

lib/bridge/IrcHandler.js
Kegsay: Given you're defining, pushing then `Promise.all()`ing - I would just inline it:
```js
return Promise.all([
  this.ircBridge...,
  this.ircBridge...,
]);
```

lib/bridge/QuitDebouncer.js
Kegsay: Ahhh I see. Fair enough.

lib/bridge/IrcHandler.js
lukebarnard1: wfm

config.sample.yaml
```diff
@@ -50,6 +50,15 @@ ircService:
       # Whether or not to send connection/error notices to real Matrix users. Default: true.
       sendConnectionMessages: true
 
+      # Whether parts due to net-splits are debounced for debouncePartsMs, to allow
+      # time for the net-split to resolve itself. A net-split is detected as being
+      # a QUIT frequency higher than 5Hz. Default: false.
```
Kegsay: This should be customisable rather than hard-coded to 5. This is why I mentioned about breaking this out into a separate config section like:
```yaml
debounce:
  enabled: true
  quitsPerSecond: 5    # If this is met or exceeded...
  waitTimeMs: 3600000  # Wait this amount of time for them to come back before parting them from Matrix
```

lib/irc/IrcServer.js
```diff
@@ -31,6 +31,25 @@ IrcServer.prototype.getReadableName = function() {
...
+ * Returns whether the server is configured to wait getDeboucePartsMs before
+ * parting a user that has disconnected due to a net-split.
+ * @return {Boolean} this.config.debounceParts.
+ */
+IrcServer.prototype.shouldDebounceParts = function() {
+    return this.config.debounceParts;
+}
+
+/**
+ * Get the number of ms to debounce before bridging a `/quit` to a leave
```
Kegsay: > before bridging a `QUIT` to Matrix

would read better.

lib/bridge/QuitDebouncer.js
```diff
@@ -0,0 +1,125 @@
...
+      // }
+    };
+
+    // Keep a track of the times at which debounceQuit was called, and use this to
+    // determine the rate at which quits are being received. This can then be used
+    // to detect net splits.
+    /*$server : [1477386173850, 1477386173825, ...]*/
+    this._quitTimesMs = {};
+    Object.keys(this.ircBridge.config.ircService.servers).forEach((domain) => {
+        this._quitTimesMs[domain] = [];
```
Kegsay: Any reason why this isn't bundled with `_debouncerForServer`? They use the same keys after all, then you wouldn't need the null guard at `:37`.

lib/bridge/QuitDebouncer.js
```diff
@@ -0,0 +1,125 @@
...
+            //   resolve: Function //function to call when a user joins
+            // }
+      // }
+    };
+
+    // Keep a track of the times at which debounceQuit was called, and use this to
+    // determine the rate at which quits are being received. This can then be used
+    // to detect net splits.
+    /*$server : [1477386173850, 1477386173825, ...]*/
+    this._quitTimesMs = {};
```
Kegsay: The resolution is ms but this name is misleading, I think it's durations but it isn't. You should mention the word "timestamp" somewhere in this variable name.

lib/bridge/QuitDebouncer.js
```diff
@@ -0,0 +1,125 @@
...
+ * @param {string} nick The nick of the IRC user joining.
+ * @param {IrcServer} server The sending IRC server.
+ */
+QuitDebouncer.prototype.onJoin = function (nick, server) {
+    if (!this._debouncerForServer[server.domain]) {
+        return;
+    }
+    let rejoin = this._debouncerForServer[server.domain][nick];
+    if (rejoin) {
+        rejoin.resolve()
```
Kegsay: `;`

lib/bridge/QuitDebouncer.js
```diff
@@ -0,0 +1,125 @@
...
+    }
+}
+
+/**
+ * Debounce a QUIT received by the IrcHandler to prevent net-splits from spamming leave events
+ * into a room when incremental membership syncing is enabled.
+ * @param {Request} req The metadata request.
+ * @param {IrcServer} server The sending IRC server.
+ * @param {string} matrixUser The virtual user of the user that sent QUIT.
```
Kegsay: Missing a `@param {string} nick`.

config.sample.yaml
```diff
@@ -50,6 +50,15 @@ ircService:
...
+      # debounceParts: true
+
+      # The time to wait before bridging a QUIT to a leave that occured during a net-split.
+      # Default: 3600000, = 1h
+      # debouncePartsMs: 3600000 # 1h
```
Kegsay: These need to be in the config schema file too!

lib/bridge/QuitDebouncer.js
```diff
@@ -0,0 +1,125 @@
...
+QuitDebouncer.prototype.debounceQuit = Promise.coroutine(function*(req, server, matrixUser, nick) {
+    // Maintain the last windowMs worth of timestamps corresponding with calls to this function.
+    let now = Date.now();
+    let windowMs = 1000;// Window starts 1s ago
+    let threshold = 5;// Need 5Hz of quits to call net-split
+    this._quitTimesMs[server.domain].push(now);
+
+    // Filter out timestamps from more than 1s ago
+    this._quitTimesMs[server.domain] = this._quitTimesMs[server.domain].filter(
+        (t) => t > now - windowMs
```
Kegsay: Brackets for clarity please:
```js
(t) => t > (now - windowMs)
```

lib/bridge/QuitDebouncer.js
```diff
@@ -0,0 +1,125 @@
...
+    );
+
+    // Wait for a short time to allow other potential splitters to send QUITs
+    yield Promise.delay(100);
```
Kegsay: This isn't necessary. Step through your code if you get 1 QUIT every 100ms. At best, you'll save a few immediate parts being sent to Matrix. At worst, you're delaying every single part by 100ms.

lib/bridge/QuitDebouncer.js
```diff
@@ -0,0 +1,125 @@
...
+    let presence = "offline";
+    try {
+        yield this.ircBridge.getAppServiceBridge().getIntent(
+            matrixUser.getId()
+        ).client.setPresence(presence);
+    }
+    catch (err) {
+        req.log.error(
+            'QuitDebouncer Failed to set presence to offline for user %s',
+            matrixUser.getId()
```
Kegsay: Shame we'll never know why we failed to set the presence to offline... 0:)

lib/bridge/QuitDebouncer.js
```diff
@@ -0,0 +1,125 @@
...
+    if (this._debouncerForServer[server.domain] === undefined) {
+        this._debouncerForServer[server.domain] = {};
+    }
+
+    // TODO: This should be replaced with "disconnected" as per matrix-appservice-irc#222
+    let presence = "offline";
+    try {
+        yield this.ircBridge.getAppServiceBridge().getIntent(
+            matrixUser.getId()
+        ).client.setPresence(presence);
```
Kegsay: Feels too early. If we aren't in a netsplit, we're just going to leave them anyway, so why bother setting their presence? The point of `disconnected` is to say "We would've left this room, but we're hanging around for a bit to wait an see if we come back".

lib/bridge/QuitDebouncer.js
lukebarnard1: OK, so should I get rid of this yield and add the caveat in the config that the first getDebounceQuitsPerSecond users will immediately bridge parts regardless of whether they are part of the net split?

lib/bridge/QuitDebouncer.js
lukebarnard1: Presence is per-user, so if they quit, don't we want to set presence anyway? Or should it be 'discconected' for netsplits and 'offline' for a legitimate quit?

lib/bridge/QuitDebouncer.js
Kegsay: There is no caveat to point out here. The config looks like:
```yaml
debounce:
  enabled: true
  quitsPerSecond: 5    # If this is met or exceeded...
  waitTimeMs: 3600000  # Wait this amount of time for them to come back before parting them from Matrix
```
In the first 400ms, there are 4 quits. This does not meet or exceed 5 quits per second, so there should be no wait time. The fact that they may be part of a netsplit or not is neither here nor there.

lib/bridge/QuitDebouncer.js
Kegsay: Hmm, let's work this out.

Ignore presence for a moment. What happens when an IRC user quits the network? Each channel gets a QUIT, which the bridge translates into leave requests. That is the best behaviour because it most accurately describes what is going on in the membership list.

Now, how can presence changes improve this? Well for the channel case, it can't. For the PM case, it would be great to tell the user that the person they are talking to has disconnected, so "disconnected" would be great, and "offline" would be bad. They can't really leave the room because there is no one to invite them back into the room when they reconnect (the AS bot isn't a part of PM rooms). That's using "disconnected" outside the context of netsplits, and isn't one I previously considered.

Separately, we don't want to flood Synapse with leave requests. In netsplits, this `QuitDebouncer` will pick up on high rates of leaves and change the behaviour of the bridge: it will set the presence to "disconnected" rather than leave them from the Matrix room. Now my comment may make more sense: if we are not in a netsplit, we will leave the room. The Matrix user will never see the presence being set to disconnected because they won't be on the membership list anymore, so my question was "why should we bother then"?

The PM use case is a good counter-example of why we *should* bother.

lukebarnard1: Oh ok, that makes sense. So should I leave it then?

Kegsay: Probably, yeah.

lib/bridge/QuitDebouncer.js
lukebarnard1: This is done (also it would be helpful if you could click the [+] on a line that will definitely change if the comment requires it (also, I prefer the new style of review - it supports 'outdated' now))

config.sample.yaml
```diff
@@ -50,6 +50,20 @@ ircService:
       # Whether or not to send connection/error notices to real Matrix users. Default: true.
       sendConnectionMessages: true
 
+      quitDebounce:
+        # Whether parts due to net-splits are debounced for debouncePartsMs, to allow
```
Kegsay: `s/debouncePartsMs/delayMs/`

lib/bridge/QuitDebouncer.js
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    let debounceMs = server.getQuitDebounceDelayMs();
+
+    // We do want to immediately bridge a leave if undefined or set to 0
+    if (!debounceMs) {
+        return true;
+    }
+
+    req.log.info('Debouncing for ' + debounceMs + 'ms\n');
```
Kegsay: Please don't insert `\n`s into logs.

Kegsay:
Otherwise LGTM! :+1:

PR #246, opened 2016-10-18
Set `rate_limited = false` reg flag on generation

lukebarnard1:
This is to prevent rate limiting from the HS, for example when starting up with lots of users to connect and lots of messages being sent I->M by NickServ/ChanServ.

This depends on https://github.com/matrix-org/matrix-appservice-node/pull/6


Kegsay:
LGTM

PR #249, opened 2016-10-24
Ignore IRC onPM if bot and not message

lukebarnard1:
The IRC handler will ignore PMs directed to a bridgedClient that is as bot unless the IRC action type is "message". This includes notices from NickServ, for example.


Kegsay:
LGTM

PR #251, opened 2016-10-25
Store passwords in user.db configs

lukebarnard1:
Send `!storepass server.name password` to have your virtual IRC client's password stored encrypted in the db.

The passwords are encrypted and decrypted using a PEM key stored in a file specified in the config file (See passwordEncryptionKeyPath).

The passwords are then used in place of the server password set in the config file, whether it is set or not.

This has been tested to work whilst bridging with freenode, and as expected, when !nick and !storepass are issued prior to the bridge restarting (or the IRC client reconnecting), NickServ asks for identification (if the nick is registered) but then immediately resolves (having received PASS with the recently decrypted password).


config.sample.yaml
```diff
@@ -328,3 +328,14 @@ ircService:
...
+  # WARNING - STORING PASSWORDS ENCRYPTED IS PRECAUTIONARY TO MAKE COMPROMISE LESS
+  #           LIKELY. THIS MEASURE IS ONLY AS SECURE AS THE KEY FILE, BELOW.
+  #
+  # The path to the RSA PEM-formatted key to use when encrypting and decrypting
+  # IRC passwords that have been stored in the database. Passwords are stored by
+  # using the admin room `!storepass server.name passw0rd. When a connection is
+  # made for the corresponding virtual IRC user of the matrix user that has issued
+  # !storepass, they will connect with the password that has been persisted, if
+  # there is one.
+  passwordEncryptionKeyPath: "passkey.pem"
```
Kegsay: NL

lib/bridge/IrcBridge.js
```diff
@@ -137,6 +137,17 @@ IrcBridge.prototype.createBridgedClient = function(ircClientConfig, matrixUser,
...
     }
+
+    if (ircClientConfig._config.isPasswordEncrypted) {
+        if (!this.getStore()._privateKey) {
+            throw new Error(
+                `Cannot create BridgeClient with encrypted password unless a private key ` +
+                `is provided to the bridge. See config 'passwordEncryptionKeyPath'.`
+            );
+        }
+        ircClientConfig.setPrivateKey(this.getStore()._privateKey);
```
Kegsay: Please don't gut-wrench private member variables.

lib/bridge/IrcBridge.js
```diff
@@ -158,7 +169,13 @@ IrcBridge.prototype.run = Promise.coroutine(function*(port) {
         this._debugApi.run();
     }
 
-    this._dataStore = new DataStore(this._bridge.getUserStore(), this._bridge.getRoomStore());
+    let pkeyPath = this.config.ircService.passwordEncryptionKeyPath;
+
+    this._dataStore = new DataStore(
+        this._bridge.getUserStore(),
+        this._bridge.getRoomStore(),
+        pkeyPath
```
Kegsay: In-line please.

lib/bridge/MatrixHandler.js
```diff
@@ -451,6 +451,27 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
...
         return;
     }
+    else if (cmd === "!storepass") {
+        let domain = ircServer.domain;
+        let userId = event.user_id;
+        let notice;
+
+        try {
+            let force = args[1] === '-f';
+            yield this.ircBridge.getStore().storePass(userId, domain, args[0], force);
```
Kegsay: Right now I see why you mentioned `-f` before. Why is `-f` A Thing? This should not be an option for random Matrix users.

lib/models/IrcClientConfig.js
```diff
@@ -37,12 +44,46 @@ class IrcClientConfig {
         return this._config.username;
     }
 
+    setPrivateKey(privateKey) {
+        if (this._privateKey) {
+            throw new Error("Cannot override previously set privateKey!");
+        }
+        if (typeof privateKey !== 'string') {
+            let actual = typeof privateKey;
+            throw new Error("Private key must be an RSA PEM-formatted string, not " + actual);
```
Kegsay: This all feels like needless boilerplate. We load the PEM key from file in `DataStore`, why aren't we just checking the types there? Why do we pass the private key through as a member variable to `IrcClientConfig`? Why is it even an option to store passwords unencrypted? It means you get these scenarios where you have no idea what `_config.password` is (unencrypted or encrypted) and you just make more work for yourself.

Kegsay:
Broadly speaking this is looking good, but the form of the API is questionable.
 - I don't like how we pass around the private key between `DataStore` and `IrcClientConfigs`, and do some checks in each: just choose one and stick with it (probably `DataStore`).
 - I don't like how users can choose to not encrypt their passwords: it's unthinkable that we should persist any password in the clear on the database.
 - I don't like how the config has to "redact" the password field - it shouldn't be stored in `_config` in the clear. `getPassword` should decrypt the encrypted form and return that and that's it.

lukebarnard1:
On your three points:

- If DataStore is the only thing to ever store the private key, we would need some way to get the privateKey from there into calls to IrcClientConfig.getPassword(privateKey) because that is how the IrcBridge class gets the passwords. To be honest, I just couldn't see a clearer way of doing it architecturally, any input would be appreciated. To clarify, the problem with storing in DataStore is that the IrcBridge would have to inspect the DataStore every time it calls `getPassword`.
- Fair enough, I'll remove the whole '-f' thing and replace 'use -f' notice with 'no private key loaded'.
- I think it's perfectly fine to store the encrypted password in `_config` and I don't understand why we wouldn't redact the encrypted form. (Also, I'm assuming 'in the clear' means plaintext as opposed to encryptedt?)

Kegsay:
Well it doesn't help you appear to be trying to reuse `getPassword` when it is currently being used for the bot password: https://github.com/matrix-org/matrix-appservice-irc/blob/master/lib/irc/IrcServer.js#L94 - but of course this will break when you add encryption and expect that field to be encrypted.

The private key is loaded from file when the `DataStore` starts up. When you hit `storePass` you give the plaintext and behind the scenes it does the necessary magic to plop it into the DB encrypted. This feels right: nothing else should really give a damn about this.

Can we do the same for `getPassword`? That would imply we have the client config store the *unencrypted* form inside itself, since it doesn't know about the encryption at all. When the bridge tries to load the `IrcClientConfig` for a user, the `DataStore` magically decrypts entries. This would also mean my first point about it breaking for the bot user is invalid: since it'll still be the unencrypted form. WDYT?

lukebarnard1:
That sounds reasonable, WFM.

config.sample.yaml
```diff
@@ -328,3 +328,13 @@ ircService:
...
+
+  # WARNING - STORED PASSWORDS ENCRYPTED ARE COMPLETELY RETRIEVABLE WITH THE KEY.
+  #           THE BRIDGE NEEDS TO SEND PASSWORDS TO THE IRC SERVERS AS PLAINTEXTS.
+  #
+  # The path to the RSA PEM-formatted private key to use when encrypting IRC passwords
+  # for storage in the database. Passwords are stored by using the admin room command
+  # `!storepass server.name passw0rd. When a connection is made for the corresponding
+  # virtual IRC user of the matrix user that has issued !storepass, they will connect
+  # with the (decrypted) password that has been persisted, if there is one.
+  passwordEncryptionKeyPath: "passkey.pem"
```
Kegsay: NL

lib/DataStore.js
```diff
@@ -51,6 +53,18 @@ function DataStore(userStore, roomStore) {
...
+
+    this._privateKey = null;
+
+    if (pkeyPath) {
+        try {
+            this._privateKey = fs.readFileSync(pkeyPath, "utf8").toString();
+            log.info(`Private key loaded from ${pkeyPath} - IRC password encryption enabled.`);
+        }
+        catch (err) {
+            log.error('Could not load private key ${err.message}.');
```
Kegsay: Then you can remove `:473` guard.

lib/bridge/MatrixHandler.js
```diff
@@ -451,6 +451,26 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
         yield this.ircBridge.sendMatrixAction(adminRoom, botUser, notice, req);
         return;
     }
+    else if (cmd === "!storepass") {
+        let domain = ircServer.domain;
+        let userId = event.user_id;
+        let notice;
+
+        try {
+            yield this.ircBridge.getStore().storePass(userId, domain, args[0]);
```
Kegsay: Are spaces allowed in NickServ passwords? If so, then `args[0]` ain't gonna cut it.

Kegsay: Or rather, it will cut it 😆

lib/models/IrcClientConfig.js
```diff
@@ -16,6 +16,11 @@ class IrcClientConfig {
         this.userId = userId;
         this.domain = domain;
         this._config = configObj || {};
+
+        // The PEM string used to encrypt passwords when setPassword is
+        // called and decrypt when getPassword is called.
+        this._privateKey = null;
```
Kegsay: Never used.

lib/models/IrcClientConfig.js
```diff
@@ -42,7 +47,7 @@ class IrcClientConfig {
     }
 
     getPassword() {
-        return this._config.password;
+       return this._config.password;
```
Kegsay: 3 space indent?

lib/DataStore.js
```diff
@@ -51,6 +53,18 @@ function DataStore(userStore, roomStore) {
         unique: true,
         sparse: false
     }, errLog("user id"));
+
+    this._privateKey = null;
+
+    if (pkeyPath) {
+        try {
+            this._privateKey = fs.readFileSync(pkeyPath, "utf8").toString();
+            log.info(`Private key loaded from ${pkeyPath} - IRC password encryption enabled.`);
```
Kegsay: We should probably assert that the key is usable first so we can fail fast if `_privateKey` isn't a PEM file. Try encrypting/decrypting a test string in this constructor.

lib/DataStore.js
lukebarnard1: But what if someone wants to use the bridge without bothering with generating a PEM key?

Kegsay: I don't follow. This block only executes `if (pkeyPath)`. So if you want to use the bridge without bothering to generate a PEM key, then don't fill in that config field?

lukebarnard1:
IRL: we discussed making the following changes:

- [ ] adding a random prefix salt to the passwords and storing them along side them

config.sample.yaml
lukebarnard1: (Done)

config.sample.yaml
```diff
@@ -328,3 +328,13 @@ ircService:
     # allotted time period, the provisioning request will fail.
     # Default: 300 seconds (5 mins)
     requestTimeoutSeconds: 300
+
+  # WARNING - STORED PASSWORDS ENCRYPTED ARE COMPLETELY RETRIEVABLE WITH THE KEY.
+  #           THE BRIDGE NEEDS TO SEND PASSWORDS TO THE IRC SERVERS AS PLAINTEXTS.
```
Kegsay: Maybe rephrase to:

> WARNING: The bridge needs to send plaintext passwords to the IRC server, it cannot send a password hash. As a result, passwords (NOT hashes) are stored encrypted in the database.

lib/DataStore.js
```diff
@@ -433,7 +461,19 @@ DataStore.prototype.getIrcClientConfig = function(userId, domain) {
...
+        let clientConfig = new IrcClientConfig(userId, domain, configData);
+        if (clientConfig.getPassword()) {
+            if (!this._privateKey) {
+                throw new Error(`Cannot decrypt password of ${userId} - no private key`);
+            }
+            let decryptedPass = crypto.privateDecrypt(
+                this._privateKey,
+                new Buffer(clientConfig.getPassword(), 'base64')
+            ).toString();
+            decryptedPass = decryptedPass.split(' ')[1];
```
Kegsay: Uhh what? Explanation please.

lib/bridge/MatrixHandler.js
```diff
@@ -451,6 +451,28 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
...
+        let domain = ircServer.domain;
+        let userId = event.user_id;
+        let notice;
+
+        try {
+            // Allow passwords with spaces
+            let pass = args.join('');
+            yield this.ircBridge.getStore().storePass(userId, domain, pass);
+            notice = new MatrixAction(
+                "notice", `Successfully stored password`
```
Kegsay: We should mention that this only applies on *reconnects* and that they still need to poke NickServ initially. Can we name the domain as well please (since the server is often implicit, they may not have typed it in).

> Successfully stored password on ${domain}. When you next reconnect to ${domain}, this password will be automatically sent in a PASS command which most IRC networks will use as your NickServ password. This means you will not need to talk to NickServ. This does NOT apply to your currently active connection: you still need to talk to NickServ one last time to authenticate your current connection if you haven't already.

Kegsay: We should probably also allow them to remove their stored password.

Kegsay:
Other than wordings on things, this is looking good.

I did have a thought though: we should probably allow people to remove their stored password somehow.

lib/DataStore.js
Kegsay: Can you put that as a comment then please :)

lukebarnard1:
- [ ] Add `!removepass` command to unset the password
- [ ] Handle empty password as an error case

Kegsay:
Should an empty pass be treated as an error or should we display the usage (similar to `!nick`)?

lib/DataStore.js
lukebarnard1: Sure!

lukebarnard1:
I went for the latter as it does indicate that the user doesn't really understand the command.

lib/bridge/MatrixHandler.js
```diff
@@ -451,6 +451,66 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
...
+
+        try {
+            // Allow passwords with spaces
+            let pass = args.join('');
+
+            if (pass.length === 0) {
+                notice = new MatrixAction(
+                    "notice",
+                    "Format: '!storepass password' " +
+                    "or '!storepass irc.server.name password'\n"
```
Kegsay: Explanation of what this does would be good. I would move the blurb you have on success ("When you next reconnect...") to a `var` and then use that here in addition to on success.

lib/bridge/MatrixHandler.js
```diff
@@ -451,6 +451,66 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
...
+            notice = new MatrixAction(
+                "notice", `Failed to store password: ${err.message}`
+            );
+            req.log.error(err.stack);
+        }
+
+        yield this.ircBridge.sendMatrixAction(adminRoom, botUser, notice, req);
+        return;
+    }
+    else if (cmd === "!removepass") {
```
Kegsay: Both this and `!storepass` need to be added to `!help`.

Kegsay:
After you add more words to the commands, LGTM!

lib/bridge/MatrixHandler.js
lukebarnard1: WFM

lib/bridge/MatrixHandler.js
lukebarnard1: Done.

3v1n0:
I'm wondering... Couldn't be the matrix protocol to be updated so that the key is encrypted using some runtime data that is valid for this user only?

I'm not into the deep of matrix, but having the service to require a matrix user pass whenever the nick pass should be used could be a way. Indeed this wouldn't make auto-connection work in case until the user has approved this, but it would make possible to trust the system whoever is hosting it.

Kegsay:
Sadly I don't think what you want is technically feasible. It always bugged me that we were/are persisting passwords on the bridge since it slaps a big target on our back, but you must understand the requirements here.

The bridge has to send the user/password to the target IRC server. Even if you only ever gave the bridge an encrypted password which you had the key for, then did some dance to decrypt it on one of your clients, it still wouldn't improve trust. There would be nothing to stop the bridge from retrieving the decrypted password from your client then doing bad things with it, in addition to logging you in to IRC. Simply put, the bridge *has* to have the password in the plain somewhere so it can do the necessary operation.

PR #253, opened 2016-10-26
Add !quit command to admin rooms

lukebarnard1:
Send `!quit server.name` to kill all bridgedClients that the sending use has associated with the specified server. This required some updates to the connection notices; the one that indicates a disconnect now doesn't read '...Reconnecting' and this has been replaced by a separate notice that says 'Connecting...' when a client is doing so. This does mean that if the server isn't starting up and the clients reconnect for any reason, 3 notices are now seen instead of 2, but I feel like that's a fair trade-off for being better informed.

### DIFF
--- a/lib/bridge/MatrixHandler.js
+++ b/lib/bridge/MatrixHandler.js
@@ -452,7 +452,9 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
         return;
     }
     else if (cmd === "!quit") {
-        clientList.forEach((bridgedClient) => {
+        clientList.filter(
+            (bridgedClient) => bridgedClient.server.domain === ircServer.domain
+        ).forEach((bridgedClient) => {
             bridgedClient.kill();
         });
     }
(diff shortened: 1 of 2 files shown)


lib/bridge/MatrixHandler.js
```diff
@@ -451,6 +451,13 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
         yield this.ircBridge.sendMatrixAction(adminRoom, botUser, notice, req);
         return;
     }
+    else if (cmd === "!quit") {
+        clientList.filter(
+            (bridgedClient) => bridgedClient.server.domain === ircServer.domain
+        ).forEach((bridgedClient) => {
+            bridgedClient.kill();
+        });
```
Kegsay: No success/failure message sent back?

lukebarnard1: They will see their connection drop and come back again. I could add even more feedback but it would feel redundant.

Kegsay:
LGTM

PR #259, opened 2016-10-27
Fix provisioning logging to give more context

lukebarnard1:


lib/provisioning/Provisioner.js
```diff
@@ -128,60 +129,48 @@ function Provisioner(ircBridge, enabled, requestTimeoutSeconds) {
...
-    }));
+    let createProvisionEndpoint = (fn, fnName) => {
+        return Promise.coroutine(function*(req, res) {
+            req = new ProvisionRequest(req, fnName);
+            req.log.info(
+                'New provisioning request: ' + JSON.stringify(req.body) +
+                ' params: ' + JSON.stringify(req.params)
+            );
+            try {
+                let result = yield fn.apply(self, [req]);
```
Kegsay: Also I would be tempted to make `fn` be a bound function so you don't need to do this (which means you can yank `createProvisionEndpoint` out to be a stand-alone function without referencing `self`), but I don't really mind either way.

lib/provisioning/Provisioner.js
```diff
@@ -805,7 +802,7 @@ Provisioner.prototype.unlink = Promise.coroutine(function*(options) {
             throw new Error(s);
         }
         else {
-            log.error(err);
+            // err is useless, change the message and throw
```
Kegsay: :cry: Poor error, crying out in the hopes that someone is listening, and yet... nothing.

What happens if `this._unlinkValidator.validate(options);` throws a `TypeError`? Will we ever know? 0:)

lib/provisioning/Provisioner.js
lukebarnard1: Binding on a `Promise.coroutine` function is apparently not straight forward... Any ideas?

lib/provisioning/Provisioner.js
```diff
@@ -128,60 +129,48 @@ function Provisioner(ircBridge, enabled, requestTimeoutSeconds) {
...
+                'New provisioning request: ' + JSON.stringify(req.body) +
+                ' params: ' + JSON.stringify(req.params)
+            );
+            try {
+                let result = yield fn.apply(self, [req]);
+                if (!result) {
+                    result = {};
+                }
+                req.log.info(`Sending result: ${JSON.stringify(result)}`);
+                res.json(result);
```
lukebarnard1: When the time comes when we do want some non-standard error codes, I'll be happy to create some way of sticking that in a result object. The twitter bridge does it by returning `{err: 404, body: "some message about not finding something"}`, and then scraping `err` to give to `res.status` and `body` to give to `res.json`.

lib/provisioning/Provisioner.js
Kegsay: Oh, it isn't? I didn't know that, what's the problem when you try to? Don't bother with making it a bound function if it's a hassle!

Kegsay:
LGTM.

PR #263, opened 2016-10-28
Allow topic setting /w mappings origin undefined

lukebarnard1:
Before 'origin' was introduced, entries were created without an origin. When setting a topic, allow them to be set on rooms with unknown origin as well as origin = 'join' | 'alias'.


Kegsay:
LGTM

PR #266, opened 2016-10-30
Make config schema path to be relative to installation directory.

OlegGirko:
This changed fixes wrong assumption that working directory of IRC bridge is always a directory where it's installed.

Schema path has been changed to be relative to installation directory, not working directory.


Kegsay:
LGTM - but can you please base this off `develop` and not `master` as per the [contrib guidelines](https://github.com/matrix-org/matrix-appservice-irc/blob/master/CONTRIBUTING.md#making-pull-requests) - Thanks!

OlegGirko:
I've rebased my change in `develop` branch and pushed it forcefully, but I found no way to change the base of pull request. So, I've created another pull request instead: #270.  Closing this one.

PR #270, opened 2016-10-31
Make config schema path to be relative to installation directory

OlegGirko:
This changed fixes wrong assumption that working directory of IRC bridge is always a directory where it's installed.

Schema path has been changed to be relative to installation directory, not working directory.


Kegsay:
LGTM - Thank you!

Issue #121, opened 2016-08-02
Initial membership list syncing gives up after 15mins


Kegsay:
This isn't always desirable if say, your DC is losing power for an hour. It leads to no M->I connections being made on startup.

This is a bit tricky to resolve because if we _do_ retry forever and **never succeed**, you kinda want to fail in some way but we can't. We can't just try once and then `exit(1)` because some people run startup scripts with this along with Synapse. I guess we should just try forever and ¯\_(ツ)_/¯

PR #275, opened 2016-11-03
Allow an infinite number of mlist sync requests

lukebarnard1:
This prevents giving up after 15mins, and instead continuing forever.

Should fix https://github.com/matrix-org/matrix-appservice-irc/issues/121


Kegsay:
LGTM

PR #276, opened 2016-11-04
Prevent races I->M with creating new PM rooms

lukebarnard1:
When creating a new PM room, immediately create a promise and store it in an object that maps each stored pair of user IDs (one virtual IRC user and one matrix user) to a promise that resolves when the pm room has been created.

### DIFF
--- a/spec/integ/pm.spec.js
+++ b/spec/integ/pm.spec.js
@@ -299,6 +299,49 @@ describe("IRC-to-Matrix PMing", function() {
         yield createRoomPromise;
         yield sentMessagePromise;
     }));
+
+    it("should not create multiple matrix rooms when several PMs are received in quick succession",
+    test.coroutine(function*() {
+        let count = 0;
+        // mock create room impl
+        let createRoomPromise = new Promise(function(resolve, reject) {
+            sdk.createRoom.andCallFake(function(opts) {
+                count++;
+                expect(count).toEqual(1);
+                resolve();
+                return Promise.resolve({
+                    room_id: tCreatedRoomId
+                });
+            });
+        });
+
+        // mock send message impl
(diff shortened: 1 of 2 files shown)


Kegsay:
Beautiful. LGTM! :heart:

Kegsay:
(Bonus points for tests :D)

lukebarnard1:
@Kegsay I'd quite like to test it but I'm not entirely sure how to test for the race or lack thereof.

Kegsay:
Just emit 2 messages from the IRC client in one turn of the event loop in your test. We have other race tests which do that.

lukebarnard1:
WFM

spec/integ/pm.spec.js
```diff
@@ -299,6 +299,49 @@ describe("IRC-to-Matrix PMing", function() {
...
+            roomMapping.server, tRealMatrixUserNick
+        );
+
+        // Send several messages, almost at once, to simulate a race
+        for (var i = 0; i < 10; i++) {
+            client.emit("message", tRealIrcUserNick, tRealMatrixUserNick, tText);
+        }
+
+        yield createRoomPromise;
+        yield sentMessagePromise;
```
Kegsay: We expect this to send **10** messages but we're only waiting for one. Why? This test can race in its current form because:
 - Send 10 messages.
 - #1 arrives
 - #2 arrives
 - #1 creates a room.
 - #1 sends the message.
 - PROMISES RESOLVED : TEST PASS :+1: 
 - #2 creates a room. `expect()` catches this but test has already passed so NOPs.
 - #2 sends the message.

Kegsay:
Other than racey test, LGTM.

Kegsay:
LGTM

PR #277, opened 2016-11-04
Use the home server domain in all templates

### LINKED ISSUE #190: Scope claimsXXX checks to the homserver domain
It's currently global which is far too loose, it means some user IDs across federation will be identified as virtual users belonging to the bridge.

lukebarnard1:
When two HSs with respective IRC bridges running, make sure that their templates for claiming things claim only on their respective HSs and not globally.

Aims to fix #190


Kegsay:
LGTM!

PR #279, opened 2016-11-07
Display join errors in admin room

### LINKED ISSUE #129: Send non-connection metadata messages when starting up
We currently filter ALL metadata messages I->M when the bridge is "starting up". This was done to avoid spamming on restarts with connection notices. However, some metadata messages are important to pass along, e.g. `err_badchannelkey` so people know that they need to re-issue the pass. We should not be filtering out those kinds of metadata messages.

lukebarnard1:
This works by ignoring startup suppression for errors that occur when joining a channel. When the bridge is restarted, users will be notified of all the channels that they must send a key to, using '!join ...'

Aims to fix #129


leonerd:
LGTM

PR #280, opened 2016-11-07
Randomly jitter quit debouncing between min, max

### LINKED ISSUE #274: QuitDebouncer needs to stagger when it eventually does quit
Else on a long netsplit, we will hammer synapse to death.  Have a min/max wait time and stagger between them, e.g. 45-60 mins.

lukebarnard1:
Add config to allow specification of jittering window.
When a QUIT occurs, jitter the timeout between a min and max time before bridging the leave.

Aims to fix #274

### DIFF
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -58,11 +58,17 @@ ircService:
         # The maximum number of quits per second acceptable above which a netsplit is
         # considered ongoing. Default: 5.
         quitsPerSecond: 5
-        # The time to wait before bridging a QUIT to Matrix that occured during a netsplit.
+        # The time window in which to wait before bridging a QUIT to Matrix that occured during
+        # a netsplit. Debouncing is jittered randomly between startMs and endMs so that the HS
+        # is not sent many requests to leave rooms all at once if a netsplit occurs and many
+        # people to not rejoin.
         # If the user with the same IRC nick as the one who sent the quit rejoins a channel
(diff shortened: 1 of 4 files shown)


leonerd:
The behaviour looks generally fine but I worry about the name of it. "jitter" usually refers to a measure of inevitable variable delay in a communication channel; i.e. something that just happens. Whereas here we're causing it. Can you rename it something like "stagger" or "delay"?

config.sample.yaml
```diff
@@ -58,11 +58,17 @@ ircService:
...
+        # a netsplit. Debouncing is jittered randomly between startMs and endMs so that the HS
+        # is not sent many requests to leave rooms all at once if a netsplit occurs and many
+        # people to not rejoin.
         # If the user with the same IRC nick as the one who sent the quit rejoins a channel
-        # before delayMs, consider them back online and do not bridge the quit.
+        # they are considered back online and the quit is not bridged, so long as the rejoin
+        # occurs before the randomly-jittered timeout is not reached.
         # Default: 3600000, = 1h
-        delayMs: 3600000 # 1h
+        startMs: 3600000 # 1h
```
leonerd: Maybe `delayMinMs` and `delayMaxMs` ?

lukebarnard1:
I think "jitter" describes it perfectly and is something that can be done on purpose.

leonerd:
LGTM

Issue #285, opened 2016-11-17
!quit is disabled because users wouldn't be kicked


lukebarnard1:
Tasks:

- [x] Re-enable the `!quit` command.
- [x] Kick the ghost matrix users associated with the user sending `!quit` from all bridged rooms. This might not be possible in provisioned room as the AS might not have power. Riot plumbed rooms are set up to give the bot power, however, so it will work with those at least.

Fixes regressions from https://github.com/matrix-org/matrix-appservice-irc/pull/253

PR #286, opened 2016-11-17
Try to kick bridged mx users on !quit

lukebarnard1:
Fixes https://github.com/matrix-org/matrix-appservice-irc/issues/285

When a !quit command is issued, make the bridge mx bot kick all instances of the bridged mx user from all bridged rooms. This currently excludes PM rooms and admin rooms and any rooms in which the user has an equal or greater power than the bot.

If users cannot be kicked from a room, log it but continue to disconnect the user from IRC.

### DIFF
--- a/lib/bridge/MatrixHandler.js
+++ b/lib/bridge/MatrixHandler.js
@@ -485,7 +485,10 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
         let bridgedClient = clients[0];
 
         if (bridgedClient.chanList.length === 0) {
-            req.log.info(`Bridged client for ${event.user_id} is not in any channels on ${ircServer.domain}`);
+            req.log.info(
+                `Bridged client for ${event.user_id} is not in any channels ` +
+                `on ${ircServer.domain}`
+            );
             let notice = new MatrixAction("notice", "You are not in any channels.");
             yield this.ircBridge.sendMatrixAction(adminRoom, botUser, notice, req);
         }
--- a/lib/bridge/MatrixHandler.js
+++ b/lib/bridge/MatrixHandler.js
@@ -468,11 +468,51 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
         return;
     }
     else if (cmd === "!quit") {
-        clientList.filter(
+        // Filter to get the clients for the [specified] server
+        let clients = clientList.filter(
             (bridgedClient) => bridgedClient.server.domain === ircServer.domain
-        ).forEach((bridgedClient) => {
+        );
+        if (clients.length === 0) {
+            req.log.info(`No bridgedClients for ${event.user_id}`);
+            let notice = new MatrixAction("notice", "You are not connected to any networks.");
+            yield this.ircBridge.sendMatrixAction(adminRoom, botUser, notice, req);
+        }
+        let getRoomsForChannel = (channel) => {
+            return this.ircBridge.getStore().getMatrixRoomsForChannel(ircServer, channel);
+        };
+        for (var i = 0; i < clients.length; i++) {
+            let bridgedClient = clients[i];
+            // Get all rooms that the bridgedClient is in
+            let rooms = yield Promise.all(bridgedClient.chanList.map(getRoomsForChannel));
+
+            // rooms is an array of arrays
+            rooms = rooms.reduce((a, b) => {return a.concat(b)});
+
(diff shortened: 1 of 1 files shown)


lib/bridge/MatrixHandler.js
```diff
@@ -468,11 +468,53 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
...
+        // Filter to get the clients for the [specified] server
+        let clients = clientList.filter(
+            (bridgedClient) => bridgedClient.server.domain === ircServer.domain
+        );
+        if (clients.length === 0) {
+            req.log.info(`No bridgedClients for ${event.user_id}`);
+            let notice = new MatrixAction("notice", "You are not connected to any networks.");
+            yield this.ircBridge.sendMatrixAction(adminRoom, botUser, notice, req);
+        }
+        let getRoomsForChannel = (channel) => {
```
Kegsay: You're using this in exactly one place. Please in-line this so I don't need to think where else you might call this.

lib/bridge/MatrixHandler.js
```diff
@@ -468,11 +468,53 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
...
-        // clientList.filter(
-        //     (bridgedClient) => bridgedClient.server.domain === ircServer.domain
-        // ).forEach((bridgedClient) => {
-        //     bridgedClient.kill();
-        // });
+        // Filter to get the clients for the [specified] server
+        let clients = clientList.filter(
+            (bridgedClient) => bridgedClient.server.domain === ircServer.domain
+        );
+        if (clients.length === 0) {
```
Kegsay: We should never have `> 1` client in this list. We should check for that and then just grab the first entry so we don't need to do loops over this 1-element list.

lib/bridge/MatrixHandler.js
```diff
@@ -468,11 +468,53 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
...
+        let getRoomsForChannel = (channel) => {
+            return this.ircBridge.getStore().getMatrixRoomsForChannel(ircServer, channel);
+        };
+        for (var i = 0; i < clients.length; i++) {
+            let bridgedClient = clients[i];
+            // Get all rooms that the bridgedClient is in
+            let rooms = yield Promise.all(bridgedClient.chanList.map(getRoomsForChannel));
+
+            // rooms is an array of arrays
+            rooms = rooms.reduce((a, b) => {return a.concat(b)});
```
Kegsay: Err, no it's not? https://github.com/matrix-org/matrix-appservice-irc/blob/master/lib/DataStore.js#L240

lib/bridge/MatrixHandler.js
```diff
@@ -468,11 +468,53 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
...
+
+            let uniqueRoomIds = [];
+            rooms.forEach((matrixRoom) => {
+                if (uniqueRoomIds.indexOf(matrixRoom.roomId) === -1) {
+                    uniqueRoomIds.push(matrixRoom.roomId);
```
Kegsay: We're using ES6. You want a `Set`.

lib/bridge/MatrixHandler.js
```diff
@@ -468,11 +468,53 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
...
+            catch (err) {
+                req.log.warn(
+                    `Could not kick ${bridgedClient.userId} from bridged rooms: ${err.message}`
+                );
+            }
+
+            req.log.info(
+                `Killing bridgedClient (nick = ${bridgedClient.nick}) for ${bridgedClient.userId}`
+            );
+            bridgedClient.kill();
```
Kegsay: No success message? IIRC the last time I queried this you said something along the lines of "the disconnect message will come through" - if so, please in-line comment this so I don't keep asking this question!

lib/bridge/MatrixHandler.js
lukebarnard1: Each channel is mapped to the array of rooms that it is bridged to.

lib/bridge/MatrixHandler.js
lukebarnard1: Yep, I'll comment.

Kegsay:
LGTM aside from inlining.

PR #290, opened 2016-11-23
V0.6.1

lukebarnard1:


CHANGELOG.md
```diff
@@ -1,3 +1,22 @@
+Changes in 0.6.1 (2016-11-23)
+=============================
+
+New features:
+ - Storing IRC Passwords Encrypted at Rest: No longer will users need to `IDENTIFY` with NickServ! Simply send the admin room command `!storepass [server.name] password`, and your password that you would normally send to NickServ will be stored in the bridge database in an encrypted form (and not as a hash). WARNING: The bridge needs to send plaintext passwords to the IRC server when the bridge is connecting the user to IRC (typically on startup). This means that the key used to decrypt passwords must be loaded into memory on startup, and the key file will typically be stored on the same machine as the bridge itself. `!removepass [server.name]` will remove the password that the user has set from the database.
```
Kegsay: Please break up this line to make it easier to read if you don't word wrap, and to make it easier to target line comments.

> Storing IRC Passwords Encrypted at Rest: No longer will users need to `IDENTIFY` with NickServ!

It feels like you've written this for the consumer (user) audience rather than the bridge admin audience so this reads weirdly.

I would keep it specific and boring: "Matrix users can now specify a [server password](https://en.wikipedia.org/wiki/List_of_Internet_Relay_Chat_commands#PASS) to authenticate with the IRC server on startup. On most IRC servers, this is an alternative mechanism to authenticate with `NickServ`. To enable this functionality in the bridge, a private key needs to be generated.

> WARNING: The bridge needs to send plaintext passwords to the IRC server when the bridge is connecting the user to IRC (typically on startup). This means that the key used to decrypt passwords must be loaded into memory on startup, and the key file will typically be stored on the same machine as the bridge itself.

This is confusing and warning about the wrong things. You need to be shouting about the fact that the bridge is forced to send the plaintext password and not a hash of the password to the IRC server, meaning that Matrix users are *trusting the bridge with their actual password*, which is very different to how password systems typically operate (which don't store the actual password, but a hash).


Overall this feature may warrant sub bullet points to break it up:
 - What it is
 - How does a sysadmin set it up
 - How does a matrix user use it
 - Risks involved

CHANGELOG.md
```diff
@@ -1,3 +1,22 @@
...
+
+Improvements:
+ - `!quit [server.name]` now attempts to kick the matrix user that issues the command from the rooms in which they are being briged. This is done after the user's respective IRC client is disconnected.
```
Kegsay: > This is done after the user's respective IRC client is disconnected.

[Lies.](https://github.com/matrix-org/matrix-appservice-irc/blob/67dead8fceda0006b7a69f99b1118f8b6e7ea658/lib/bridge/MatrixHandler.js#L501-L518)

CHANGELOG.md
```diff
@@ -1,3 +1,22 @@
...
+ - The bridge now randomly jitters quit debounce delays between a minimum and maximum amount of time. This is in order to prevent the HS being sent many part requests all at once following a net-split that lasts a very long time. (See `quitDebounce` in config.sample.yaml)
+ - Errors received by the bridge when _joining_ an IRC client to a channel can now be seen in the admin room at startup.
+ - The bridge will now only claim for the namespaces pertaining to the HS connected, rather than any HS (which might also have an IRC bridge running).
```
Kegsay: Wording feels wonky, you don't even mention whether this relates to users or rooms or both or neither. This is also a bug not an improvement.

> Previously the bridge may have treated users on other Homeservers as its own users. This meant they would not be bridged to IRC. The bridge will now respect the homeserver domain when determining whether it owns users or not.

README.md
```diff
@@ -24,7 +24,7 @@ you use this method, the bridge can be run via `node app.js`.
 
 
 ### Requirements
- - Node.js v4 or above.
+ - Node.js v4.5 or above.
```
lukebarnard1: I don't know, you committed it :stuck_out_tongue: https://github.com/matrix-org/matrix-appservice-irc/pull/290/commits/340b8b921b61328aa8075288fd9e2bbb203ddeb2

Kegsay: Aha, now I remember. Yay commit message.

Kegsay:
I would link to wikipedia for server password:
```
Matrix users can now specify a [server password](https://en.wikipedia.org/wiki/List_of_Internet_Relay_Chat_commands#PASS) to authenticate....
```

Other than that, LGTM!

Kegsay:
LGTM

PR #295, opened 2016-12-01
Leave matrix users if their associated IRC nick is no longer in a bridged channel

lukebarnard1:
On startup, matrix member lists are stored in memory for later reference. They may not be entirely accurate long after startup, but it won't stop the majority of users being cleaned up.

The ClientPool listens for any "irc-names" events being emitted from `BridgedClient`s. This event is triggered whenever an IRC client receives NAMES. The `ClientPool` calls `updateIrcMemberList` on `MemberListSyncer`. This function leaves any matrix users that do not appear to be in the IRC member list of the channel whose mlist has just been updated from all bridged rooms of that channel.

### DIFF
--- a/lib/bridge/MemberListSyncer.js
+++ b/lib/bridge/MemberListSyncer.js
@@ -15,6 +15,22 @@ function MemberListSyncer(ircBridge, appServiceBot, server, appServiceUserId, in
     this.appServiceUserId = appServiceUserId;
     this.injectJoinFn = injectJoinFn;
     this._syncableRoomsPromise = null;
+    this._memberLists = {
+        matrix: {
+            //$roomId : [
+            //    {
+            //        id: roomId,
+            //        state: stateEvents,
+            //        realJoinedUsers: [],
+            //        remoteJoinedUsers: []
+            //    },
+            //    ...
+            //  ]
+        },
+        irc: {
+            //$channel : nick[]
+        }
+    }
 }
 
 MemberListSyncer.prototype.sync = Promise.coroutine(function*() {
@@ -31,6 +47,8 @@ MemberListSyncer.prototype.sync = Promise.coroutine(function*() {
     let start = Date.now();
     let rooms = yield this._getSyncableRooms(server);
     log.info("Found %s syncable rooms (%sms)", rooms.length, Date.now() - start);
+    log.info("Marking Matrix rooms for cleanup when updateIrcMemberList called...");
+    yield this.leaveIrcUsersFromRooms(rooms, server);
     start = Date.now();
     log.info("Joining Matrix users to IRC channels...");
     yield joinMatrixUsersToChannels(rooms, server, this.injectJoinFn);
@@ -38,10 +56,6 @@ MemberListSyncer.prototype.sync = Promise.coroutine(function*() {
     // NB: We do not need to explicitly join IRC users to Matrix rooms
     // because we get all of the NAMEs/JOINs as events when we connect to
     // the IRC server. This effectively "injects" the list for us.
-    start = Date.now();
-    log.info("Leaving IRC users from Matrix rooms (cleanup)...");
-    yield leaveIrcUsersFromRooms(rooms, server);
-    log.info("Left IRC users from Matrix rooms. (%sms)", Date.now() - start);
 });
 
 MemberListSyncer.prototype.getChannelsToJoin = Promise.coroutine(function*() {
@@ -129,6 +143,16 @@ MemberListSyncer.prototype.checkBotPartRoom = Promise.coroutine(function*(ircRoo
 // so we don't do it unless they need absolutely fresh data. On startup, this can be called
 // multiple times, so we cache the first request's promise and return that instead of making
 // double hits.
+//
+// returns [
+//   {
+//       id: roomId,
+//       state: stateEvents,
+//       realJoinedUsers: [],
+//       remoteJoinedUsers: []
+//   },
+//   ...
+// ]
 MemberListSyncer.prototype._getSyncableRooms = function(server, ignoreCache) {
     if (!ignoreCache && this._syncableRoomsPromise) {
         log.debug("Returning existing _getSyncableRooms Promise");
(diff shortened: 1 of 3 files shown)


Kegsay:
Failing tests.

lib/bridge/MemberListSyncer.js
```diff
@@ -31,17 +47,15 @@ MemberListSyncer.prototype.sync = Promise.coroutine(function*() {
     let start = Date.now();
     let rooms = yield this._getSyncableRooms(server);
     log.info("Found %s syncable rooms (%sms)", rooms.length, Date.now() - start);
+    log.info("Marking Matrix rooms for cleanup when updateIrcMemberList called...");
```
Kegsay: This log line adds nothing. Remove it please.

lib/bridge/MemberListSyncer.js
```diff
@@ -247,10 +271,63 @@ function joinMatrixUsersToChannels(rooms, server, injectJoinFn) {
     return d.promise;
 }
 
-function leaveIrcUsersFromRooms(rooms, server) {
+MemberListSyncer.prototype.leaveIrcUsersFromRooms = function(rooms, server) {
+    log.info("leaveIrcUsersFromRooms");
```
Kegsay: Add some more information here for it to be useful: perhaps the number of rooms and the server domain.

lib/bridge/MemberListSyncer.js
```diff
@@ -15,6 +15,22 @@ function MemberListSyncer(ircBridge, appServiceBot, server, appServiceUserId, in
     this.appServiceUserId = appServiceUserId;
     this.injectJoinFn = injectJoinFn;
     this._syncableRoomsPromise = null;
+    this._memberLists = {
+        matrix: {
+            //$roomId : [
+            //    {
```
Kegsay: Lies? It looks like it's `$roomId: {}` rather than `$roomId: [ {}, {} ]`

lib/bridge/MemberListSyncer.js
```diff
@@ -247,10 +271,63 @@ function joinMatrixUsersToChannels(rooms, server, injectJoinFn) {
     return d.promise;
 }
 
-function leaveIrcUsersFromRooms(rooms, server) {
+MemberListSyncer.prototype.leaveIrcUsersFromRooms = function(rooms, server) {
+    log.info("leaveIrcUsersFromRooms");
+
+    // Store the matrix room info in memory for later retrieval when NAMES is received
+    // and updateIrcMemberList is called. At that point, we have enough information to
+    // leave users from the channel that the NAMES is for.
```
Kegsay: :tada: good comment

lib/irc/ClientPool.js
```diff
@@ -314,4 +315,9 @@ ClientPool.prototype._onJoinError = Promise.coroutine(function*(bridgedClient, c
     yield Promise.all(promises);
 });
 
+ClientPool.prototype._onNames = Promise.coroutine(function*(bridgedClient, chan, names) {
+    let mls = this._ircBridge.memberListSyncers[bridgedClient.server.domain];
+    yield mls.updateIrcMemberList(chan, names);
```
Kegsay: What if member list syncing is disabled on `server`? Need null check.

lib/bridge/MemberListSyncer.js
```diff
@@ -247,10 +271,63 @@ function joinMatrixUsersToChannels(rooms, server, injectJoinFn) {
...
+    });
+
     return Promise.resolve();
 }
 
+// Update the MemberListSyncer with the IRC NAMES_RPL that has been received for channel.
+// This will leave any matrix users that do not have their associated IRC nick in the list
+// of names for this channel.
+MemberListSyncer.prototype.updateIrcMemberList = Promise.coroutine(function*(channel, names) {
+    log.info(`updateIrcMemberList: Updating IRC member list for ${channel}`);
```
Kegsay: Careful. The bridge is [configurable](https://github.com/matrix-org/matrix-appservice-irc/blob/master/config.sample.yaml#L170) as to whether this should be done, and you aren't checking for it.

It might be an idea to also log the *number* of names so we have some knowledge whether we're about to kill Synapse or not :D

lib/bridge/MemberListSyncer.js
```diff
@@ -247,10 +271,66 @@ function joinMatrixUsersToChannels(rooms, server, injectJoinFn) {
...
+        log.info(`updateIrcMemberList: No bridged rooms for channel ${channel}`);
+        return;
+    }
+
+    // If a userId is in remoteJoinedUsers, but not ircUserIds, intend on leaving roomId
+    yield Promise.all(roomsForChannel.map((matrixRoom) => {
+        return matrixRoom.getId();
+    }).filter((roomId) => { // Filter rooms to remove those without member list data stored
+        return this._memberLists.matrix[roomId] &&
+            this._memberLists.matrix[roomId].remoteJoinedUsers;
```
Kegsay: `_getSyncableRooms` should've already filtered this out.

lib/bridge/MemberListSyncer.js
lukebarnard1: > Careful. The bridge is configurable as to whether this should be done, and you aren't checking for it.

I did think about that, but is this "incremental" or "initial"? The bridge will receive NAMES potentially at any time.

Kegsay: > is this "incremental" or "initial"?

This is initial. We only care about the first NAMES we get, and then we process it and mark that channel as synced, right?

lib/bridge/MemberListSyncer.js
```diff
@@ -15,6 +15,19 @@ function MemberListSyncer(ircBridge, appServiceBot, server, appServiceUserId, in
...
+        matrix: {
+            //$roomId : {
+            //    id: roomId,
+            //    state: stateEvents,
+            //    realJoinedUsers: [],
+            //    remoteJoinedUsers: []
+            //  }
+        },
+        irc: {
+            //$channel : nick[]
```
Kegsay: Thinking about this a bit more: do we stand the chance of racing with ourselves? What happens if the NAMES includes one of the bridge's nicks? We obviously shouldn't be trying to make a matrix entry for that nick. Are we checking for this?

lib/irc/ClientPool.js
```diff
@@ -314,4 +315,15 @@ ClientPool.prototype._onJoinError = Promise.coroutine(function*(bridgedClient, c
     yield Promise.all(promises);
 });
 
+ClientPool.prototype._onNames = Promise.coroutine(function*(bridgedClient, chan, names) {
+    let mls = this._ircBridge.memberListSyncers[bridgedClient.server.domain];
+    if (!mls) {
+        log.info(`MemberListSyncer does not exist for ${bridgedClient.server.domain}: ` +
+            `not calling updateIrcMemberList and not syncing leaves (for channel ${chan}).`
```
Kegsay: I wouldn't log this at all. This will be spammy (every time someone joins a channel a NAMES comes down) for users who aren't syncing membership lists.

lib/bridge/MemberListSyncer.js
```diff
@@ -247,10 +267,86 @@ function joinMatrixUsersToChannels(rooms, server, injectJoinFn) {
...
+    });
 }
 
+// Update the MemberListSyncer with the IRC NAMES_RPL that has been received for channel.
+// This will leave any matrix users that do not have their associated IRC nick in the list
+// of names for this channel.
+MemberListSyncer.prototype.updateIrcMemberList = Promise.coroutine(function*(channel, names) {
+    log.info(`updateIrcMemberList: Updating IRC member list for ${channel}`);
+    if (this._memberLists.irc[channel] !== undefined) {
+        log.info(`Names already received for ${channel}`);
```
Kegsay: Any reason to not just be using a truthy value for this?

lib/bridge/MemberListSyncer.js
```diff
@@ -247,10 +267,86 @@ function joinMatrixUsersToChannels(rooms, server, injectJoinFn) {
...
+        return;
+    }
+    if (!this.server.shouldSyncMembershipToMatrix("initial", channel)) {
+        log.info(`Initial membership list syncing disabled for ${channel} and/or globally`);
```
Kegsay: Again, this will probably be a bit too spammy for this to be useful.

lib/bridge/MemberListSyncer.js
```diff
@@ -247,10 +267,86 @@ function joinMatrixUsersToChannels(rooms, server, injectJoinFn) {
...
+        return;
+    }
+
+    this._memberLists.irc[channel] = Object.keys(names);
+
```
Kegsay: I would log something on this line to say how many IRC users you're looking at, what channel/server, etc *before* you do all the work so we can gauge how long it takes and can correlate logs if something below goes wrong.

lib/bridge/MemberListSyncer.js
```diff
@@ -247,10 +267,86 @@ function joinMatrixUsersToChannels(rooms, server, injectJoinFn) {
...
+        log.info(
+            `updateIrcMemberList: Leaving ${userIdsToLeave.length} users from ${roomId} ` +
+            `because they are not in ${channel}.`
+        );
+        return userIdsToLeave.map(
+            (userId) => {
+                return this.ircBridge.getAppServiceBridge().getIntent(userId).leave(roomId);
+            }
+        );
+    }).reduce((a, b) => a.concat(b));
```
Kegsay: Still feels too complicated to me: you have a map -> filter -> map -> reduce -> concat, whereas what you're really just doing is:
```js
let promises = [];
bridgedRoomIds.forEach((roomId) => {
    this._memberLists.matrix[roomId].remoteJoinedUsers.forEach((userId) => {
        if (ircUserIds.indexOf(userId) === -1) {
            promises.push(this.ircBridge.getAppServiceBridge().getIntent(userId).leave(roomId));
        }
    });
});
log.info(
    `updateIrcMemberList: Leaving ${userIdsToLeave.length} users from ${roomId} ` +
    `because they are not in ${channel}.`
);
yield Promise.all(promises);
```

Not only is this about the same lines of code, it's more performant (fewer loops) and IMHO clearer to see what you're manipulating.

lib/bridge/MemberListSyncer.js
lukebarnard1: (Done)

lib/bridge/MemberListSyncer.js
```diff
@@ -247,10 +267,73 @@ function joinMatrixUsersToChannels(rooms, server, injectJoinFn) {
...
+    });
 }
 
+// Update the MemberListSyncer with the IRC NAMES_RPL that has been received for channel.
+// This will leave any matrix users that do not have their associated IRC nick in the list
+// of names for this channel.
+MemberListSyncer.prototype.updateIrcMemberList = Promise.coroutine(function*(channel, names) {
+    log.info(`updateIrcMemberList: Updating IRC member list for ${channel}`);
+    if (this._memberLists.irc[channel] !== undefined ||
+        !this.server.shouldSyncMembershipToMatrix("initial", channel)) {
```
Kegsay: Indent moar please!

Kegsay:
LGTM!

lukebarnard1:
Merged into v0.6.1 ( #290 )

PR #297, opened 2016-12-05
Listen for err_yourebannedcreep and wait 6hrs before reconnecting

Kegsay:
When `err_yourebannedcreep` is received, 2 events fire: `error` and `raw`. It
looks like `error` consistently happens before `raw`, which would lead to
`disconnect(irc_error)` which then flips the connection state to `dead`. When
the `raw` event fires, it tries to fail with `disconnect(banned)` but because
we've already flipped the connection state to `dead`, this does nothing.

As a consequence, banned connections would retry much more frequently than they
should.

This PR fixes this by looking for "banned" reasons in the `error` event, and failing
appropriately.

### DIFF
--- a/lib/irc/ConnectionInstance.js
+++ b/lib/irc/ConnectionInstance.js
@@ -172,7 +172,7 @@ ConnectionInstance.prototype._listenForErrors = function() {
                 return; // don't disconnect for these error codes.
             }
         }
-        if (err && err.command && err.command === "err_yourebannedcreep") {
+        if (err && err.command === "err_yourebannedcreep") {
             self.disconnect("banned");
             return;
         }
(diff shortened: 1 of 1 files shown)


lib/irc/ConnectionInstance.js
```diff
@@ -172,6 +172,10 @@ ConnectionInstance.prototype._listenForErrors = function() {
                 return; // don't disconnect for these error codes.
             }
         }
+        if (err && err.command && err.command === "err_yourebannedcreep") {
```
lukebarnard1: I'm not sure a truthy check of `err.command` is necessary here given the `===` check with the string.

Kegsay: Done

Issue #296, opened 2016-12-02
Add Prometheus-style metrics


leonerd:
(This is somewhat of a notes-to-self whiteboard to keep track of work on https://github.com/matrix-org/matrix-appservice-irc/tree/paul/metrics)

In particular, as well as adding whatever the bridge library has, it would be nice to preserve the values of the current statsd-based reporting. The four variables are:

    ircas_connected_clients_gauge{}

    ircas_membership_counter{protocol,state}

    ircas_memory_gauge

    ircas_requests_timer{direction,outcome}
    ircas_requests_timer_count{...}
    ircas_requests_timer_sum{...}

leonerd:
> ircas_memory_gauge
Already handled by the prometheus client library, as `process_memory_*`

leonerd:
> ircas_connected_clients_gauge

Well this is a fun one. The underlying `statsd`-exporting code in the IRC bridge exports this metric on a per-domain basis, but a bug in the statsd-to-prometheus bridge config (https://github.com/matrix-org/synapse-prometheus-config/blob/master/irc-as/statsd-prometheus-mappings.cfg#L14) currently means that only one of these metrics will turn up in prometheus, and not annotated by a domain name. This doesn't affect us in practice because we only run one domain per bridge instance anyway.

a75cb6a adds code to export this, as the bridge-standard metric of `bridge_remote_ghosts`

leonerd:
> ircas_membership_counter{protocol,state}

Having IRL'ed this with @Kegsay as to the intention of it, it seems it can be dropped.

The intention is to measure the rate of *outbound* join and part requests the bridge makes to IRC and Matrix. On the Matrix side we already have the `bridge_matrix_api_calls` counter which is automatically managed by the matrix SDK, and adding the usual complementary `bridge_remote_api_calls` counter to the IRC side in the bridge will cover it.

leonerd:
> ircas_requests_timer{direction,outcome}
> ircas_requests_timer_count{...}
> ircas_requests_timer_sum{...}

These are supposed to track the total "end-to-end" lag of events within the bridge. The main bridge library doesn't have a direct equivalent for these yet, but they'd also be useful to add generally - gitter and slack can make equally-good use of it.

*However*...

This metric is a Histogram, and the prometheus library we're currently using (`npm prometheus-client`) doesn't actually support those. To get that we'd want to switch to `prom-client`; which I think would be good anyway as that generally seems to be a better library.

[...]

PR #299, opened 2016-12-08
Export prometheus-style metrics corresponding to existing 'statsd' ones

leonerd:
This PR doesn't yet attempt to address all of the "bridge gauge" metrics, only enough to cover the same statistics as exported by the current `statsd`-based code.

Part of #296

### DIFF
--- a/lib/irc/ClientPool.js
+++ b/lib/irc/ClientPool.js
@@ -208,11 +208,8 @@ ClientPool.prototype.countTotalConnections = function() {
     var count = 0;
 
     Object.keys(this._virtualClients).forEach((domain) => {
-        var clients = this._virtualClients[domain];
-
-        Object.keys(clients).forEach((nick) => {
-            if (clients[nick]) { count++ }
-        });
+        let server = this._ircBridge.getServer(domain);
+        count += this._getNumberOfConnections(server);
     });
 
     return count;
--- a/lib/irc/ClientPool.js
+++ b/lib/irc/ClientPool.js
@@ -208,7 +208,7 @@ ClientPool.prototype.countTotalConnections = function() {
     var count = 0;
 
     Object.keys(this._virtualClients).forEach((domain) => {
-        var clients = this._virtualClients;
+        var clients = this._virtualClients[domain];
 
         Object.keys(clients).forEach((nick) => {
             if (clients[nick]) { count++ }
--- a/lib/irc/ClientPool.js
+++ b/lib/irc/ClientPool.js
@@ -204,6 +204,20 @@ ClientPool.prototype._getNumberOfConnections = function(server) {
     return numConnectedNicks;
 };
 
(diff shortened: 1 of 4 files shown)


lib/irc/ClientPool.js
```diff
@@ -204,6 +204,20 @@ ClientPool.prototype._getNumberOfConnections = function(server) {
     return numConnectedNicks;
 };
 
+ClientPool.prototype.countTotalConnections = function() {
+    var count = 0;
+
+    Object.keys(this._virtualClients).forEach((domain) => {
+        var clients = this._virtualClients;
```
Kegsay: Presumably meant `this._virtualClients[domain]`

leonerd: Er.. wow. I'm surprised this works. Indeed I did. ((also: I really want `Object.values()`))

leonerd: Fixed e0600a2

lib/irc/ClientPool.js
```diff
@@ -204,6 +204,20 @@ ClientPool.prototype._getNumberOfConnections = function(server) {
     return numConnectedNicks;
 };
 
+ClientPool.prototype.countTotalConnections = function() {
+    var count = 0;
+
+    Object.keys(this._virtualClients).forEach((domain) => {
```
Kegsay: I'd prefer it if you used `_getNumberOfConnections` instead, given you are reproducing the same code and they are arguably doing the same thing (albeit one has a `server` and the other does not).

You can get servers from a domain via https://github.com/matrix-org/matrix-appservice-irc/blob/0f82ec6599546473672ea84d2fe439207b0eba9a/lib/bridge/IrcBridge.js#L648

Kegsay: So the total code would be:

```js
Object.keys(this._virtualClients).forEach((domain) => {
    let server = this. _ircBridge.getServer(domain);
    count += this. _getNumberOfConnections(server);
});

Kegsay:
Otherwise LGTM!

lib/irc/ClientPool.js
leonerd: So, you want me to iterate the list of domain names in the map, converting each one into a server instance object (a process itself that involves a linear scan in the code linked above), just so I can pass the server instance into `_getNumberOfConnections` which will turn it back into a domain name to use as the key into the original map I iterated over in the first place?

Kegsay: In 99% of cases, there will be precisely 1 server, so linear scans are not a problem. I agree the back and forth from a `server` is excessive, so am happy for you to make `_getNumberOfConnections` take a server domain instead of a server.

leonerd: Done 96334ca

Kegsay:
LGTM

PR #305, opened 2016-12-12
Add admin scripts which we use at matrix.org

Kegsay:
Tested on: Python 2.7, but should probably work on 3.

### DIFF
--- a/scripts/remove-user.py
+++ b/scripts/remove-user.py
@@ -15,11 +15,11 @@ def get_appservice_token(reg):
 def get_rooms(hs_url, token, user_id):
     res = requests.get(hs_url + "/_matrix/client/r0/sync?access_token=" + token).json()
     room_ids = []
-    for r in res["rooms"]["join"]:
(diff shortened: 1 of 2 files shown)


scripts/remove-user.py
```diff
@@ -0,0 +1,56 @@
...
+
+def get_appservice_token(reg):
+    with open(reg, "r") as f:
+        reg_yaml = yaml.load(f)
+        return reg_yaml["as_token"]
+
+def get_rooms(hs_url, token, user_id):
+    res = requests.get(hs_url + "/_matrix/client/r0/sync?access_token=" + token).json()
+    room_ids = []
+    for r in res["rooms"]["join"]:
```
lukebarnard1: Perhaps rename `r` to `roomId`

lukebarnard1: `room_id` even

Kegsay: Done

PR #306, opened 2016-12-12
Use AS-specific room publication API

lukebarnard1:
Corresponds to the changes outlined here: matrix-org/synapse#1676

- `network_id` (which is an IRC `server.domain`) has been added to each instance returned by `/thirdparty/protocols`.
- When room visibility is changed, the new endpoint specifically for ASes is used, which specifies the `network_id` to be the `server.domain` of the first bridged channel in the list of channels mapped to that room.

This requires https://github.com/matrix-org/matrix-js-sdk/pull/306

### DIFF
--- a/lib/DataStore.js
+++ b/lib/DataStore.js
@@ -170,7 +170,7 @@ DataStore.prototype.getAllChannelMappings = Promise.coroutine(function*() {
             (e) => e.matrix_id === roomId
         ).map((e) => {
             return {
-                networkId: this._serverMappings[e.remote.domain],
+                networkId: this._serverMappings[e.remote.domain].getNetworkId(),
                 channel: e.remote.channel
             }
         });
--- a/lib/DataStore.js
+++ b/lib/DataStore.js
@@ -170,7 +170,7 @@ DataStore.prototype.getAllChannelMappings = Promise.coroutine(function*() {
             (e) => e.matrix_id === roomId
         ).map((e) => {
             return {
-                domain: e.remote.domain,
+                networkId: this._serverMappings[e.remote.domain],
                 channel: e.remote.channel
             }
         });
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -560,7 +560,7 @@ IrcBridge.prototype.getThirdPartyProtocol = function(protocol) {
         },
         instances: servers.map((server) => {
             return {
-                network_id: server.config.name || server.domain,
+                network_id: server.domain,
                 bot_user_id: this.getAppServiceUserId(),
                 desc: server.config.name || server.domain,
                 icon: server.config.icon,
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -560,7 +560,7 @@ IrcBridge.prototype.getThirdPartyProtocol = function(protocol) {
         },
         instances: servers.map((server) => {
             return {
-                network_id: server.domain,
+                network_id: server.getNetworkId(),
                 bot_user_id: this.getAppServiceUserId(),
                 desc: server.config.name || server.domain,
                 icon: server.config.icon,
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -560,6 +560,7 @@ IrcBridge.prototype.getThirdPartyProtocol = function(protocol) {
         },
         instances: servers.map((server) => {
             return {
+                network_id: server.config.name || server.domain,
                 bot_user_id: this.getAppServiceUserId(),
                 desc: server.config.name || server.domain,
                 icon: server.config.icon,
--- a/lib/bridge/IrcHandler.js
+++ b/lib/bridge/IrcHandler.js
@@ -571,7 +571,7 @@ IrcHandler.prototype.onMode = Promise.coroutine(function*(req, server, channel, 
             req.log.info("Not syncing publicity: shouldPublishRooms is false");
             return Promise.resolve();
         }
-        const key = this.ircBridge.publicitySyncer.getIRCVisMapKey(server, channel);
+        const key = this.ircBridge.publicitySyncer.getIRCVisMapKey(server.getNetworkId(), channel);
 
         // Update the visibility for all rooms connected to this channel
         return this.ircBridge.publicitySyncer.updateVisibilityMap(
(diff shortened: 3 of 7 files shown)


lib/bridge/PublicitySyncer.js
```diff
@@ -181,8 +181,12 @@ PublicitySyncer.prototype._solveVisibility = Promise.coroutine(function*() {
         let currentState = this._visibilityMap.roomVisibilities[roomId];
         let correctState = shouldBePrivate(roomId, []) ? 'private' : 'public';
 
+        // Use the server domain of the first mapping
+        // 'localhost #channel1' => 'localhost'
+        const networkId = this._visibilityMap.mappings[roomId][0].split(' ')[0];
```
Kegsay: Is `this._visibilityMap.mappings[roomId]` **guaranteed** to be non-null? We seem to do a null guard on `:140` but we don't here. But `:127` makes me think that yes, it is guaranteed to be non-null. But then why the null guard on `:140`? Some clarification here would be good.

Kegsay: Also, ignoring the null guard issue, will there always be a mapped channel for this room ID? Now that I think is no, which would then NPE on this line.

Kegsay:
Otherwise LGTM

lib/bridge/IrcBridge.js
```diff
@@ -560,6 +560,7 @@ IrcBridge.prototype.getThirdPartyProtocol = function(protocol) {
         },
         instances: servers.map((server) => {
             return {
+                network_id: server.domain,
```
erikjohnston: Can this change? e.g. if we change the domain we use to connect to a network? Its important that that this remains constant unless the AS republishes all the rooms after a change.

lib/bridge/PublicitySyncer.js
```diff
@@ -181,8 +181,12 @@ PublicitySyncer.prototype._solveVisibility = Promise.coroutine(function*() {
...
+
         if (currentState !== correctState) {
-            return cli.setRoomDirectoryVisibility(roomId, correctState).then(
+            return cli.setRoomDirectoryVisibilityAppService(networkId, roomId, correctState).then(
```
lukebarnard1: Much in the same way it would work for those not in the main list.

lib/bridge/IrcBridge.js
Kegsay: It's rare, but it can change like when we hop to IPv6 from IPv4. We may want to indirect here (e.g. have a `network_id` value in the config file, but default to `server.domain` if one is not supplied).

This becomes relevant when we inevitably move Moznet to v6 (it's still v4 atm).

lib/bridge/PublicitySyncer.js
lukebarnard1: The null guard shouldn't be needed, you're right. And yes the mappings are guaranteed to be set for a room, and contain at least one channel mapping otherwise that room wouldn't be in the `RoomStore`

lib/bridge/IrcBridge.js
lukebarnard1: I had totally forgotten about this. I shall add `network_id` to the config.

lib/irc/IrcServer.js
```diff
@@ -34,6 +34,14 @@ IrcServer.prototype.getReadableName = function() {
 }
 
 /**
+ * Return the network ID of this server, which should be unique for this
+ * bridge across all servers/instances. Defaults to the domain of the server.
+ */
+IrcServer.prototype.getNetworkId = function() {
+    return this.config.networkId || server.domain;
```
Kegsay: What's a `server`? :)

lib/irc/IrcServer.js
```diff
@@ -34,6 +34,14 @@ IrcServer.prototype.getReadableName = function() {
 }
 
 /**
+ * Return the network ID of this server, which should be unique for this
+ * bridge across all servers/instances. Defaults to the domain of the server.
+ */
```
Kegsay: JSDoc please.

lib/DataStore.js
```diff
@@ -170,7 +170,7 @@ DataStore.prototype.getAllChannelMappings = Promise.coroutine(function*() {
             (e) => e.matrix_id === roomId
         ).map((e) => {
             return {
-                domain: e.remote.domain,
+                networkId: this._serverMappings[e.remote.domain],
```
Kegsay: Updated the response object without updating docs. Please update docs.

lib/bridge/IrcHandler.js
```diff
@@ -573,7 +573,7 @@ IrcHandler.prototype.onMode = Promise.coroutine(function*(req, server, channel,
         }
         // Update the visibility for all rooms connected to this channel
         return this.ircBridge.publicitySyncer.updateVisibilityMap(
-            true, server.domain + ' ' + channel, enabled
+            true, server.getNetworkId() + ' ' + channel, enabled
```
Kegsay: I really don't like this. I didn't say anything at the time but I'm in a less accepting mood today. I appreciate that the publicity syncer needs to key off a tuple of network ID + channel. I don't feel it is the caller's responsibility to craft the unique key. We should be calling a method which accepts a `server` and `channel` and returns the key, so we can consistently apply the concatenation rules. Otherwise, as I'm sure you now appreciate, you need to find alllll the places where we make the key and change them in the same way in order to apply a fix, which makes it really really easy to miss one place. See https://github.com/matrix-org/matrix-appservice-irc/blob/master/lib/DataStore.js#L484 for an example.

lib/bridge/PublicitySyncer.js
```diff
@@ -125,7 +140,7 @@ PublicitySyncer.prototype._solveVisibility = Promise.coroutine(function*() {
 
     roomIds.forEach((roomId) => {
         this._visibilityMap.mappings[roomId] = mappings[roomId].map((mapping) => {
-            return mapping.domain + ' ' + mapping.channel
+            return mapping.networkId + ' ' + mapping.channel
```
Kegsay: Shouldn't we be calling `getIRCVisMapKey`?

Kegsay:
Otherwise LGTM

lib/bridge/PublicitySyncer.js
```diff
@@ -140,7 +141,7 @@ PublicitySyncer.prototype._solveVisibility = Promise.coroutine(function*() {
 
     roomIds.forEach((roomId) => {
         this._visibilityMap.mappings[roomId] = mappings[roomId].map((mapping) => {
-            return mapping.networkId + ' ' + mapping.channel
+            return getIRCVisMapKey(mapping.networkId, mapping.channel);
```
Kegsay: You forgot a `this.`

Issue #303, opened 2016-12-11
initial /sync sometimes fails


ara4n:
looks like we have a 5 minute timeout somewhere on /sync responses (in dendron, perhaps?). if the final attempt to /sync completes close to the 5 minute mark, then the response may get truncated, which the ircas seems to silently ignore and thus fail to run.  Need to increase the timeout, use a better LB, or move away from /sync.  Also unclear why this wouldn't affect normal clients too (or perhaps it does)

Kegsay:
Interesting! So *that's* why it sometimes seems to give a valid response with 0 rooms. This will hopefully go away when https://github.com/matrix-org/synapse/pull/1680 lands on matrix.org

Kegsay:
We've moved away from `/sync`.

Issue #288, opened 2016-11-21
Freenode bridge outage (19-21 Nov)


Kegsay:
This is a post-mortem post of an outage over the weekend. This issue has been *mitigated* for now; there are a number of outstanding performance issues to address before this can be closed.

---

### Summary of the outage

*All times GMT.*

Nov 19 / 17:40 : Prometheus graphs show a consistent spike in requests to and from IRC being delayed (>10s to be passed through the bridge).
Nov 19 / 18:40 : Prometheus graphs show a spike in request failures to and from IRC.
Nov 19 / 19:48 : The IRC bridge is restarted. It begins to spin on restarts.
Nov 19 / 20:15 : The IRC bridge is restarted with membership list syncing disabled.
Nov 21 / 10:20 : The IRC bridge is restarted with membership list syncing enabled again.

### Initial mitigations

The restart at 19:48 was done by @leonerd who noticed it then started to spin:

```
<LeoNerd> It's getting stuck
<LeoNerd> 'main' terminated and then it restarted. it seems to be spinning
<LeoNerd> 2016-11-19 20:07:54 ERROR:main FATAL EXCEPTION
2016-11-19 20:07:54 ERROR:main Error: toString failed
    at Buffer.toString (buffer.js:448:11)
    at BufferList.toString (/mnt/disk/home/ircas/matrix-org/freenode/matrix-appservice-irc/node_modules/request/node_modules/bl/bl.js:178:33)
    at Request.<anonymous> (/mnt/disk/home/ircas/matrix-org/freenode/matrix-appservice-irc/node_modules/request/request.js:1023:32)
    at emitOne (events.js:77:13)
    at Request.emit (events.js:169:7)
    at IncomingMessage.<anonymous> (/mnt/disk/home/ircas/matrix-org/freenode/matrix-appservice-irc/node_modules/request/request.js:969:12)
    at emitNone (events.js:72:20)
    at IncomingMessage.emit (events.js:166:7)
    at endReadableNT (_stream_readable.js:921:12)
    at nextTickCallbackWith2Args (node.js:442:9)
    at process._tickCallback (node.js:356:17)
2016-11-19 20:07:54 ERROR:main Terminating (exitcode=1)
<LeoNerd> It's done that 4 times in a row now and just keeps doing it
```

I stepped in to put a plaster on it by blindly applying a try catch block to the `bl` library. 

```
<kegan> have blindly wrapped it in a try catch
```

[...]

PR #308, opened 2016-12-12
Use the new membership list APIs instead of /sync

Kegsay:
/sync is slow (#288) and unreliable (#303) when the response is large. Use
dedicated membership list APIs instead.

Currently, this PR is just a literal translation of the new APIs from /sync.
To make full use of this API, we need to change when we begin processing the
results, as we can now do so incrementally.

### DIFF
--- a/lib/bridge/MemberListSyncer.js
+++ b/lib/bridge/MemberListSyncer.js
@@ -162,20 +162,55 @@ MemberListSyncer.prototype._getSyncableRooms = function(server, ignoreCache) {
         let attempts = 0;
         while (true) { // eslint-disable-line no-constant-condition
             try {
-                // roomDict = { room_id: RoomInfo }
-                let roomDict = yield self.appServiceBot.getMemberLists();
-                return Object.keys(roomDict).map(function(roomId) {
-                    return roomDict[roomId];
-                }).filter(function(roomInfo) {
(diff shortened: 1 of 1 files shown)


lib/bridge/MemberListSyncer.js
```diff
@@ -162,20 +162,57 @@ MemberListSyncer.prototype._getSyncableRooms = function(server, ignoreCache) {
...
+                                    display_name: userMap[userId].display_name,
+                                    avatar_url: userMap[userId].avatar_url,
+                                }
+                            });
+                        }
+                        roomInfoList.push(roomInfo);
+                    }
+                    catch (err) {
+                        log.error(`Failed to getJoinedMembers in room ${roomId}: ${err}`);
+                        i--; // try again
```
lukebarnard1: Retrying forever on a room that consistently produces errors should probably be avoided. Although, I can't think of a good reason why one room would fail, despite it technically being possible. WDYT?

Kegsay: It mimics the same behaviour as `/sync` - if that fails we keep trying ad infinitum. We just extend that to all the tiny requests we're now making. In the future we can be smarter about it somehow.

lukebarnard1: ok fair enough

PR #310, opened 2016-12-14
Add config option for dropping old Matrix messages

### LINKED ISSUE #301: After a big outage we should drop history on the floor
The queued up M->I traffic is incredibly confusing and obnoxious

Kegsay:
With tests. I also tweaked the config schema for the `homeserver` - `appService` has not been a valid config option for a very long time!

This only concerns **messages** as we still always want to sync membership list states.

Fixes #301

### DIFF
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -11,6 +11,18 @@ homeserver:
   #
   # media_url: "http://media.repo:8008"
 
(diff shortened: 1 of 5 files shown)


Kegsay:
@lukebarnard1 PTAL

lukebarnard1:
LGTM

PR #315, opened 2016-12-19
Allow multiple in-flight /joined_members requests at once

Kegsay:
Add a new utility class:
 - `QueuePool` : A number of `Queues` all of which can be active at once. New items are put into the first free `Queue` in a FIFO manner.

Use `QueuePool` to make multiple in-flight requests at once. Set to 50 as this seems good enough for Freenode/matrix.org (1m30s for 12k rooms).

### DIFF
--- a/lib/bridge/MemberListSyncer.js
+++ b/lib/bridge/MemberListSyncer.js
@@ -136,10 +136,8 @@ MemberListSyncer.prototype.checkBotPartRoom = Promise.coroutine(function*(ircRoo
 });
 
 // grab all rooms the bot knows about which have at least 1 real user in them.
-// ignoreCache exists because this function hammers /initialSync and that is expeeeensive,
-// so we don't do it unless they need absolutely fresh data. On startup, this can be called
-// multiple times, so we cache the first request's promise and return that instead of making
-// double hits.
+// On startup, this can be called multiple times, so we cache the first request's promise
+// and return that instead of making double hits.
 //
 // returns [
 //   {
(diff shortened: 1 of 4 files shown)


lib/util/QueuePool.js
```diff
@@ -0,0 +1,97 @@
...
+            });
+        }
+        // wait for any queue to become available
+        let promises = this.queues.map((q) => {
+            return q.onceFree();
+        });
+        return Promise.any(promises).then(() => {
+            queue = this._freeQueue();
+            if (!queue) {
+                throw new Error(`QueuePool overflow: starvation. No free queues.`);
```
lukebarnard1: But surely one of them must be free because `onceFree` resolved, so maybe there should be a way to reference the queue that had it's `onceFree` promise resolved? I only say this because to maintain FIFO, the overflow should have higher priority than new items to the pool. WDYT?

spec/unit/QueuePool.spec.js
```diff
@@ -0,0 +1,138 @@
...
+    it("should process overflows FIFO", test.coroutine(function*() {
+        pool.enqueue("a", "a");
+        pool.enqueue("b", "b");
+        pool.enqueue("c", "c");
+        pool.enqueue("d", "d");
+        pool.enqueue("e", "e");
+        yield nextTick();
+        expect(Object.keys(itemToDeferMap).sort()).toEqual(["a", "b", "c"]);
+        pool.enqueue("f", "f");
+        resolveItem("b");
```
lukebarnard1: Ah, but what happens when you resolve an item and _then_ enqueue before the nextTick? I suspicion is that "f" will take "b"s place (see previous comment).

lib/util/QueuePool.js
```diff
@@ -0,0 +1,97 @@
...
+        this.queues = [];
+        for (let i = 0; i < poolSize; i++) {
+            this.queues.push(new Queue(fn));
+        }
+        this.overflow = new Queue(this._overflow.bind(this));
+    }
+
+    // Add an item to the queue. ID and item are passed directly to the Queue.
+    // Index is optional and should be between 0 ~ poolSize-1. It determines
+    // which queue to put the item into.
```
lukebarnard1: I would mention that specifying an index also ignores the overflow, and enqueues an item immediately

lib/bridge/MemberListSyncer.js
```diff
@@ -150,77 +149,74 @@ MemberListSyncer.prototype.checkBotPartRoom = Promise.coroutine(function*(ircRoo
...
-                let joinedRoomIds = yield self.appServiceBot.getJoinedRooms();
-                for (let i = 0; i < joinedRoomIds.length; i++) {
-                    let roomId = joinedRoomIds[i];
-                    try {
-                        let userMap = yield self.appServiceBot.getJoinedMembers(roomId);
-                        let roomInfo = {
-                            id: roomId,
-                            state: [], // JSON objects (Matrix events)
-                            realJoinedUsers: [], // user IDs
-                            remoteJoinedUsers: [], // user IDs
```
Kegsay: Not really: have added 3s delay.

lib/util/Queue.js
```diff
@@ -31,6 +32,34 @@ function Queue(processFn, intervalMs) {
...
+ */
+Queue.prototype.size = function() {
+    return this._queue.length + (this._processing ? 1 : 0);
+};
+
+/**
+ * Return a promise which is resolved when this queue is free (0 items in queue).
+ * @return {Promise}
+ */
+Queue.prototype.onceFree = function() {
```
Kegsay: Usage wise `onceFree` reads better IMO.

lib/util/QueuePool.js
Kegsay: > But surely one of them must be free because onceFree resolved

Yes, I don't ever expect this error to be hit.

> maybe there should be a way to reference the queue that had it's onceFree promise resolved

That would be clearer, I'll do that.

spec/unit/QueuePool.spec.js
Kegsay: Nope. Just tried re-ordering (resolve then enqueue before the next tick) and it works fine. The overflow itself is a `Queue` so won't be prone to race conditions like this.

lib/util/QueuePool.js
Kegsay: Done

lib/util/QueuePool.js
Kegsay: Done

lib/util/Queue.js
```diff
@@ -31,6 +32,34 @@ function Queue(processFn, intervalMs) {
 }
 
 /**
+ * Return the length of the queue.
```
Kegsay: Done

lib/bridge/MemberListSyncer.js
Kegsay: Done

PR #321, opened 2016-12-21
Queue up initial leave requests to Matrix

### LINKED ISSUE #300: Queue initial leaves I->M
This is particularly relevant for when we enable membership list syncing on freenode:
 - Bucket requests by room ID and only have 1 in-flight req per room (Synapse doesn't allow concurrent membership list modifications anyway)
 - Only have N concurrent room queues in progress at once (3?)

Kegsay:
Using an abomination of a `QueuePool` with "per-room" `Queues`, we can
get the desired behaviour of 1 in-flight req per room, max N rooms.

Fixes #300

TODO: Needs tests.


lib/bridge/MemberListSyncer.js
```diff
@@ -319,6 +330,25 @@ MemberListSyncer.prototype.leaveIrcUsersFromRooms = function(rooms, server) {
...
+// {
+//   roomId: "!foo:bar", userIds: [ "@alice:bar", "@bob:bar", ... ]
+// }
+MemberListSyncer.prototype._leaveUsersInRoom = Promise.coroutine(function*(item) {
+    // We need to queue these up in ANOTHER queue so as not to have
+    // 2 in-flight requests at the same time. We return a promise which resolves
+    // when this room is completely done.
+    let self = this;
+    let q = new Queue(Promise.coroutine(function*(userId) {
+        yield self.ircBridge.getAppServiceBridge().getIntent(userId).leave(roomId)
```
lukebarnard1: semi colon missing

lib/bridge/MemberListSyncer.js
```diff
@@ -359,16 +389,17 @@ MemberListSyncer.prototype.updateIrcMemberList = Promise.coroutine(function*(cha
...
-                        this.ircBridge.getAppServiceBridge().getIntent(userId).leave(roomId)
-                    );
-                    stats.membership(true, "part");
-                }
+                return ircUserIds.indexOf(userId) === -1;
             }
         );
+        // ID is the channel as we can't use the room ID: we might have 2 channels mapping
+        // to the same room ID.
+        promises.push(this._leaveQueuePool.enqueue(channel, {
```
lukebarnard1: What happens when one channel has 20 users to leave and get's enqueued followed by the same channel with 100 usersToLeave? Will it not just pass back the promise for the first NAMES, leaving only 20 users?

Or maybe we shouldn't care _too_ much about these weird multi-mapping edge cases?

lib/bridge/MemberListSyncer.js
```diff
@@ -359,16 +389,17 @@ MemberListSyncer.prototype.updateIrcMemberList = Promise.coroutine(function*(cha
...
-        this._memberLists.matrix[roomId].remoteJoinedUsers.forEach(
+        let usersToLeave = this._memberLists.matrix[roomId].remoteJoinedUsers.filter(
             (userId) => {
-                if (ircUserIds.indexOf(userId) === -1) {
-                    promises.push(
-                        this.ircBridge.getAppServiceBridge().getIntent(userId).leave(roomId)
-                    );
-                    stats.membership(true, "part");
-                }
+                return ircUserIds.indexOf(userId) === -1;
```
lukebarnard1: could use `ircUserIds.includes(userId)`

Kegsay: Does not exist in Node v4.

lib/bridge/MemberListSyncer.js
Kegsay: Oh, you're right - I should be doing a tuple of `(channel, room_id)` really.

lib/bridge/MemberListSyncer.js
lukebarnard1: ah, fair

lib/bridge/MemberListSyncer.js
Kegsay: Done

PR #324, opened 2016-12-21
Start adding MemberListSyncer tests

Kegsay:
Based on https://github.com/matrix-org/matrix-appservice-irc/pull/323

Add M->I join test annnnd fix the first bug as a result. Turns out `MemberListSyncer` was not
passing a `displayname` through to `onJoin` so the generated IRC nicks would
incorrectly be based off the user ID! Also removed the `state` mapping as nothing used it.

### DIFF
--- a/spec/integ/MemberListSyncer.spec.js
+++ b/spec/integ/MemberListSyncer.spec.js
@@ -1,0 +1,102 @@
+"use strict";
+let test = require("../util/test");
+let Promise = require("bluebird");
+
+// set up integration testing mocks
+let env = test.mkEnv();
+
+// set up test config
+let config = env.config;
+let roomMapping = {
+    server: config._server,
+    botNick: config._botnick,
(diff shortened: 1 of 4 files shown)


spec/integ/MemberListSyncer.spec.js
```diff
@@ -0,0 +1,99 @@
...
+                            avatar_url: null,
+                        },
+                        "@bob:bar": {
+                            display_name: "Bob",
+                            avatar_url: null,
+                        }
+                    },
+                });
+            }
+            return Promise.reject("unhandled path");
```
lukebarnard1: Should really reject with an Error, I think?

Kegsay: Oh derp, yes!

Kegsay: Done

PR #325, opened 2016-12-21
Add initial leave I->M test and fix bugs as a result

Kegsay:
Isn't testing *fun*?!

On the plus side:

```
------------------------|----------|----------|----------|----------|----------------|
File                    |  % Stmts | % Branch |  % Funcs |  % Lines |Uncovered Lines |
------------------------|----------|----------|----------|----------|----------------|
  MemberListSyncer.js   |    63.16 |    41.67 |       80 |    63.16 |... 420,424,427 |

Previously:
  MemberListSyncer.js   |    24.87 |    15.52 |       15 |    24.87
```

### DIFF
--- a/spec/integ/MemberListSyncer.spec.js
+++ b/spec/integ/MemberListSyncer.spec.js
@@ -43,6 +43,60 @@ describe("MemberListSyncer", function() {
         yield test.afterEach(env);
     }));
 
+    it("should sync initial leaves from IRC to Matrix", test.coroutine(function*() {
+        env.ircMock._autoConnectNetworks(
+            roomMapping.server, "M-alice", roomMapping.server
+        );
+        env.ircMock._whenClient(roomMapping.server, "M-alice", "join", (client, chan, cb) => {
+            expect(chan).toEqual(roomMapping.channel);
+            client._invokeCallback(cb);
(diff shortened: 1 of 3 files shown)


spec/integ/MemberListSyncer.spec.js
```diff
@@ -43,6 +43,60 @@ describe("MemberListSyncer", function() {
...
+            // 'alpha' should leave
+            let alphaClient = env.clientMock._client(ircUserId("alpha"));
+            alphaClient.leave.and.callFake((roomId) => {
+                expect(roomId).toEqual(roomMapping.roomId);
+                resolve();
+                return Promise.resolve({});
+            });
+        });
+
+        yield test.initEnv(env);
```
lukebarnard1: Seems odd to `initEnv` after the test?... but I guess it doesn't matter because we don't yield on `promise` until after this?

Kegsay: `initEnv` starts the bridge, which starts the `MemberListSyncer`, so we need to add hooks in before that happens.

lukebarnard1: yep, ok makes sense

PR #327, opened 2016-12-22
Add support for bridging channel operator modes to power levels

lukebarnard1:


spec/util/client-sdk-mock.js
```diff
@@ -154,3 +155,5 @@ module.exports._client = function(userId) {
 module.exports._reset = function() {
     mockClients = {};
 };
+
+module.exports.clients = () => mockClients;
```
Kegsay: You don't use this, please remove it.

lukebarnard1: oops..

spec/integ/irc-to-matrix.spec.js
```diff
@@ -240,6 +240,87 @@ describe("IRC-to-Matrix message bridging", function() {
...
+        };
+
+        const promise = new Promise((resolve, reject) => {
+            sdk.setPowerLevel.and.callFake(
+            function(roomId, userId, powerLevel, event, callback) {
+                expect(roomId).toBe(roomMapping.roomId);
+                expect(userId).toBe(tRealUserId);
+                expect(powerLevel).toBe(50);
+                resolve();
+            });
```
Kegsay: You're not returning a Promise, so it will error out since `matrix-appservice-bridge` will blindly call `.then()` on the return value.

Kegsay:
Merge conflicts.

lib/bridge/IrcHandler.js
```diff
@@ -538,15 +538,82 @@ IrcHandler.prototype.onMode = Promise.coroutine(function*(req, server, channel,
...
+
+    if (!bridgedClient.unsafeClient) {
+        req.log.info(`Bridged client for ${arg} has no IRC client.`);
+        return;
+    }
+
+    // Bridge operator modes to power levels
+    let modeToPower = server.getModePowerMap();
+
+    if (Object.keys(modeToPower).indexOf(mode) === -1) {
```
Kegsay: I'd prefer it if you did this waaaaay earlier, before we've done a DB lookup. It also makes it clearer what this code is going to handle (power level stuff).

lib/bridge/IrcHandler.js
```diff
@@ -538,15 +538,82 @@ IrcHandler.prototype.onMode = Promise.coroutine(function*(req, server, channel,
...
+        // Not an operator power mode
+        return;
+    }
+
+    const userPrefixes = bridgedClient.unsafeClient.chanData(channel).users[arg];
```
Kegsay: Dangerous. Don't assume there will always be chanData for the given `channel`. Remember that some other random client may have invoked this `onMode`.

lib/bridge/IrcHandler.js
```diff
@@ -538,15 +538,82 @@ IrcHandler.prototype.onMode = Promise.coroutine(function*(req, server, channel,
...
+    console.log(userPrefixes);
```
Kegsay: Remove this.

lib/bridge/IrcHandler.js
```diff
@@ -538,15 +538,82 @@ IrcHandler.prototype.onMode = Promise.coroutine(function*(req, server, channel,
...
+
+            if (modeToPower[m] !== undefined) {
+                userPowers.push(modeToPower[m]);
+            }
+        }
+    );
+
+    let level = 0;
+    if (userPowers.length > 0) {
+        level = userPowers.sort((a, b) => b - a)[0];
```
Kegsay: No comments explaining what you're doing.

lib/config/schema.yml
```diff
@@ -114,6 +114,8 @@ properties:
                                     type: "integer"
                                     minimum: 0
                                     exclusiveMinimum: true
+                        modePowerMap:
+                            type: "object"
```
Kegsay: You can document more than just the fact it's an object.

lib/bridge/IrcHandler.js
```diff
@@ -538,15 +538,82 @@ IrcHandler.prototype.onMode = Promise.coroutine(function*(req, server, channel,
...
-        return; // ignore everything but privateModes
-    }
     req.log.info(
         "onMode(%s) in %s by %s (arg=%s)",
         (enabled ? ("+" + mode) : ("-" + mode)),
         channel, by, arg
     );
 
+    if (privateModes.indexOf(mode) !== -1) {
+        req.log.info("Private mode");
```
Kegsay: This isn't helpful. We can already infer that we will branch here because we log the `mode`  just above on `:541`.

lib/bridge/IrcHandler.js
```diff
@@ -538,15 +538,82 @@ IrcHandler.prototype.onMode = Promise.coroutine(function*(req, server, channel,
...
+
+    const matrixRooms = yield this.ircBridge.getStore().getMatrixRoomsForChannel(server, channel);
+
+    if (matrixRooms.length === 0) {
+        req.log.info("No mapped matrix rooms for IRC channel %s", channel);
+        return;
+    }
+
+    const bridgedClient = this.ircBridge.getClientPool().getBridgedClientByNick(
+        server, arg
```
Kegsay: `arg` is an unfortunate variable name. If know it's a nick, please alias it.

spec/integ/irc-to-matrix.spec.js
```diff
@@ -266,78 +258,268 @@ describe("IRC-to-Matrix name bridging", function() {
...
+                user_id: tRealUserId,
+                room_id: roomMapping.roomId,
+                type: "m.room.message"
+            });
+        });
     }));
 
     afterEach(test.coroutine(function*() {
-        yield test.afterEach(this, env); // eslint-disable-line no-invalid-this
+        yield test.afterEach(this, env) // eslint-disable-line no-invalid-this
```
Kegsay: This is probably why you're merge conflicting.

spec/integ/irc-to-matrix.spec.js
```diff
@@ -266,78 +258,268 @@ describe("IRC-to-Matrix name bridging", function() {
...
+                );
+            });
         });
-    });
+
+        yield promise;
+    }));
 });
+
+// describe("IRC-to-Matrix name bridging", function() {
```
Kegsay: Uhhhh...

lib/config/schema.yml
lukebarnard1: I was unable to find the relevant jsdoc docs for this, but I shall try again

lukebarnard1: This is done FTR.

lib/bridge/IrcHandler.js
```diff
@@ -538,15 +538,90 @@ IrcHandler.prototype.onMode = Promise.coroutine(function*(req, server, channel,
...
+            if (modeToPower[m] !== undefined) {
+                userPowers.push(modeToPower[m]);
+            }
+        }
+    );
+
+    // Sort the userPowers for this user in descending order
+    // and grab the highest value at the start of the array.
+    let level = 0;
+    if (userPowers.length > 0) {
```
Kegsay: If `userPowers` is empty, we will give this user a power level of `0`. What we probably mean instead is to default to `users_default` in the `m.room.power_levels` event which is not the same thing.

spec/integ/irc-to-matrix.spec.js
```diff
@@ -240,6 +240,188 @@ describe("IRC-to-Matrix message bridging", function() {
...
+                },
+                user_id: tRealUserId,
+                room_id: roomMapping.roomId,
+                type: "m.room.message"
+            });
+        });
+    }));
+
+    afterEach(test.coroutine(function*() {
+        yield test.afterEach(this, env); // eslint-disable-line no-invalid-this
```
Kegsay: Check again. There is no `this` anymore.

lib/bridge/IrcHandler.js
```diff
@@ -538,15 +538,90 @@ IrcHandler.prototype.onMode = Promise.coroutine(function*(req, server, channel,
...
+
+    req.log.info(
+        `Mode ${mode} received for ${nick}, prefixes now = ${userPrefixes} ` +
+        `and powers are now = ${userPowers}, bridging power level of ${userId} ` +
+        `at ${level}.`
+    );
+
+    const promises = matrixRooms.map((room) => {
+        return this.ircBridge.getAppServiceBridge().getIntent()
+            .setPowerLevel(room.getId(), userId, level);
```
Kegsay: You're not giving the JS SDK enough information. This will *clobber* the previous power level event, but you just want to alter it. Sadly there is no PATCH or equiv for power levels in the CS API. You need to supply the entire `event`. http://matrix-org.github.io/matrix-js-sdk/0.7.2/client.js.html#line775

lukebarnard1: Ah, but this isn't the JS SDK , it's the bridge library, which does the GET for us.

lib/bridge/IrcHandler.js
lukebarnard1: I _think_ that would be a simple case of passing `undefined` instead as the desired power level. By unsetting the power level, that should be regarded as the  `users_default`.

lib/bridge/IrcHandler.js
Kegsay: Magic! Forgot it did that, carry on then.

lib/bridge/IrcHandler.js
Kegsay: Yup, I concur - setting `undefined` should do it.

Kegsay:
LGTM

PR #333, opened 2017-01-07
Allow IPv6 connections and add option to require IPv6.

OlegGirko:
This change allows IPv6 connections to IRC servers when there is no `prefix` option in `ipv6` section.
Also, it adds `only` boolean option to `ipv6` section that forces using IPv6 only for connection to server.

### DIFF
--- a/lib/irc/ConnectionInstance.js
+++ b/lib/irc/ConnectionInstance.js
@@ -309,7 +309,7 @@ ConnectionInstance.create = function(server, opts, onCreatedCallback) {
         port: server.getPort(),
         selfSigned: server.useSslSelfSigned(),
         retryCount: 0,
-        family: server.getIpv6Prefix() ? 6 : 4,
+        family: server.getIpv6Prefix() || server.getIpv6Only() ? 6 : null,
         bustRfc3484: true,
     };
 
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -264,15 +264,17 @@ ircService:
         # disconnected and replaced.
         # Optional. Default: 30.
         maxClients: 30
-        #
-        # Optional. The IPv6 prefix to use for generating unique addresses for each
-        # connected user. If not specified, all users will connect from the same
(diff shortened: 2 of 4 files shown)


lib/irc/ConnectionInstance.js
```diff
@@ -309,7 +309,7 @@ ConnectionInstance.create = function(server, opts, onCreatedCallback) {
         port: server.getPort(),
         selfSigned: server.useSslSelfSigned(),
         retryCount: 0,
-        family: server.getIpv6Prefix() ? 6 : 4,
+        family: server.getIpv6Prefix() || server.getIpv6Only() ? 6 : null,
```
Kegsay: Please keep this as `4` in the false branch.

Kegsay:
- Missing documentation in `config.sample.yaml`.
- Missing default value in https://github.com/matrix-org/matrix-appservice-irc/blob/master/lib/irc/IrcServer.js#L464 so `getIpv6Only` can return `undefined` rather than what was intended: `false`.

Otherwise, LGTM thank you!

lib/irc/ConnectionInstance.js
OlegGirko: No, this breaks the whole purpose of this patch. This *has* to be `null` to not force to use IPv4.

Intended behaviour is:
* if there is `prefix` option present in `ipv6` section or `only` option set to `true`, force IPv6 connection;
* otherwise, use IPv6 if available, fall back to IPv4 if not, but *do not* force IPv4.

OlegGirko:
I've pushed revised change.

* Documentation added to `config.sample.yaml`.
* Default for `only` option added to default configuration.
* Change rebased on updated `develop` branch.

lib/irc/ConnectionInstance.js
Kegsay: Right, I understand now.

Kegsay:
LGTM, thank you!

Issue #258, opened 2016-10-26
What should uploaded content look like on IRC channels?


Kegsay:
When a Matrix user uploads an image/file/etc to a room, the IRC bridge needs to convert that into text to send into the IRC channel. What should this text look like?

Related: When a Matrix user sends a block of text into a room, the IRC bridge "pastebins" the content by uploading it as a `.txt` file, which then is sent into the IRC channel. What should this text look like?

There seems to be 2 camps of people who want just the raw link vs people who want a wordier "so-and-so uploaded a thing", and I am using this issue to build consensus.

See:
- https://github.com/matrix-org/matrix-appservice-irc/issues/239
- https://github.com/matrix-org/matrix-appservice-irc/issues/142

Mikaela:
I prefer there being text too as random URL without context can look shady and leave unclicked.

jansol:
No IRC user I know sends anything other than the URL when pastebinning stuff. And for convenience reasons the entire message should fit on a single line in a standard 80-column terminal.

Of course here the pastebinning is not always intended so some explanation might be warranted. How "smart" do we want to be? Pastebinning triple-backtick-quoted text rarely needs an explanation, so triple backticks could just be replaced with URLs in most cases.

> random URL without context can look shady

Well, it's a URL on your HS...

Mikaela:
Generally pastebins are easy to recognise as pastebins based on the URL (doesn't apply to https://matrix.org/...) and while I know the URL belongs to my homeserver, does random IRC user know what is homeserver or even Matrix other than movie trilogy?

jansol:
I hope your random IRC user at least knows where they are telling their client to connect. Especially if they are the paranoid sort... It shouldn't be difficult to recognize the URL based on that.

Another problem with adding unnecessary text is that it'll mess up non-english channels.

ThatGeoGuy:
I am personally fine with just a link, but maybe there's a third option? I know a lot of channels have bots that lookup links that are pasted and provide some information about them. In one channel I'm on in Rizon, the channel bot will speak and provide some context about links posted. e.g. for a Youtube video:

```
ThatGeoGuy+> https://youtu.be/some-youtube-link
Ultrabot&> [URI 1191 by "ThatGeoGuy"] Title: My Awesome Video - Youtube
ThatGeoGuy+> https://thatgeoguy.ca/some-image.png
[...]

PR #256, opened 2016-10-26
Restructure uploaded file bridging IRC messages to IRC actions

lukebarnard1:
Instead of what we had previously (see #239), restructure messages to indicate the uploading of files to make them clearer to users.

Now looks like:
`* lukeb[m] uploaded an image: Slack App Icon.png (612KB) - 
http://localhost:8080/_matrix/media/v1/download/localhost:8480/kStIBZcfVoxiMjdiQiyTOTfz`

(Commit message unintentionally useless, sorry)

PR #336, opened 2017-01-09
Current separator causes confusion for some users

tjgillies:
The current separator causes confusion when people are mentally parsing the link. Creating PR to make an obvious distinction that this is not part of the URL. Someone suggested using "GNU Parallel" syntax.


Kegsay:
Sorry, but we already have an open PR for the syntax of uploaded files. Please see https://github.com/matrix-org/matrix-appservice-irc/pull/256 

There is a lot of contention over what the precise syntax should be: please comment on https://github.com/matrix-org/matrix-appservice-irc/issues/258 if you care about this.

PR #337, opened 2017-01-09
Coroutinise ConnectionInstance.connect

Kegsay:
Mainly to improve readability.

Also set `no-constant-condition` to not fail linting, because there are valid cases where `while(true)` loops are appropriate.

### DIFF
--- a/lib/irc/ConnectionInstance.js
+++ b/lib/irc/ConnectionInstance.js
@@ -292,12 +292,12 @@ ConnectionInstance.prototype._resetPingSendTimer = function() {
  * @param {Function} onCreatedCallback Called with the client when created.
  * @return {Promise} Resolves to an ConnectionInstance or rejects.
  */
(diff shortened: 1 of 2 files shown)


lib/irc/ConnectionInstance.js
```diff
@@ -351,49 +362,12 @@ ConnectionInstance.create = function(server, opts, onCreatedCallback) {
...
-
-        // Initially, pretend to have failed in order to start things off
-        retryFailed({});
+        }
     }
-    retryForever();
-
-    return d.promise;
-};
+    throw new Error("ConnectionInstance.create: failed to connect");
```
lukebarnard1: I am not convinced this code will ever be executed.

Kegsay: Agreed. The intention is mainly for it to throw up in our faces if it ever did occur. I'll make the error here clearer.

Kegsay: Done.

PR #338, opened 2017-01-09
Don't log stack traces when disconnect() times out

Kegsay:
It's misleading as it makes it look like a Bad Thing when really it isn't.
This logging was actually being done by bluebird, so we now set a rejection
handler which just logs the message.


lukebarnard1:
Would it not be simpler to just not reject when there is a time out and just log that it's timed-out?

Kegsay:
Eh, could do. It feels wrong to resolve the promise though, and we can't leave it unfulfilled.

lukebarnard1:
Ah I see your point. In that case I'd prefer rejection. LGTM :+1:

PR #340, opened 2017-01-09
Add script to kick idle bridged users from a room

lukebarnard1:
$ python remove-idle-users.py
 --token=someaccesstoken23478586496849569
--alias='#somealias:some.domain' 
--homeserver=http://some.domain 
--domain=some.domain 
--since=50 
--template=@irc_$NICK
```
--token : the access token for the AS
--alias : the alias of the room to kick from
--homeserver : the URL of the HS
--domain : the domain part of the HS URL
--template : the userPattern template (in config.sample.yaml, matrixClients)
```

### DIFF
--- a/scripts/remove-idle-users.py
+++ b/scripts/remove-idle-users.py
@@ -47,15 +47,15 @@ def kick_idlers(homeserver, room_id, token, since, user_prefix):
             })
         )
         if res.status_code >= 400:
-            failure = { "user_id": user_id }
+            failure = { "user_id" : user_id }
             try:
                 failure["response_json"] = res.json()
             except Exception as e:
                 print("Could not get JSON body from failure response: %s" % e)
-            failure_responses.append()
+            failure_responses.append(failure)
         else:
             count += 1
-    print("Kicked %s/%s users in total (%s failed requests)" % (count, len(user_ids), len(failure_responses)))
+    print("Kicked %s/%s users in total (%s failed requests)" % (count, count + len(failure_responses), len(failure_responses)))
 
     if len(failure_responses) == 0:
         return
--- a/scripts/remove-idle-users.py
+++ b/scripts/remove-idle-users.py
@@ -90,7 +90,6 @@ def escapeRegExp(s):
 
 def main(token, alias, homeserver, homeserver_domain, since, user_template):
     print("Removing idle users in %s" % alias)
-    token = token
     room_id = get_room_id(homeserver, alias, token)
     if not room_id:
         raise Exception("Cannot resolve room alias to room_id")
@@ -99,11 +98,11 @@ def main(token, alias, homeserver, homeserver_domain, since, user_template):
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser("Remove idle users from a given Matrix room")
-    parser.add_argument("-t", "--token", help="The AS token", required=True)
+    parser.add_argument("-t", "--token", help="The access token", required=True)
     parser.add_argument("-a", "--alias", help="The alias of the room eg '#freenode_#matrix-dev:matrix.org'", required=True)
     parser.add_argument("-u", "--homeserver", help="Base homeserver URL eg 'https://matrix.org'", required=True)
-    parser.add_argument("-d", "--domain", help=" matrix.org'", required=True)
-    parser.add_argument("-s", "--since", type=int, help="Since idle users have been offline for", required=True)
+    parser.add_argument("-d", "--domain", help="Domain of the homserver eg 'matrix.org'", required=True)
+    parser.add_argument("-s", "--since", type=int, help="Days since idle users have been offline for eg '30'", required=True)
     parser.add_argument("-e", "--template", help="User template to determine whether a user should be kicked. E.g. @$SERVER_$NICK", required=True)
     args = parser.parse_args()
     if not args.token or not args.alias or not args.homeserver:
--- a/scripts/remove-idle-users.py
+++ b/scripts/remove-idle-users.py
@@ -5,11 +5,6 @@ import sys
 import json
 import urllib
 import requests
-import re
-
-## debug request
-import httplib as http_client
-http_client.HTTPConnection.debuglevel = 1
 
 def get_room_id(homeserver, alias, token):
     res = requests.get(homeserver + "/_matrix/client/r0/directory/room/" + urllib.quote(alias) + "?access_token=" + token)
@@ -24,22 +19,23 @@ def is_idle(homeserver, user_id, token, activity_threshold_ms):
     return get_last_active_ago(homeserver, user_id, token) > activity_threshold_ms
 
 def get_idle_users(homeserver, room_id, token, since):
-    res = requests.get(homeserver + "/_matrix/client/r0/rooms/" + urllib.quote(room_id) + "/members?access_token=" + token)
-    user_ids = [event["state_key"] for event in res.json().get("chunk", None)]
+    res = requests.get(homeserver + "/_matrix/client/r0/rooms/" + urllib.quote(room_id) + "/joined_members?access_token=" + token)
+    user_ids = [user_id for user_id in res.json().get("joined", None)]
 
     activity_threshold_ms = since * 24 * 60 * 60 * 1000
 
     return [user_id for user_id in user_ids if is_idle(homeserver, user_id, token, activity_threshold_ms)]
 
-def kick_idlers(homeserver, homeserver_domain, room_id, token, since, user_template=None):
+def kick_idlers(homeserver, homeserver_domain, room_id, token, since, user_prefix):
     reason = "Being idle for >%s days" % since
 
     user_ids = get_idle_users(homeserver, room_id, token, since)
     failure_responses = []
-    print("Kicking %s idle users from %s" % (len(user_ids), room_id))
+    count = 0
+    print("There are %s idle users in %s" % (len(user_ids), room_id))
     for user_id in user_ids:
-        # Ignore unclaimed users, if user_template is specified
-        if user_template and not claims_user_id(user_id, user_template, homeserver_domain):
+        # Ignore users that do not start with the user_prefix
+        if not user_id.startswith(user_prefix):
             continue
         res = requests.post(
             homeserver + "/_matrix/client/r0/rooms/" + urllib.quote(room_id) + "/kick?access_token=" + token,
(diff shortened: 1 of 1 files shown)


scripts/remove-idle-users.py
```diff
@@ -0,0 +1,112 @@
...
+    token = token
+    room_id = get_room_id(homeserver, alias, token)
+    if not room_id:
+        raise Exception("Cannot resolve room alias to room_id")
+
+    kick_idlers(homeserver, homeserver_domain, room_id, token, since, user_template)
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser("Remove idle users from a given Matrix room")
+    parser.add_argument("-t", "--token", help="The AS token", required=True)
```
Kegsay: Technically any token, not just AS.

scripts/remove-idle-users.py
```diff
@@ -0,0 +1,112 @@
...
+    parser.add_argument("-a", "--alias", help="The alias of the room eg '#freenode_#matrix-dev:matrix.org'", required=True)
+    parser.add_argument("-u", "--homeserver", help="Base homeserver URL eg 'https://matrix.org'", required=True)
+    parser.add_argument("-d", "--domain", help=" matrix.org'", required=True)
```
Kegsay: `'` mismatch. Also missing the sentence!

scripts/remove-idle-users.py
```diff
@@ -0,0 +1,112 @@
...
+    parser.add_argument("-s", "--since", type=int, help="Since idle users have been offline for", required=True)
```
Kegsay: Missing units and example.

scripts/remove-idle-users.py
```diff
@@ -0,0 +1,112 @@
...
+
+    return regex + suffix
+
+def escapeRegExp(s):
+    return re.escape(s);
+
+
+def main(token, alias, homeserver, homeserver_domain, since, user_template):
+    print("Removing idle users in %s" % alias)
+    token = token
```
Kegsay: Uhh, nop?

scripts/remove-idle-users.py
```diff
@@ -0,0 +1,111 @@
...
+import requests
+import re
+
+## debug request
+import httplib as http_client
+http_client.HTTPConnection.debuglevel = 1
+
+def get_room_id(homeserver, alias, token):
+    res = requests.get(homeserver + "/_matrix/client/r0/directory/room/" + urllib.quote(alias) + "?access_token=" + token).json()
+    return res.get("room_id", None)
```
Kegsay: Nit: Rather than return a nullable `room_id`, I would just do:
```py
res = requests.get(homeserver + "/_matrix/client/r0/directory/room/" + urllib.quote(alias) + "?access_token=" + token)
res.raise_for_status()
return res.json()["room_id"]
```

This will throw on non-2xx and if `room_id` doesn't exist, both of which you want to do. You can then remove the null guard on `:94`.

scripts/remove-idle-users.py
```diff
@@ -0,0 +1,111 @@
...
+
+def get_last_active_ago(homeserver, user_id, token):
+    res = requests.get(homeserver + "/_matrix/client/r0/presence/" + urllib.quote(user_id) + "/status?access_token=" + token).json()
+    return res.get("last_active_ago", None)
+
+def is_idle(homeserver, user_id, token, activity_threshold_ms):
+    return get_last_active_ago(homeserver, user_id, token) > activity_threshold_ms
+
+def get_idle_users(homeserver, room_id, token, since):
+    res = requests.get(homeserver + "/_matrix/client/r0/rooms/" + urllib.quote(room_id) + "/members?access_token=" + token)
```
Kegsay: Don't use this API, it'll take far too long. Instead, use `/joined_members` - https://github.com/matrix-org/gomatrix/blob/master/client.go#L546

scripts/remove-idle-users.py
```diff
@@ -0,0 +1,111 @@
...
+        if user_template and not claims_user_id(user_id, user_template, homeserver_domain):
+            continue
+        res = requests.post(
+            homeserver + "/_matrix/client/r0/rooms/" + urllib.quote(room_id) + "/kick?access_token=" + token,
+            data = json.dumps({
+                "reason": reason,
+                "user_id": user_id
+            })
+        )
+        res.raise_for_status()
```
Kegsay: Don't raise for status here. It'll be annoying as all hell if it spends the time working out who to kick then dies on the first failure. I would add the failures to a list and then whine at the end saying "I couldn't kick these guys because $RESPONSE_JSON".

scripts/remove-idle-users.py
```diff
@@ -0,0 +1,111 @@
...
+    kick_idlers(homeserver, homeserver_domain, room_id, token, since, user_template)
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser("Remove idle users from a given Matrix room")
+    parser.add_argument("-t", "--token", help="The access token", required=True)
+    parser.add_argument("-a", "--alias", help="The alias of the room eg '#freenode_#matrix-dev:matrix.org'", required=True)
+    parser.add_argument("-u", "--homeserver", help="Base homeserver URL eg 'https://matrix.org'", required=True)
+    parser.add_argument("-d", "--domain", help="Domain of the homserver eg 'matrix.org'", required=True)
+    parser.add_argument("-s", "--since", type=int, help="Days since idle users have been offline for eg '30'", required=True)
+    parser.add_argument("-e", "--template", help="User template to determine whether a user should be kicked. E.g. @$SERVER_$NICK", required=True)
```
Kegsay: I would just do prefix matching to save yourself the hassle from trying to work out what to correctly parse things as. It also makes it clearer to the end-user what they should put as a template value ("@freenode_" for example). Bear in mind people who run this may not know the IRC bridge config file syntax so `$SERVER` and `$NICK` may be foreign to them.

scripts/remove-idle-users.py
lukebarnard1: This should really be documented... would have saved me some time

scripts/remove-idle-users.py
lukebarnard1: wfm

lukebarnard1: But then you would get a quite useless error when the `room_id` doesn't get sent back? If this does indicate that the server couldn't resolve it it, would it not be better to throw an error that suggests as such?

scripts/remove-idle-users.py
lukebarnard1: Oh fair point. Matching by static prefix string makes sense.

scripts/remove-idle-users.py
lukebarnard1: Ah but then you don't get to see the error. Fair point, I'll have it raise for status.

scripts/remove-idle-users.py
Kegsay: It's a new API Erik added not long ago, hence missing docs.

scripts/remove-idle-users.py
```diff
@@ -0,0 +1,79 @@
...
+def kick_idlers(homeserver, homeserver_domain, room_id, token, since, user_prefix):
+    reason = "Being idle for >%s days" % since
+
+    user_ids = get_idle_users(homeserver, room_id, token, since)
+    failure_responses = []
+    count = 0
+    print("There are %s idle users in %s" % (len(user_ids), room_id))
+    for user_id in user_ids:
+        # Ignore users that do not start with the user_prefix
+        if not user_id.startswith(user_prefix):
```
Kegsay: We surely want the opposite here. I want to ignore users that *start with* `@freenode_` because they are managed by the bridge.

scripts/remove-idle-users.py
```diff
@@ -0,0 +1,79 @@
...
+            continue
+        res = requests.post(
+            homeserver + "/_matrix/client/r0/rooms/" + urllib.quote(room_id) + "/kick?access_token=" + token,
+            data = json.dumps({
+                "reason": reason,
```
Kegsay: If only it were that easy. Whilst the [spec](http://matrix.org/docs/spec/client_server/r0.2.0.html#post-matrix-client-r0-rooms-roomid-kick) states that `reason` is a valid key, it looks like [synapse](https://github.com/matrix-org/synapse/blob/master/synapse/rest/client/v1/room.py#L564) doesn't abide by this.

Use [this API](http://matrix.org/docs/spec/client_server/r0.2.0.html#put-matrix-client-r0-rooms-roomid-state-eventtype-statekey) instead. Will file an issue with Synapse peeps.

Kegsay: https://github.com/matrix-org/synapse/issues/1788

scripts/remove-idle-users.py
```diff
@@ -0,0 +1,79 @@
...
+                "user_id": user_id
+            })
+        )
+        if res.status_code >= 400:
+            failure_responses.append({ "user_id": user_id, "response_json": res.json()})
```
Kegsay: No guarantee that the server returned JSON in this case. I'm guessing `requests` will throw if you call `.json()` in that case? No idea. We should handle it though (I'm thinking 504 Gateway Timeouts from the Apache).

scripts/remove-idle-users.py
```diff
@@ -0,0 +1,79 @@
...
+        else:
+            count += 1
+    print("Kicked %s users in total" % count)
```
Kegsay: More stats in general would be super useful for me:
 - How many kicks went through (`count`)
 - How many kicks were attempted (`count` + `len(failure_responses)`)
 - How many kicks failed (`len(failure_responses)`)

Probably tied together as:
```
Kicked 44/55 users in total
```

scripts/remove-idle-users.py
```diff
@@ -0,0 +1,79 @@
...
+
+    if len(failure_responses) == 0:
+        return
+    print("Could not kick the following users:")
+    for failure in failure_responses:
+        print("%s : %s - %s" % (failure["user_id"], failure["response_json"]["error"], failure["response_json"]["errcode"]))
```
Kegsay: No guarantees that `error` and `errcode` even exist. It would be a shame to die when logging the failed requests. I would just log the `dict`.

scripts/remove-idle-users.py
```diff
@@ -0,0 +1,79 @@
...
+    kick_idlers(homeserver, homeserver_domain, room_id, token, since, user_prefix)
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser("Remove idle users from a given Matrix room")
+    parser.add_argument("-t", "--token", help="The access token", required=True)
+    parser.add_argument("-a", "--alias", help="The alias of the room eg '#freenode_#matrix-dev:matrix.org'", required=True)
+    parser.add_argument("-u", "--homeserver", help="Base homeserver URL eg 'https://matrix.org'", required=True)
+    parser.add_argument("-d", "--domain", help="Domain of the homserver eg 'matrix.org'", required=True)
+    parser.add_argument("-s", "--since", type=int, help="Days since idle users have been offline for eg '30'", required=True)
+    parser.add_argument("-p", "--prefix", help="User prefix to determine whether a user should be kicked. E.g. @freenode_", required=True)
```
Kegsay: Technically the prefix isn't *required*?

scripts/remove-idle-users.py
```diff
@@ -0,0 +1,79 @@
...
+        raise Exception("Cannot resolve room alias to room_id")
+
+    kick_idlers(homeserver, homeserver_domain, room_id, token, since, user_prefix)
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser("Remove idle users from a given Matrix room")
+    parser.add_argument("-t", "--token", help="The access token", required=True)
+    parser.add_argument("-a", "--alias", help="The alias of the room eg '#freenode_#matrix-dev:matrix.org'", required=True)
+    parser.add_argument("-u", "--homeserver", help="Base homeserver URL eg 'https://matrix.org'", required=True)
+    parser.add_argument("-d", "--domain", help="Domain of the homserver eg 'matrix.org'", required=True)
```
Kegsay: Why is this A Thing? You never use it.

scripts/remove-idle-users.py
lukebarnard1: The `continue` on the line after makes sure they are skipped

scripts/remove-idle-users.py
lukebarnard1: It's possible that the bridge does have enough power to kick users that it does not have claim to, so I suppose it could be optional. Although I'm not really sure why an AS token would be used to kick real users.

scripts/remove-idle-users.py
lukebarnard1: ah, 'twas left over from the user template thing

scripts/remove-idle-users.py
```diff
@@ -37,43 +37,49 @@ def kick_idlers(homeserver, homeserver_domain, room_id, token, since, user_prefi
...
             })
         )
         if res.status_code >= 400:
-            failure_responses.append({ "user_id": user_id, "response_json": res.json()})
+            failure = { "user_id": user_id }
+            try:
+                failure["response_json"] = res.json()
+            except Exception as e:
+                print("Could not get JSON body from failure response: %s" % e)
+            failure_responses.append()
```
Kegsay: Appending nothing?

scripts/remove-idle-users.py
```diff
@@ -37,43 +37,49 @@ def kick_idlers(homeserver, homeserver_domain, room_id, token, since, user_prefi
...
         else:
             count += 1
-    print("Kicked %s users in total" % count)
+    print("Kicked %s/%s users in total (%s failed requests)" % (count, len(user_ids), len(failure_responses)))
```
Kegsay: My earlier comment was explicit for a reason: `len(user_ids)` will include ones you want to filter out according to `prefix`. It's probably fine since you're giving all the information required to work out the values, but it'll be confusing to see things like:

> Kicked 5/10 users in total (2 failed requests)

and wonder what happened to the remaining 3.

lukebarnard1: oh, good point

Kegsay:
LGTM

PR #346, opened 2017-01-16
Extend !whois and document !nick without args

Kegsay:
- `!nick` without args shows your current nick. Document it in `!help`.
- `!whois` has now been extended to allow Matrix user IDs to be used. This
  then returns information about that user's IRC connection (nick/server).
  The presence of a leading `@` determines the difference between querying
  on IRC vs Matrix because IRC nicks cannot start with a leading `@` and
  Matrix user IDs all start with a leading `@`.

Fixes https://github.com/vector-im/riot-web/issues/2939

### DIFF
--- a/lib/bridge/MatrixHandler.js
+++ b/lib/bridge/MatrixHandler.js
@@ -208,11 +208,13 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
             },
             "!nick": {
                 example: `!nick irc.example.com DesiredNick`,
-                summary: `Change your nick`,
+                summary: "Change your nick. If no arguments are supplied, " +
+                         "your current nick is shown.",
             },
             "!whois": {
-                example: `!whois nick`,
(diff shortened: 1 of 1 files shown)


lib/bridge/MatrixHandler.js
```diff
@@ -208,11 +208,13 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
...
             "!nick": {
                 example: `!nick irc.example.com DesiredNick`,
-                summary: `Change your nick`,
+                summary: "Change your nick. If no arguments are supplied, " +
+                         "your current nick is shown.",
             },
             "!whois": {
-                example: `!whois nick`,
-                summary: `Do a /whois lookup`,
+                example: `!whois nick|mxid`,
```
lukebarnard1: would be nice to see `@alice:matrix.org` as an example mxid

Kegsay: Done.

PR #365, opened 2017-02-22
Query for channel modes AFTER room mappings are created

### LINKED ISSUE #362: Dynamic mapping to a +i channel does not create invite-only room
Hi,

I tried to join a room corresponding to a channel with the +i mode, and the created room is accessible to "Anyone who knows the room's link, apart from guests".

Reproduction:
- on the IRC network, join a new channel
- set mode +i
- join it from matrix

Version: 0.7.1 (a2b025683176924fcc5abb8dbb05b75572fc6b33)
Network: irc.iiens.net (UnrealIRCd)

Here is the `debug.log` file when trying with a channel name `#ohlol`: [ohlol.txt](https://github.com/matrix-org/matrix-appservice-irc/files/791112/ohlol.txt).

Kegsay:
Fixes #362

### DIFF
--- a/lib/bridge/MatrixHandler.js
+++ b/lib/bridge/MatrixHandler.js
@@ -427,10 +427,10 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
             );
             // /mode the channel AFTER we have created the mapping so we process
             // +s and +i correctly.
-            this.ircBridge.publicitySyncer.initModeForChannel(ircServer, ircRoom.channel).catch(
+            this.ircBridge.publicitySyncer.initModeForChannel(ircServer, ircChannel).catch(
             (err) => {
                 log.error(
-                    `Could not init mode for channel ${ircRoom.channel} on ${ircServer.domain}`
+                    `Could not init mode for channel ${ircChannel} on ${ircServer.domain}`
                 );
             });
             req.log.info(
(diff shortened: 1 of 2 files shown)


lib/bridge/MatrixHandler.js
```diff
@@ -425,6 +425,14 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
             yield this.ircBridge.getStore().storeRoom(
                 ircRoom, mxRoom, 'join'
             );
+            // /mode the channel AFTER we have created the mapping so we process
+            // +s and +i correctly.
+            this.ircBridge.publicitySyncer.initModeForChannel(ircServer, ircRoom.channel).catch(
```
lukebarnard1: could use `ircChannel` instead of `ircRoom.channel`

lib/bridge/MatrixHandler.js
```diff
@@ -425,6 +425,14 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
...
+            (err) => {
+                log.error(
+                    `Could not init mode for channel ${ircRoom.channel} on ${ircServer.domain}`
```
lukebarnard1: as above

lib/bridge/MatrixHandler.js
Kegsay: Done.

lib/bridge/MatrixHandler.js
Kegsay: Done.

PR #367, opened 2017-02-26
Forward invite events from IRC to Matrix

erdnaxeli:
The IrcHandler now forward invite events from IRC to Matrix users. The room is created if needed.

I am just not sure about how to handle the case where many rooms are bridged with the channel. Now the code invite the user on every room bridged but it seems a little overkill. I think we should invite the user to only one room, but which one do we choose? Or maybe invite the user to all the rooms and let him choose is good enough?


Kegsay:
> how to handle the case where many rooms are bridged with the channel

Typically we fork "1 -> Many" as you are doing (so invite to all mapped rooms).

Excellent PR, thank you! LGTM!

PR #374, opened 2017-03-01
Set a sensible HTTP(S) socket limit

### LINKED ISSUE #345: Set max connections on HTTP request pooling
Freenode bridge died just now with:

```
Error: EMFILE: too many open files, uv_resident_set_memory
```

Caused by a very slow Synapse when making many HTTP requests. This led to a runaway fd count on our already stretched process. We should be pooling HTTP connections, but we don't appear to be.

Kegsay:
Manually tested on a `localhost` Synapse by artificially `sleep(5)`ing calls to
`/joined_members` with a limit of 1 and checking it did requests sequentially.

Fixes #345.

### DIFF
--- a/lib/main.js
+++ b/lib/main.js
@@ -14,6 +14,11 @@ var ident = require("./irc/ident");
 var logging = require("./logging");
 var log = logging.get("main");
 
+// Allow up to 1000 HTTP(S) sockets. Default is Infinity, which means that a slow HS coupled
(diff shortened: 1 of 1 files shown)


lukebarnard1:
LGTM but ooi, what happens when the sockets run out...? Does it error when trying to create one?

Kegsay:
It's just a pool, so it'll block until a socket is free.

PR #375, opened 2017-03-01
Clear timeouts on disconnect to allow faster GC cycles

Kegsay:
Otherwise we'll be holding onto references to clients for up to the idle timeout time, which can be days!


Kegsay:
> so long as we don't need to do _keepAlive in BridgedClient if the client is disconnected.

Can confirm, we don't need to.

Issue #358, opened 2017-02-09
metrics: Don't treat dropped Matrix messages as requests


Kegsay:
https://github.com/matrix-org/matrix-appservice-irc/blob/master/lib/bridge/IrcBridge.js#L483

We should be returning an error code which then will not treat that request as a metrics-able request (like we do for echoed messages: IGNORE_VIRTUAL_USER).

PR #393, opened 2017-03-15
Don't treat dropped messages as success

Kegsay:
Fixes https://github.com/matrix-org/matrix-appservice-irc/issues/358

### DIFF
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -342,6 +342,19 @@ IrcBridge.prototype.run = Promise.coroutine(function*(port) {
     this._startedUp = true;
 });
 
+IrcBridge.prototype._logMetric = function(req, outcome) {
+    if (!this._timers) {
+        return; // metrics are disabled
+    }
+    const isFromIrc = Boolean((req.getData() || {}).isFromIrc);
+    const timer = this._timers[
+        isFromIrc ? "remote_request_seconds" : "matrix_request_seconds"
+    ];
+    if (timer) {
+        timer.observe({outcome: outcome}, req.getDuration() / 1000);
+    }
+}
+
 IrcBridge.prototype._addRequestCallbacks = function() {
     function logMessage(req, msg) {
         const data = req.getData();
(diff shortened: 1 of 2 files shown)


lib/bridge/IrcBridge.js
```diff
@@ -360,11 +361,20 @@ IrcBridge.prototype._addRequestCallbacks = function() {
...
+                timer.observe({outcome: "dropped"}, req.getDuration() / 1000);
+            }
+            return;
+        }
         logMessage(req, "SUCCESS");
-        var isFromIrc = Boolean((req.getData() || {}).isFromIrc);
         stats.request(isFromIrc, "success", req.getDuration());
         if (this._timers) {
-            var timer = this._timers[
+            let timer = this._timers[
```
lukebarnard1: `const` I guess

lib/bridge/IrcBridge.js
```diff
@@ -360,11 +361,20 @@ IrcBridge.prototype._addRequestCallbacks = function() {
             logMessage(req, "IGNORE not mapped");
             return; // these aren't true successes so don't skew graphs
         }
+        else if (res === BridgeRequest.ERR_DROPPED) {
+            logMessage(req, "IGNORE dropped");
+            if (this._timers) {
+                let timer = this._timers[
+                    isFromIrc ? "remote_request_seconds" : "matrix_request_seconds"
+                ];
+                timer.observe({outcome: "dropped"}, req.getDuration() / 1000);
```
lukebarnard1: It would be nice to factor out the 4 lines above as they seem to repeat a lot. Or possibly the entire if, but then you'd still need some sort of `if (...) return` .

Kegsay: Done. https://github.com/matrix-org/matrix-appservice-irc/pull/393/commits/b09bdf4cd752e37ccb052573354b5a407800766b

lib/bridge/IrcBridge.js
Kegsay: Done. https://github.com/matrix-org/matrix-appservice-irc/pull/393/commits/b09bdf4cd752e37ccb052573354b5a407800766b

PR #409, opened 2017-03-29
Speed up PublicitySyncer._solveVisibility by keeping reverse lookups

Kegsay:
Reduces time taken on Freenode's database from 6.3s to 430ms.

### DIFF
--- a/lib/bridge/PublicitySyncer.js
+++ b/lib/bridge/PublicitySyncer.js
@@ -22,6 +22,9 @@ function PublicitySyncer(ircBridge) {
         mappings: {
             //room_id: ['funNetwork #channel1', 'funNetwork channel2',...]
         },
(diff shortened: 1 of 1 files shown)


lib/bridge/PublicitySyncer.js
```diff
@@ -141,7 +144,13 @@ PublicitySyncer.prototype._solveVisibility = Promise.coroutine(function*() {
 
     roomIds.forEach((roomId) => {
         this._visibilityMap.mappings[roomId] = mappings[roomId].map((mapping) => {
-            return this.getIRCVisMapKey(mapping.networkId, mapping.channel);
+            let key = this.getIRCVisMapKey(mapping.networkId, mapping.channel);
```
lukebarnard1: `const`

PR #410, opened 2017-03-29
Do not publish any created rooms to global dir

lukebarnard1:
The PublicitySyncer will take care of doing all publishing.

I've tested this by creating an alias to an existing channel. It is indeed only published to the AS-specific dir and only when +s is not set on the bridged channel.


Kegsay:
LGTM

Issue #388, opened 2017-03-11
Please can we regularly and automatically reap idle-presence connections on all networks


ara4n:
e.g. Snoonet IRC users are enraged at the amount of join/part spam when the bridge farts

Kegsay:
This can't be done until presence on matrix.org Synapse is turned back on.

PureTryOut:
Please giving this some more priority, IRC users are getting more and more enraged every time the bridge has some issue related to connection.

![screenshot_2017-04-07_11-50-08](https://cloud.githubusercontent.com/assets/1565836/24795168/a93a9c8e-1b88-11e7-9410-04c6d1df9660.png)

justjanne:
Especially because the joinspam delays all messages a matrix user sends until after the joinspam is over, so after the joinspam you get another few hundred lines of text spam relevant to stuff that was half an hour ago.

This means the matrix bridge having a hiccup makes the entire channel unusable, far worse than any kind of netsplit.

Kegsay:
I'm not familiar with what you're describing @justjanne - if you can give me a network/channel/timestamp I can look into what happened.

Joinspam doesn't delay messages a matrix user sends (unless they were never joined in which case messages will be stacked up until they join), but we set restrictions on how long ago we will send messages through to IRC, so you shouldn't be getting a "few hundred lines of text spam", which is why I'm interested in looking into this.

justjanne:
@Kegsay: "unless they were never joined", but that's exactly the case when the bridge has an hiccup.

Every matrix user leaves the channel.

All of them rejoin, one after another, and as soon as they rejoin, all they had written in the past minutes will appear at once.

Kegsay:
Hmm, I think I understand what this is then. The check which drops old messages is only done when the message is received from Matrix, it doesn't account for the time taken to reconnect the user to IRC, which would explain why it sends old messages. I can definitely fix that.

As for the hiccup, it looks like Snoonet stopped responding to our pings so we ended up knifing the connections. It took an exorbitant amount of time to reconnect what amounts to a few hundred users because it looks like we still have a global rate-limiting queue in place for Snoonet. I can also disable that, so in the event of a connection problem between the bridge and Snoonet it will reconnect much more rapidly.

Kegsay:
[...]

PR #412, opened 2017-04-11
Check dropMatrixMessagesAfterSecs just before sending to IRC

Kegsay:
This is to handle cases where it is set to drop messages after
5 minutes, the message arrives at 4m30s and it takes 1 minute
to go through the bridge (establish connection, join channel, etc).
Previously, the message would go through, 5m30s after the initial
send. Now, it won't go through.

Fixes issues mentioned in #388 by @justjanne

### DIFF
--- a/lib/irc/IrcServer.js
+++ b/lib/irc/IrcServer.js
@@ -11,8 +11,12 @@ var log = logging.get("IrcServer");
  * @constructor
  * @param {string} domain : The IRC network address
  * @param {Object} serverConfig : The config options for this network.
+ * @param {string} homeserverDomain : The domain of the homeserver
+ * e.g "matrix.org"
+ * @param {Number} expiryTimeSeconds : The max number of seconds a
+ * message can be in order for it to be sent to this server.
  */
-function IrcServer(domain, serverConfig, homeserverDomain) {
+function IrcServer(domain, serverConfig, homeserverDomain, expiryTimeSeconds) {
     this.domain = domain;
     this.config = serverConfig;
 
@@ -22,6 +26,16 @@ function IrcServer(domain, serverConfig, homeserverDomain) {
     }
     this._addresses.push(domain);
     this._homeserverDomain = homeserverDomain;
+    this._expiryTimeSeconds = expiryTimeSeconds;
+}
+
+/**
+ * Get the max number of seconds a message can be before it is dropped from
+ * being sent to this server.
+ * @return {Number} The number of seconds
+ */
+IrcServer.prototype.getExpiryTimeSeconds = function() {
+    return this._expiryTimeSeconds || 0;
 }
 
 /**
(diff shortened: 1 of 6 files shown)


lib/irc/IrcServer.js
```diff
@@ -11,8 +11,12 @@ var log = logging.get("IrcServer");
  * @constructor
  * @param {string} domain : The IRC network address
  * @param {Object} serverConfig : The config options for this network.
+ * @param {string} homeserverDomain : The domain of the homeserver
+ * e.g "matrix.org"
+ * @param {Number} expiryTimeSeconds : The max number of seconds a
+ * message can be in order for it to be sent to this server.
```
lukebarnard1: I'm not sure this makes sense. What does "a message can be" mean?

lib/irc/IrcServer.js
```diff
@@ -22,6 +26,16 @@ function IrcServer(domain, serverConfig, homeserverDomain) {
     }
     this._addresses.push(domain);
     this._homeserverDomain = homeserverDomain;
+    this._expiryTimeSeconds = expiryTimeSeconds;
+}
+
+/**
+ * Get the max number of seconds a message can be before it is dropped from
```
lukebarnard1: again, I'm not sure this is phrased well, see above.

spec/integ/matrix-to-irc.spec.js
```diff
@@ -548,6 +550,47 @@ describe("Matrix-to-IRC message bridging with media URL and drop time", function
...
+        let said = false;
+        env.ircMock._whenClient(roomMapping.server, testUser2.nick, "say",
+        function(client, channel, text) {
+            said = true;
+        });
+
+        let connected = false;
+        env.ircMock._whenClient(roomMapping.server, testUser2.nick, "connect",
+        function(client, cb) {
+            jasmine.clock().tick(20 * 1000); // advance 20s to take it over the drop time
```
lukebarnard1: I would make it obvious that "drop time" is actually `dropMatrixMessagesAfterSecs`

Kegsay: Done. https://github.com/matrix-org/matrix-appservice-irc/pull/412/commits/e02ee95691f1f4407bc6e2a8735aa148c834f14b

lib/irc/IrcServer.js
Kegsay: Done. https://github.com/matrix-org/matrix-appservice-irc/pull/412/commits/e02ee95691f1f4407bc6e2a8735aa148c834f14b

lib/irc/IrcServer.js
Kegsay: Done. https://github.com/matrix-org/matrix-appservice-irc/pull/412/commits/e02ee95691f1f4407bc6e2a8735aa148c834f14b

justjanne:
Aren't you still missing the word "old"? As in, `the maximum number of seconds a message can be old before it is dropped`, or `the maximum age of a message before it is dropped`

Kegsay:
Yep, I've re-phrased with "old" already:

```
 + * @param {Number} expiryTimeSeconds : How old a matrix message can be
 + * before it is considered 'expired' and not sent to IRC. If 0, messages
 + * will never expire.
```

PR #426, opened 2017-05-03
Breaking: rotate logs based on time rather than size

### LINKED ISSUE #326: Log per day not based on log size
This is a limitation of winston, but it's annoying nonetheless. Maybe it's time to look at the state of loggers for node again

Kegsay:
This is a breaking change since I don't really want to maintain 2 separate
ways to rotate logs. Fixes #326 and gives me @erikjohnston's undying love.

We still have support for `maxFiles` though to only keep the most recent N
days of logs.

### DIFF
--- a/lib/logging.js
+++ b/lib/logging.js
@@ -10,7 +10,6 @@ var loggerConfig = {
     logfile: undefined, // path to file
     errfile: undefined, // path to file
     toConsole: true, // make a console logger
-    maxFileSizeBytes: (1024 * 1024 * 128), // 128 MB
     maxFiles: 5,
     verbose: false
 };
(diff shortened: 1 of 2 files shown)


lib/logging.js
```diff
@@ -44,28 +43,28 @@ var makeTransports = function() {
...
             filename: loggerConfig.logfile,
             json: false,
             name: "logfile",
             level: loggerConfig.level,
             timestamp: timestampFn,
             formatter: formatterFn,
-            maxsize: loggerConfig.maxFileSizeBytes,
             maxFiles: loggerConfig.maxFiles,
+            datePattern: ".yyyy-MM-dd",
             tailable: true // most recent stuff always in filename.0.log
```
lukebarnard1: Comment needs an update.

lib/logging.js
```diff
@@ -44,28 +43,28 @@ var makeTransports = function() {
...
             filename: loggerConfig.errfile,
             json: false,
             name: "errorfile",
             level: "error",
             timestamp: timestampFn,
             formatter: formatterFn,
-            maxsize: loggerConfig.maxFileSizeBytes,
             maxFiles: loggerConfig.maxFiles,
+            datePattern: ".yyyy-MM-dd",
             tailable: true // most recent stuff always in filename.0.log
```
lukebarnard1: See previous comment

lib/logging.js
```diff
@@ -44,28 +43,28 @@ var makeTransports = function() {
...
+        transports.push(new (winston.transports.DailyRotateFile)({
             filename: loggerConfig.logfile,
             json: false,
             name: "logfile",
             level: loggerConfig.level,
             timestamp: timestampFn,
             formatter: formatterFn,
-            maxsize: loggerConfig.maxFileSizeBytes,
             maxFiles: loggerConfig.maxFiles,
+            datePattern: ".yyyy-MM-dd",
```
lukebarnard1: I can imagine that people might want to at least configure the `datePattern`, it might be an idea to stick it in config and default it to ".yyyy-MM-dd"

Kegsay: The problem with doing this is that winston will adjust its rotation frequency based on this pattern e.g if you give a pattern which includes hours, it'll rotate hourly. I don't really want to preemptively give the ability to configure this because I think it's going to make more hassle as people flood me with complaints about how it isn't rotating daily because the pattern they used doesn't contain `dd` or does contain `HH` or `mm`.

lib/logging.js
Kegsay: Done. https://github.com/matrix-org/matrix-appservice-irc/pull/426/commits/3ac94fe98276afaed4f73ab63e076bf835f59adc

lib/logging.js
Kegsay: Done. https://github.com/matrix-org/matrix-appservice-irc/pull/426/commits/3ac94fe98276afaed4f73ab63e076bf835f59adc

Kegsay:
@lukebarnard1 PTAL

lib/logging.js
lukebarnard1: Fair enough. Applying semantics to the format of the filename is totally bizarre but oh well!

PR #435, opened 2017-05-16
Make ident address configurable

silkeh:
This adds a configuration option for the listen address for incoming ident requests.

### DIFF
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -321,6 +321,9 @@ ircService:
     # on instead of 113 here.
     # Default: 113.
     port: 1113
+    # The address to listen on for incoming ident requests.
+    # Default: 0.0.0.0
+    address: "::"
 
   # Configuration for logging. Optional. Default: console debug level logging
   # only.
--- a/lib/irc/ident.js
+++ b/lib/irc/ident.js
@@ -20,7 +20,8 @@ var net = require('net');
 var log = require("../logging").get("irc-ident");
 
 var config = {
-    port: 113
+    port: 113,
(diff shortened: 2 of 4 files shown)


lib/irc/ident.js
```diff
@@ -20,7 +20,8 @@ var net = require('net');
 var log = require("../logging").get("irc-ident");
 
 var config = {
-    port: 113
+    port: 113,
+    adddress: "0.0.0.0"
```
Kegsay: > adddress

config.sample.yaml
```diff
@@ -321,6 +321,9 @@ ircService:
     # on instead of 113 here.
     # Default: 113.
     port: 1113
+    # The address to listen on for incoming ident requests.
+    # Default: 0.0.0.0
+    address: "::"
```
Kegsay: If you run this on a machine which does not support IPv6, does this work? I suspect it will fall over, so I would prefer if this value were set to `"0.0.0.0"` instead. The [docs](https://nodejs.org/api/net.html#net_server_listen_port_hostname_backlog_callback) aren't super clear either way.

silkeh: `::` binds for both IPv6 and IPv4 on most operating systems (including Linux and macOS). In which case this works on a machine that doesn't have an IPv6 address.
Also see the conversation for [this pull request](https://github.com/matrix-org/synapse/pull/2232).

Kegsay: SGTM

Kegsay:
LGTM, thanks!

PR #440, opened 2017-05-22
Upload text files with correct UTF-8 encoding

### LINKED ISSUE #439: Long messages should be uploaded as UTF-8
When the IRC bridge sees a long message from matrix it uploads it to the media repository. It currently uploads UTF-8 with a content type of `text/plain`, but the default charset used when none is specified is `us-ascii` causing browsers (and everything else) to use the wrong charset and so display garbled characters.

The bridge should upload with a content type of `text/plain; charset=utf-8`

erikjohnston:
Fixes #439


Kegsay:
LGTM - thanks!

PR #464, opened 2017-06-28
Fix ident response format

silkeh:
Remove the whitespace and add a CR-LF in order to follow the formal syntax specified in RFC1413.


Kegsay:
Right you are, thanks. It's somewhat annoying that the RFC's examples include whitespace:
```
6193, 23 : USERID : UNIX : stjohns
```
though it does later go on to say that you shouldn't use whitespace.

PR #465, opened 2017-06-28
Handle +m by setting events_default to 1 (and back to 0 on -m) 

### LINKED ISSUE #385: No way to see who’s an IRC chanop on Matrix
With ops bridging we have users as Moderators when their ghost is given op. There is currently no way to see which IRC users (that are not ghosts of Matrix users) are ops. Granted, giving them Moderator on Matrix doesn’t make sense, but there should still be an indicator.

### LINKED ISSUE #453: Matrix.org does not enforce IRC +m (mute)
When a matrix user joins a #room in IRC that is moderated,moderates expect control over who speaks in a channel and who is allowed in the channel/room. 

+m channel flag in IRC prevents users from speaking(stands for mute),I use this control to selectively decide who gets to speak in a channel.

Matrix however allows matrix users to see each other messages sent to an IRC room even when the room is set to +m. 

This is unacceptable since it effectively overrides our controls. We treat matrix and other webshells as IRC  clients not as an independently moderated rooms that happen to send messages to irc channels outside of irc channel operators oversight and moderation.

Since this bug effectively bypasses our most basic moderation control,we have been forced to ban matrix.org users from our irc channel.

We truly welcome matrix users and hope the development team,being aware of the limits and nuances of moderating IRC channels considers this to be a very serious bug and fixes it to enable a more complete IRC integration.

Kegsay:
In order for this to work, I also needed to fix #385 so that real IRC users can send into moderated Matrix rooms.

Fixes #453.

### DIFF
--- a/lib/bridge/IrcHandler.js
+++ b/lib/bridge/IrcHandler.js
@@ -654,8 +654,6 @@ IrcHandler.prototype.onPart = Promise.coroutine(function*(req, server, leavingUs
     yield Promise.all(promises);
 });
 
-// TODO: Pass in the prefix letter along with the special char letter so we don't need to
-//       look up an unsafeClient constantly and therefore can make this work for IRC users.
 IrcHandler.prototype.onMode = Promise.coroutine(function*(req, server, channel, by,
                                                 mode, enabled, arg) {
     req.log.info(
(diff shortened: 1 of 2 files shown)


lib/bridge/IrcHandler.js
```diff
@@ -742,6 +746,30 @@ IrcHandler.prototype.onMode = Promise.coroutine(function*(req, server, channel,
...
+    let matrixRooms = yield this.ircBridge.getStore().getMatrixRoomsForChannel(server, channel);
+    // modify power levels for all mapped rooms to set events_default to something >0 so by default
+    // people CANNOT speak into it (unless they are a mod or have voice, both of which need to be
+    // configured correctly in the config file).
+    const botClient = this.ircBridge.getAppServiceBridge().getIntent().getClient();
+    for (let i = 0; i < matrixRooms.length; i++) {
+        const roomId = matrixRooms[i].getId();
+        try {
+            const plContent = yield botClient.getStateEvent(roomId, "m.room.power_levels", "");
+            plContent.events_default = enabled ? 1 : 0;
```
Kegsay: No, Synapse *should* make sure that every room has a `m.room.power_levels` event, and you can't delete state. The CS API states it'll 404 if the event doesn't exist which will throw instead of returning `undefined`.

lukebarnard1: >The CS API states it'll 404 if the event doesn't exist which will throw instead of returning undefined.

Ah ok, then no worries

lib/irc/IrcEventBroker.js
```diff
@@ -360,9 +360,43 @@ IrcEventBroker.prototype.addHooks = function(client, connInst) {
             return null;
         }
         var req = createRequest();
-        return complete(req, ircHandler.onJoin(
+        let promise = complete(req, ircHandler.onJoin(
             req, server, createUser(name.nick), name.chan, "names"
         ));
+        if (!name.opLevel || name.opLevel === "") {
```
Kegsay: Fixed. https://github.com/matrix-org/matrix-appservice-irc/pull/465/commits/ef5c3f11cbeacd1b171707ab4dab1dd3f6e36365

PR #483, opened 2017-08-05
Refine the colours to match the mIRC colours.

silkeh:
The CSS/HTML colour names now better match the colours defined by mIRC.

Comparison (the first word/colour is 'white'):

irccolours

Sources: http://www.mirc.com/colors.html, https://en.wikichip.org/wiki/irc/colors


silkeh:
I just noticed that only the CSS1 colour names work in the Android application (everything except orange).

Kegsay:
LGTM - thanks for this!

PR #485, opened 2017-08-07
Rewrite IRC to HTML parsing

silkeh:
Rewrite the parsing to support more features:

- Background colours
- Reverse colours
- Valid HTML (tags are closed/opened correctly)

The biggest changes are the addition of a function that ensures valid HTML (by keeping a history of open tags) and a new regular expression that only matches the control characters. The latter also simplifies the replace.

### DIFF
--- a/lib/irc/formatting.js
+++ b/lib/irc/formatting.js
@@ -45,15 +45,21 @@ var colorCodesToHtmlNames = {};
 var htmlNames = Object.keys(htmlNamesToColorCodes);
 htmlNames.forEach(function(htmlName) {
     htmlNamesToColorCodes[htmlName].forEach(function(colorCode) {
-        colorCodesToHtmlNames[colorCode] = htmlNamesToHex[htmlName];
+        colorCodesToHtmlNames[colorCode.substr(1)] = htmlNamesToHex[htmlName];
     });
 });
 
-var STYLE_BOLD = '\u0002';
-var STYLE_ITALICS = '\u001d';
-var STYLE_UNDERLINE = '\u001f';
-var STYLE_CODES = [STYLE_BOLD, STYLE_ITALICS, STYLE_UNDERLINE];
-var RESET_CODE = '\u000f';
+const STYLE_BOLD = '\u0002';
+const STYLE_ITALICS = '\u001d';
+const STYLE_UNDERLINE = '\u001f';
+const STYLE_CODES = [STYLE_BOLD, STYLE_ITALICS, STYLE_UNDERLINE];
+const RESET_CODE = '\u000f';
+const REVERSE_CODE = '\u0016';
+const STYLE_DEFAULT_STATE = {
+    "color": null,
+    "bcolor": null,
+    "history": []
+};
 
 function escapeHtmlChars(text) {
     return text
@@ -62,6 +68,50 @@ function escapeHtmlChars(text) {
          .replace(/>/g, "&gt;")
          .replace(/"/g, "&quot;")
          .replace(/'/g, "&#39;"); // to work on HTML4 (&apos; is HTML5 only)
+}
+
+function htmlTag(state, name, open) {
+    let text = '';
+
+    if (typeof open === 'undefined') {
+        open = (state.history.indexOf(name) === -1);
+    }
+
+    if (open) {
+        state.history.push(name);
+        if (name === 'font') {
+            // Create style from state
+            let style = '';
+            if (state.color) {
+                style += ' color="' + state.color + '"';
+            }
+            if (state.bcolor) {
+                style += ' data-mx-bg-color="' + state.bcolor + '"';
+            }
+            text = '<font' + style + '>';
+        }
+        else {
+            text = '<' + name + '>';
+        }
+    }
+    else {
+        // Get tags that need to be closed
+        let index = name === 'all' ? 0 : state.history.lastIndexOf(name);
+        let tags = state.history.splice(index);
+
+        // Close tags
+        tags.reverse().forEach( function(t) {
+            text += '</' + t + '>';
+        });
+
+        // Open tags again
+        if (name !== 'all') {
+            tags.slice(0, -1).forEach( function (t) {
+                text += htmlTag(state, t, true);
+            });
+        }
+    }
+    return text;
 }
 
 module.exports.stripIrcFormatting = function(text) {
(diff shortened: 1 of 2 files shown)


Kegsay:
I haven't forgotten about this, I just need to find the time to devote to this as it's important to get this vetted correctly which takes a lot of time.

lib/irc/formatting.js
```diff
@@ -64,6 +70,50 @@ function escapeHtmlChars(text) {
          .replace(/'/g, "&#39;"); // to work on HTML4 (&apos; is HTML5 only)
 }
 
+function htmlTag(state, name, open) {
```
Kegsay: Please comment what you expect this function to do.

lib/irc/formatting.js
```diff
@@ -164,103 +214,67 @@ module.exports.htmlToIrc = function(html) {
...
-            }
-            else {
-                // unknown font colour
-                text = text.replace(groups[0], coloredText);
-            }
-            groups = colorRegex.exec(text);
-        }
-    }
+    // Escape HTML characters and add reset character to close all tags at the end.
+    text = escapeHtmlChars(text) + '\x0f';
```
Kegsay: Nit: We have a `RESET_CODE` const, let's use it please.

lib/irc/formatting.js
```diff
@@ -164,103 +214,67 @@ module.exports.htmlToIrc = function(html) {
...
 
-    // force there to always be a RESET code at the end of the text to make sure
-    // we close up all open style tags at the end.
-    if (text.length > 0 && text[text.length - 1] !== RESET_CODE) {
-        text = text + RESET_CODE;
-    }
+    // Replace all mIRC formatting characters.
+    // The color character can have arguments.
+    const colorRegex = /[\x02\x1d\x1f\x0f\x16]|\x03(\d{0,2})(?:,(\d{1,2}))?/g;
```
Kegsay: You're going to need to do a lot more to convince me that this regexp works. Tell me:
 - What it matches (ideally with an example)
 - What groups does it extract

If this is from another piece of code (after all, we aren't the first people to want to do this), then please link to documentation/source to aid verification.

lib/irc/formatting.js
```diff
@@ -164,103 +214,67 @@ module.exports.htmlToIrc = function(html) {
...
-                if (boldOpen) {
-                    charsToInsert += "</b>";
+                // Close tags
+                tags = htmlTag(state, 'all', false);
+                // Reset state
+                state = Object.assign({}, STYLE_DEFAULT_STATE);
+                return tags;
+
+            default:
+                // Close font tag
```
Kegsay: Why do we do all this in the `default` block? When will this execute?

lib/irc/formatting.js
```diff
@@ -164,103 +214,67 @@ module.exports.htmlToIrc = function(html) {
...
+                return htmlTag(state, 'i');
+
             case STYLE_UNDERLINE:
-                underlineOpen = !underlineOpen;
-                charsToInsert = underlineOpen ? "<u>" : "</u>";
-                break;
+                return htmlTag(state, 'u');
+
+            case REVERSE_CODE:
+                state.color = [state.bcolor, state.bcolor = state.color][0];
```
Kegsay: Style: What are you actually doing on this line? It looks like you're doing destructuring assignment coupled with array accesses. Can you please split this up onto multiple lines to aid readability.

lib/irc/formatting.js
```diff
@@ -45,15 +45,21 @@ var colorCodesToHtmlNames = {};
...
+const STYLE_BOLD = '\u0002';
+const STYLE_ITALICS = '\u001d';
+const STYLE_UNDERLINE = '\u001f';
+const STYLE_CODES = [STYLE_BOLD, STYLE_ITALICS, STYLE_UNDERLINE];
+const RESET_CODE = '\u000f';
+const REVERSE_CODE = '\u0016';
+const STYLE_DEFAULT_STATE = {
+    "color": null,
+    "bcolor": null,
+    "history": []
```
Kegsay: You need to go into much more detail about:
 - What will `history` contain? When will things get added/removed from it?
 - What will `color` and `bcolor` be? What does `null` represent?

Kegsay:
Overall, this looks like an elegant solution, and it's pleasing that the tests pass mostly without modification.

However, I'm not convinced that the state machine is adequately documented, which will be important should we ever want to make changes to it. I've left some open-ended questions on this, please take some time to explain non-obvious sections.

lib/irc/formatting.js
silkeh: Swapping colours. This was a quick replacement for `[state.color, state.bcolor] = [state.bcolor, state.color]` (which is ES6 syntax). I've replaced it with something readable, because this isn't.

lib/irc/formatting.js
silkeh: We shouldn't. I have replaced it by a case for `STYLE_COLOR`.

silkeh:
Thanks for the feedback. I have attempted to improve the clarity of the code. Let me know if something is still unclear.

lib/irc/formatting.js
```diff
@@ -70,6 +80,21 @@ function escapeHtmlChars(text) {
...
+ * appropriate attributes.
+ *
+ * Tags are stored in an array, which is used as a stack.
+ * Opening a tag is pushed to the end.
+ * Closing a tag closes all tags up to the given name,
+ * and re-opens the ones below.
+ *
+ * @param {Object} state Current state of a message
+ * @param {string} name Name of the HTML tag.
+ * @param {boolean} open Open or close the tag (optional) .
```
Kegsay: What happens if you open twice or close twice?

lib/irc/formatting.js
```diff
@@ -215,10 +240,16 @@ module.exports.htmlToIrc = function(html) {
...
+    text = escapeHtmlChars(text) + RESET_CODE;
 
     // Replace all mIRC formatting characters.
     // The color character can have arguments.
+    // The regex matches:
+    // - Any single 'simple' formatting character: \x02, \x1d, \x1f, \x0f and
+    //   \x16 for bold, italics, underline, reset and reverse respectively.
+    // - The colour formatting character (\x03) followed by 0 to 2 digits for
+    //   the foreground colour and (optionally) a comma and 1-2 digits for the
+    //   background colour.
```
Kegsay: Excellent, this is exactly what I was hoping to see. Thanks!

lib/irc/formatting.js
```diff
@@ -241,7 +272,11 @@ module.exports.ircToHtml = function(text) {
                 return htmlTag(state, 'u');
 
             case REVERSE_CODE:
-                state.color = [state.bcolor, state.bcolor = state.color][0];
+                // Swap the foreground and background colours.
+                var temp = state.color;
```
Kegsay: Nit: We should probably be using `let`.

Kegsay:
Otherwise LGTM, thank you for this!

lib/irc/formatting.js
silkeh: Opening twice adds the tag to the history twice. Closing twice closes everything after the second, re-opens everything that came after the second, closes everything up to the first (which includes everything that was re-opened) and re-opens everything again.

In summary:
Opening `span`, `span`, `b` results in: `<span>`, `<span>`, `<b>` respectively (with history being `["span", "span", "b"]`).
Subsequently closing  `span`, `span`, `b` results in: `</b></span><b>` (history `["span", "b"]`), `</b></span><b>` (history `["b"]`), `</b>` (history `[]`) respectively.

This isn't very efficient, but at least it's valid HTML.

I have updated the description to make it a bit more clear.

Kegsay:
LGTM!

Issue #48, opened 2016-06-02
Propagate matrix display name changes to IRC (and vice versa)


Kegsay:
https://matrix.org/jira/browse/BOTS-52

Matrix to irc propagation is hard because the two protocols scope names differently. Matrix allows per room display names but irc is per connection (global). We could manage this by just applying the most recent name change, but we need to accept mismatches are unavoidable.
Irc to matrix propagation is possible, though fiddly to implement (since the nick change necessitates a new user id, invites to all of the rooms the user was previously in, etc).

After thinking about this more, I wonder whether it would be more accurate to map Matrix user IDs to IRC user@host instead - in which case nick changes are just display name changes. Other points still hold though.

jansol:
This seems to summarize #72 and #71 .

Mikaela:
Is this issue about syncing display name between Matrix and IRC? 

How does that work if the registered nickname at IRC is different from the Matrix one or registered/used by someone else? What if the nickname can be used at Matrix and Mozilla, but is registered to someone else at freenode or OFTC?

Does this try to deprecate !nick?

jansol:
IIRC this predates the persistently saved !nick.

This also gets more complicated if/when per-room display names are introduced. For now, how about making !nick accept a template like in the config file? Or even just "$DISPLAYNAME".

Mikaela:
I think https://github.com/matrix-org/matrix-appservice-irc/pull/486 was supposed to close this.

PR #486, opened 2017-08-09
Propagate display name changes to nick changes

### LINKED ISSUE #218: Slack bridged users are given useless IRC nicknames
Users who speak on the Slack side of the integration is given IRC nicknames like M_slack_<org>_7 which makes it impossible to know who is actually talking.

They should be given sane nicknames like which happens with gitter bridged users.
Could be related to the display name being set on Matrix seemingly after the IRC user is created.

silkeh:
Any join event now changes the nickname (if different).
This is not done for configured nicknames (stored in the database).
Generated nicknames are now stored as null.

~This is a work in progress as tests are required for the following situations:~

- ~A user without custom nick changes their display name.~
- ~A user with a configured nick changes their display name.~
- ~A user without custom nick sets a nick and then changes their display name.~

Feedback would be appreciated!

Resolves #218

### DIFF
--- a/lib/bridge/MatrixHandler.js
+++ b/lib/bridge/MatrixHandler.js
@@ -321,7 +321,7 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
         // change the nick
         let bridgedClient = yield this.ircBridge.getBridgedClient(ircServer, event.user_id);
         try {
-            let response = yield bridgedClient.changeNick(nick);
+            let response = yield bridgedClient.changeNick(nick, true);
             let noticeRes = new MatrixAction("notice", response);
             yield this.ircBridge.sendMatrixAction(adminRoom, botUser, noticeRes, req);
             // persist this desired nick
@@ -837,7 +837,6 @@ MatrixHandler.prototype._onJoin = Promise.coroutine(function*(req, event, user) 
             // Check for a displayname change and update nick accordingly.
             if (event.content.displayname !== bridgedClient.displayName) {
                 bridgedClient.displayName = event.content.displayname;
-
                 // Changing the nick requires that:
                 // - the server allows nick changes
                 // - the nick is not custom
@@ -845,7 +844,8 @@ MatrixHandler.prototype._onJoin = Promise.coroutine(function*(req, event, user) 
                     bridgedClient.getClientConfig().getDesiredNick() === null
                 ) {
                     bridgedClient.changeNick(
-                        room.server.getNick(bridgedClient.userId, event.content.displayname));
+                        room.server.getNick(bridgedClient.userId, event.content.displayname),
+                        false);
                 }
             }
 
--- a/lib/bridge/MatrixHandler.js
+++ b/lib/bridge/MatrixHandler.js
@@ -840,8 +840,11 @@ MatrixHandler.prototype._onJoin = Promise.coroutine(function*(req, event, user) 
                 // Changing the nick requires that:
                 // - the server allows nick changes
                 // - the nick is not custom
+                let config = yield self.ircBridge.getStore().getIrcClientConfig(
+                    bridgedClient.userId, room.server.domain
+                );
                 if (room.server.allowsNickChanges() &&
-                    bridgedClient.getClientConfig().getDesiredNick() === null
+                    config.getDesiredNick() === null
                 ) {
                     bridgedClient.changeNick(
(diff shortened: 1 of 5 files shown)


14mRh4X0r:
How would this work with room-specific display names?

silkeh:
All changes to display names are propagated to nick changes. This means that the last set display name for a bridged room on an IRC network will be the nickname on that network..

Kegsay:
I'm not sure how I feel about this. What's the rationale behind it?

silkeh:
This implements the Matrix to IRC part of #48.

This PR, specifically, was prompted by issues with both the Discord and Slack bridges, where the display name may not have been set when joining a bridged channel (but is set afterwards). This results in unreadable nicknames because the MXID for both is a number.

erdnaxeli:
Kegsay: I just deployed an matrix server for a little community, bridged to an IRC server. Some users get confused with the dissociation between the nick and the display name. They expect them to be related.

Kegsay:
> They expect them to be related.

This isn't possible in a lot of cases though. E.g. anything non-ascii will not be supported as a nick, anything which doesn't abide by the nick rules (must start with a letter, etc) won't be supported as a nick, etc.

The question is "Will the confusion be more or less by merging this PR?". At the moment it's pretty noddy and clear what will happen:
 - It'll make you a random nick based off the user ID or display name if you had one.
 - Change it via `!nick`.

If we merge this PR, it gets more confusing:
 - It'll make you a random nick based off the user ID or display name if you had one.
 - If you change your display name, it'll update your nick on a best-effort basis (changing to Japanese characters will result in a nick of `[m]` since all the unicode will be stripped...
 - If you have different display names on different rooms (most clients don't support this but the protocol *does*), then the most-recent change will be your nick.
 - If you set your nick via `!nick` it'll use that forever, regardless of whether you update your display name.

silkeh:
> >  They expect them to be related.
>
> This isn't possible in a lot of cases though. E.g. anything non-ascii will not be supported as a nick, anything which doesn't abide by the nick rules (must start with a letter, etc) won't be supported as a nick, etc.

I would still call that 'related'. It's filtered and changed to abide by some rules, sure, but it's still based on the display name or user ID.

> It'll make you a random nick based off the user ID or display name if you had one.

I wouldn't describe it as 'random' if it's based on something either 😉 

> changing to Japanese characters will result in a nick of [m] since all the unicode will be stripped...

This is also the case if the display name was all Japanese characters in the first place, this PR doesn't affect that.

I disagree that the new situation would be more confusing: The nickname is currently based on the display name provided in the first join event for a room bridged to a network. From a user perspective this could be *any* room joined somewhere in the past (the user doesn't even have to be in this room any more).

The new and old situation could also be described differently. In the current situation the nick is:

- based on the display name that was set for some room somewhere in the past, or
- configured manually.

In the new situation, the nick is:

- based on the current display name for some room, or
- configured manually.

Kegsay:
Okay, you've convinced me this is probably worth doing then. Fixup travis failures and I'll give this a look.

silkeh:
Thanks! Fixed the failure, let me know where things need changing.

lib/bridge/MatrixHandler.js
```diff
@@ -833,6 +833,25 @@ MatrixHandler.prototype._onJoin = Promise.coroutine(function*(req, event, user)
             let bridgedClient = yield self.ircBridge.getBridgedClient(
                 room.server, user.getId(), (event.content || {}).displayname
             );
+
+            // Check for a displayname change and update nick accordingly.
+            if (event.content.displayname !== bridgedClient.displayName) {
+                bridgedClient.displayName = event.content.displayname;
```
Kegsay: Presumably we don't persist this, so every bridge restart followed by `onJoin` will trigger a nick change once for each user who doesn't have an explicit display name set?

Kegsay:
Otherwise LGTM

lib/bridge/MatrixHandler.js
silkeh: In that case both `event.content.displayname` and `bridgedClient.displayName` should be `null` because `bridgedClient.displayName`  is initialised as follows:

```JavaScript
this.displayName = matrixUser ? this.matrixUser.getDisplayName() : null;
```

Kegsay: SGTM

Kegsay:
LGTM! Thanks for all your work on this.

PR #495, opened 2017-08-25
migrate-users script

SohumB:
To remove suffixes from users the bridge controls. Hacky rate-limiting to 10/s; this resulted in a few rate-limited requests on my homeserver but only maybe 5% of requests failed, a quick rerun was the easiest way to solve that.

I get that you may not want to run this on matrix.org directly, but it's still useful to those of us running our own bridges!

### DIFF
--- a/scripts/migrate-users.py
+++ b/scripts/migrate-users.py
@@ -50,7 +50,6 @@ def migrate_displayname(uid, oldname, suffix, homeserver, token):
         print(res.json())
         if 'M_NOT_JSON' in str(res.json()):
             print("python-requests strips the body of the request on 301 redirects (https://github.com/requests/requests/issues/2590). Make sure you're using the direct url of your homeserver.")
-    time.sleep(0.1)
 
 def main(registration, homeserver, prefix, suffix):
     token = get_appservice_token(registration)
@@ -62,6 +61,7 @@ def main(registration, homeserver, prefix, suffix):
     merged_users = { k: v for d in per_room_users for k,v in d.items() }
     for uid, display in merged_users.iteritems():
         migrate_displayname(uid, display, suffix, homeserver, token)
+        time.sleep(0.1)
 
 
 if __name__ == "__main__":
--- a/scripts/migrate-users.py
+++ b/scripts/migrate-users.py
@@ -57,8 +57,9 @@ def main(registration, homeserver, prefix, suffix):
         raise Exception("Cannot read as_token from registration file")
 
     rooms = get_rooms(homeserver, token)
-    users = { k: v for d in [get_users(homeserver, room, token, prefix, suffix) for room in rooms] for k,v in d.items() }
-    for uid, display in users.iteritems():
+    per_room_users = [get_users(homeserver, room, token, prefix, suffix) for room in rooms]
+    merged_users = { k: v for d in per_room_users for k,v in d.items() }
+    for uid, display in merged_users.iteritems():
         migrate_displayname(uid, display, suffix, homeserver, token)
 
 
--- a/scripts/migrate-users.py
+++ b/scripts/migrate-users.py
@@ -1,0 +1,80 @@
+#!/usr/bin/env python
+from __future__ import print_function
+import argparse
+from datetime import datetime
+import sys
+import json
+import yaml
+import urllib
+import requests
+import time
+
+# import logging
+# import httplib as http_client
+# http_client.HTTPConnection.debuglevel = 1
+# logging.basicConfig()
+# logging.getLogger().setLevel(logging.DEBUG)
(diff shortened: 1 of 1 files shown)


scripts/migrate-users.py
```diff
@@ -0,0 +1,80 @@
...
+        return reg_yaml["as_token"]
+
+def get_users(homeserver, room_id, token, user_prefix, name_suffix):
+    res = requests.get(homeserver + "/_matrix/client/r0/rooms/" + urllib.quote(room_id) + "/joined_members?access_token=" + token)
+    joined = res.json().get("joined", None)
+    user_ids = [user_id for user_id in joined if user_id.startswith(user_prefix) and (joined.get(user_id).get("display_name") or "").endswith(name_suffix) ]
+    return { uid: joined.get(uid).get("display_name") for uid in user_ids }
+
+def get_rooms(homeserver, token):
+    res = requests.get(homeserver + "/_matrix/client/r0/sync?access_token=" + token).json()
```
Kegsay: You don't want to do this. On large bridges this will absolutely destroy Synapse. Use `/joined_rooms` instead. https://github.com/matrix-org/matrix-appservice-bridge/blob/8f60f3e84a98d8bfd4dbd52d5403ff1c841f7d32/lib/components/app-service-bot.js#L40

scripts/migrate-users.py
```diff
@@ -0,0 +1,80 @@
...
+            print("python-requests strips the body of the request on 301 redirects (https://github.com/requests/requests/issues/2590). Make sure you're using the direct url of your homeserver.")
+    time.sleep(0.1)
+
+def main(registration, homeserver, prefix, suffix):
+    token = get_appservice_token(registration)
+    if not token:
+        raise Exception("Cannot read as_token from registration file")
+
+    rooms = get_rooms(homeserver, token)
+    users = { k: v for d in [get_users(homeserver, room, token, prefix, suffix) for room in rooms] for k,v in d.items() }
```
Kegsay: Come on now.. can you make this a little more comprehensible for me please :)

scripts/migrate-users.py
```diff
@@ -0,0 +1,80 @@
...
+def get_rooms(homeserver, token):
+    res = requests.get(homeserver + "/_matrix/client/r0/sync?access_token=" + token).json()
+    room_ids = []
+    for room_id in res["rooms"]["join"]:
+        room = res["rooms"]["join"][room_id]
+        room_ids.append(room_id)
+    return room_ids
+
+def migrate_displayname(uid, oldname, suffix, homeserver, token):
+    newname = oldname.replace(suffix, "")
```
Kegsay: Won't this incorrectly apply if `suffix` isn't a suffix? 0:)

scripts/migrate-users.py
```diff
@@ -0,0 +1,80 @@
...
+    print("Migrating %s from %s to %s" % (uid, oldname, newname))
+    headers = { 'Content-Type': 'application/json' }
+    res = requests.put(homeserver + "/_matrix/client/r0/profile/" + urllib.quote(uid) + "/displayname?access_token=" + token + "&user_id=" + urllib.quote(uid),
+                       data = json.dumps({ 'displayname': newname }), headers=headers)
+    if res.json():
+        print(res.json())
+        if 'M_NOT_JSON' in str(res.json()):
+            print("python-requests strips the body of the request on 301 redirects (https://github.com/requests/requests/issues/2590). Make sure you're using the direct url of your homeserver.")
+    time.sleep(0.1)
```
Kegsay: Nit: I would prefer to have the sleep be a part of the for loop which calls `migrate_displayname` than have it here.

Kegsay:
Please base this off `develop`.

scripts/migrate-users.py
SohumB: Will do; I'm just not a python user so I have no idea what is idiomatic for them :p

SohumB:
Sorry it still says from `Sohumb:master`, I didn't follow best practice of making a feature branch, and github doesn't let you change the branch a PR is _from_, so I had to overwrite my fork's `master` to base off develop instead!

scripts/migrate-users.py
Kegsay: It generally won't matter what is idiomatic, clarity / readability should always take priority. Here's a [StackOverflow](https://stackoverflow.com/questions/8068251/why-is-python-list-comprehension-sometimes-frowned-upon) post discussing this a bit more. tl;dr:

> list comprehensions that either
>
> - Have side effects, or
> - Are more readable as a multi-line for loop
are generally frowned upon.
>
> In other words, readability is important.

Kegsay:
LGTM, thank you!

PR #503, opened 2017-09-25
Set custom power_levels on room creation

erdnaxeli:
An user can already set custom level for IRC modes. The idea of this PR is to allow to set also custom privilege levels needed to configure the room (Matrix side).

My first need was to be able to let the matrix users who are op on IRC set the history visibility, but I added some other parameters (avatar, name, canonical alias, ...).

Edit: currently this PR only set power levels on room creation, but I would like to set them on reconnection, to support config changes. Any advices on how to do this?

### DIFF
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -97,6 +97,16 @@ ircService:
       modePowerMap:
         o: 50
 
+      # A map of levels needed to do some action on the Matrix room. Those levels will be set
+      # on the room on creation. With corresponding modePowerMap, this allows Matrix users who
+      # are ops on the IRC side to edit the room settings.
+      powerLevels:
+        avatar: 50
+        name: 50
+        canonicalAlias: 50
+        historyVisibility: 50
+        redact: 50
+
       botConfig:
         # Enable the presence of the bot in IRC channels. The bot serves as the entity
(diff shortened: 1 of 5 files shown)


Kegsay:
What's the use case for this?

erdnaxeli:
Let Matrix users who are op on IRC set the history visibility, mainly. And have more precision (for +h, +o, +a, +q).

Kegsay:
Can't they do this already via plumbing?

erdnaxeli:
Plumbing is a complex operation, I prefer to use portal rooms, and that's
what I've documented for my synapses users.

Kegsay:
Hmm, the difference is that plumbed rooms allow configuration on a per-channel basis, whereas this PR applies on a per-bridge basis. Configuration of sensitive things like message history really should be a per-channel thing, so the idea of adding a per-bridge configuration isn't too appetising.

That being said, flexibility is always a good thing, and there may be certain use cases where this makes sense e.g. private IRC networks. I'll have a look at this in detail at some point.

erdnaxeli:
I agree, this is a sensitive setting and history visibility should be let
to level 100 for big public bridges. But my use case is a small bridge, and
the irc and HS / bridge admins are the same.

I see this as a complementary feature to the power level map.

Half-Shot:

PR #515, opened 2017-10-27
Kick user regardless of his type on kick from irc

erdnaxeli:
With this modification, when a kick occurs from an IRC user, the bot always kick the kickee, regardless if it's an IRC user or a Matrix user. This makes kick much more cleaner, and no more "user leaving but actually it's a kick".

### DIFF
--- a/lib/bridge/IrcHandler.js
+++ b/lib/bridge/IrcHandler.js
@@ -534,72 +534,39 @@ IrcHandler.prototype.onKick = Promise.coroutine(function*(req, server,
 
     /*
     We know this is an IRC client kicking someone.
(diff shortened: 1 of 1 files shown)


erdnaxeli:
I break the test as I changed the behavior, but I have no idea how to fix it. I've looked at the test for an IRC user kicking a Matrix user but it's so different that this one…

Kegsay:
The point of making users self-leave rooms was to make sure that the membership list would remain consistent.

What happens in your PR if the kicker does not have permission to kick on the other side of the bridge?

Half-Shot:

Issue #71, opened 2016-06-15
Track nick changes on IRC


jansol:
IRC users changing nicks should not spawn new matrix users, rather update their display names.
#72 is needed for this.

neilalexander:
This still seems to be a problem - if a user in IRC changes their nick then a new user appears on the Matrix side, the old nick still seems to sit in the channel forever and "dead" users on the Matrix side never ever get cleaned up by the appservice.

Is there any plan to fix this?

Kegsay:
> even after a server restart or with full membership syncing enabled

Restarting the *bridge* will remove those users, assuming initial leave syncing is enabled, as that will do a full sweep of all bridged rooms looking for users to evict.

As an aside, it may surprise you to know that comments like the following on a project which is actively maintained and triaged come across as pretty rude.

> This still seems to be a problem

Really? I wonder if that's why this issue is still open.

> Is there any plan to fix this? 

No, I like leaving issues open forever and never plan on fixing them. I just [triage](https://github.com/matrix-org/matrix-appservice-irc/issues?q=is%3Aissue+label%3Awontfix+is%3Aclosed) issues for fun.

---

ansiwen:
@Kegsay: This is a big issue IMHO. I'm wondering why this is not a problem for others, still think I might do something wrong. When an IRC user joins a bridged channel, and changes it's nick, this will be not reflected on the matrix side, although member sync is enabled. So I don't even see the "spawn new matrix user", which would at least work somehow. When this user leaves the channel, the old "dead" nick will still be left over, because the user with the new nick is tried to be removed. This is not a feature request, but a bug in my eyes.

I checked the code, and the "nick" command from node-irc is not taken care of at all. Why not handling that and change the display name of the according virtual matrix user? I have a patch for this ready, and will try it now. But I'm not sure if I forgot something important, because this issue is tagged with "refactor".

jansol:
@ansiwen the hard part about this is that completely unrelated people can be using the same IRC nick. What should happen when Bob changes their nick to Alice, goes offline and then Charlie joins and changes his nick to Bob? This and other (non-)edge cases break tons of stuff starting from 1:1 chats. The only way to retain a semblance of sanity is to generate ghost users based on the IRC users' host+username, which has its own share of Fun.

Anyway if this is an important issue to you and you are interested in contributing I recommend starting from #72

erdnaxeli:
> Why not handling that and change the display name of the according virtual matrix user?

[...]

PR #517, opened 2017-10-30
Track nick changes by parting and joining a new user

ansiwen:
The only way to track nick changes on the IRC side at the moment is
to let the user leave the channel and letting a new user with the new
nick join. This looks weird but still is better than many zombie
nicks and hidden users in bridged channels.

Fixes: #71

### DIFF
--- a/lib/irc/IrcEventBroker.js
+++ b/lib/irc/IrcEventBroker.js
@@ -347,6 +347,18 @@ IrcEventBroker.prototype.addHooks = function(client, connInst) {
             req, server, createUser(nick), chan, "join"
         ));
     });
+    this._hookIfClaimed(client, connInst, "nick", function(oldNick, newNick, chans, msg) {
+        chans = chans || [];
+        chans.forEach((chan) => {
+            const req = createRequest();
+            complete(req, ircHandler.onPart(
+                req, server, createUser(oldNick), chan, "nick"
+            ));
+            complete(req, ircHandler.onJoin(
+                req, server, createUser(newNick), chan, "nick"
+            ));
+        });
+    });
     // bucket names and drain them one at a time to avoid flooding
     // the matrix side with registrations / joins
     var namesBucket = [


erdnaxeli:
To be a lil less weird, I though the user leaving could do "/me is now known as <new nick>" and then leave. Or even better, it could add a reason to the leaving, but I don't know if this is supported yet by the protocol and Matrix.

ansiwen:
@erdnaxeli I added a "* foo is now known as bar" message

Kegsay:
This looks good, though there's a risk that this can be used as a spam vector, since it triggers 3 messages in each room. I guess you can rely on IRC side rate limiting here.

LGTM.

ansiwen:
In large channels the "emote" message is pretty noisy. I now changed it to a "notice" message. However, even this could be too noisy. I don't know what is more annoying, noisy nick-change messages or not having that bit of information...

erdnaxeli:
Not having the information is definitively more annoying.

ansiwen:
@erdnaxeli well, you would still have the information that ansiwen left and ansiwen_brb joined. So in most cases nickname changes are pretty obvious in the leave and join notifications (which have the advantage of being displayed small and are being collapsed). Maybe configurable, if this is easy to do?

erdnaxeli:
What if ansiwen leaved and erdnaxeli joined right after? It is a nick change? It is a real leave and join?

I think the message is mandatory, but it can be configurable.

ansiwen:
@erdnaxeli yes, in these edge-cases you couldn't tell, that's true. But still a lot better than right now, where you _never_ can tell what is going on. Anyway, I'm fine with the "notice", because it's displayed less prominent (at least in Riot).

erdnaxeli:
Ok, I just tried this PR, and there is a bug. Consider this:
* a user enter a bridged room
* the bridge open a new connection to the IRC server
* the nick is registered, after 60s NickServ force the nick change
* the bridge see a nick event, make a new user join => fail

We must check is the user is virtual or not before doing the notice-part-join. I got an even weirder case when doing the notice made the user join first, then the notice, then the leave did not happened (?), then the new used joined.

ansiwen:
@erdnaxeli I don't understand your issue and I'm not sure it's related to your change. The functions I am calling (onPart and inJoin) make sure it's not a virtual IRC user and it only works with virtual Matrix users: https://github.com/matrix-org/matrix-appservice-irc/blob/master/lib/bridge/IrcHandler.js#L624-L626

ansiwen:
So, we are using this patch on our internal matrix server with a bridge to freenode and an internal irc network, and it seems to work well. @ara4n, if you have a second, what do you think?

erdnaxeli:
@ansiwen: here is what happened 2017-11-06-202025_648x52_scrot. The thing is JuCh is a Matrix user.

The problem I guess is based on the fact that `virtual` what set to `false`, but I don't know why.

ansiwen:
@erdnaxeli So, as I see it, you are seeing an unrelated bug here, that was invisible before, right? I don't see, how to do it differently, because both functions (onPart and onJoin) make the correct tests and rely on a correct state of the bridge registry. I would file a new bug. If JuCh is a matrix user, why does the localpart starts with "@irc_"? I see the problem rather in JuCh not being removed than Poiral joining.

erdnaxeli:
JuCh is a real Matrix user, it can't be removed. @irc_JuCh is the same user, seen from irc.

This happened only on the boot first sync right after start.

erdnaxeli:
I can confirm, when restarting the bridge, if you a Matrix user got renamed by the IRC server (because he did not authentify) the bridge totally fails. Reproduced again.

erdnaxeli:
It's worth than I thought, I got this bug (matrix user seen as irc one) today, but the bridge was running since days (so it was not in the first sync).
2017-11-25-121028_521x189_scrot

erdnaxeli:
@ansiwen It's was indeed a bug in another part of the code. #523 fixes it.

ansiwen:
So, what about merging this then? I'm using this on our internal matrix server without problems...

ansiwen:
ping

ansiwen:
ping @ara4n @Kegsay

ansiwen:
ping @Half-Shot 
since you seem to be the only active one recently, maybe you want to have a look at this PR? I think it is certainly better to have it than not.

ansiwen:
@Half-Shot: ok, cool! :-) btw, we at Red Hat found the "is now known as" notice very annoying, so we commented it out for our bridges. I'm happy to remove it from this PR as well. :-)

Half-Shot:
I suspect most of the hosted bridges would also find it annoying, sorry :(. Would be better removed.

Half-Shot:
Otherwise, code looks good. Can we have a test to ensure that when a nick change happens, we do the right thing :angel:

lib/irc/IrcEventBroker.js
```diff
@@ -347,6 +347,18 @@ IrcEventBroker.prototype.addHooks = function(client, connInst) {
             req, server, createUser(nick), chan, "join"
         ));
     });
+    this._hookIfClaimed(client, connInst, "nick", function(oldNick, newNick, chans, msg) {
+        chans = chans || [];
+        chans.forEach(function(chan) {
```
Half-Shot: Arrow functions please. ``(chan) => {``

lib/irc/IrcEventBroker.js
```diff
@@ -347,6 +347,18 @@ IrcEventBroker.prototype.addHooks = function(client, connInst) {
...
+            var req = createRequest();
```
Half-Shot: ``var`` is evil, please use ``const`` here

lib/irc/IrcEventBroker.js
ansiwen: done

lib/irc/IrcEventBroker.js
ansiwen: done

Half-Shot:
I'm happy to accept this without a test on further examination, it's going to be a lot of work to mess around with that area for testing. Thanks for the PR and sorry it took a long time to get to it!

PR #523, opened 2017-11-26
On nick change, register the BridgedClient with the new nick

erdnaxeli:
This fixes my bugs with the PR #517.


erdnaxeli:
I need to give it a look again. IIRC it was not actually totally fixing my bug.

Half-Shot:

PR #526, opened 2017-12-03
Add a servce file for systemd

erdnaxeli:
Here is a example of service file for systemd (actually it's just a copy of the one I use).

I don't know if I put it on the good location.

### DIFF
--- a/scripts/matrix-appservice-irc.service
+++ b/scripts/matrix-appservice-irc.service
@@ -1,0 +1,13 @@
+[Unit]
+Require=network.target
+After=network.target
+
+[Service]
+Type=simple
+User=bridge
+Goup=bridge
+ExecStart=/home/bridge/node/bin/node /home/bridge/matrix-appservice-irc/app.js -c /home/bridge/matrix-appservice-irc/config.yaml -f /home/bridge/matrix-appservice-irc/registration.yaml -p 9999
+Restart=on-failure
+
+[Install]
+WantedBy=multi-user.target


silkeh:
When running the bridge as a system service, I don't think it should run from a home directory (or to be more precise: the home directory should be located in `/var/lib`).

Also, from the Freedesktop Wiki on Systemd:

>  Services using the network should hence simply place an After=network.target dependency in their unit files, and avoid any Wants=network.target or even Requires=network.target.

The example below is for a system-wide installation and includes some additional documentation and helpful comments:

```ini
[Unit]
Description=Matrix IRC Bridge
Documentation=https://github.com/matrix-org/matrix-appservice-irc
After=network.target

[Service]
User=matrix-appservice-irc
ExecStart=/usr/local/bin/matrix-appservice-irc -c /etc/matrix-appservice-irc/config.yaml -f /etc/matrix-appservice-irc/registration.yaml -p 8009

# The following loads the freebindfree library, which is required for individual IPv6 addresses.
# This needs to be installed separately, see the main configuration.
# Environment=LD_PRELOAD=/opt/freebindfree/libfreebindfree.so

# Allow more open file descriptors, useful for a large bridge
# LimitNOFILE=65536

[Install]
WantedBy=multi-user.target
```

Maybe some confinement is appropriate as well?

Perflyst:
> When running the bridge as a system service, I don't think it should run from a home directory 

Why?

silkeh:
> > When running the bridge as a system service, I don't think it should run from a home directory
> 
> Why?

Because system users and actual users should be kept separate, and  system-wide software should be installed properly to the `/usr(/local)` hierarchy, with state information stored in `/var/lib` and config in `/etc`.

Perflyst:
It is much easier to handle in different home directories. Also I would not install this appservice system wide, it is much better in a home.

Even synapse runs in its own home directory (and the official guide even says to do so).

silkeh:
> It is much easier to handle in different home directories. Also I would not install this appservice system wide, it is much better in a home.

The system file for this PR is explicitly for a system-wide install. A systemd user service would be used when running it in a user session. I'm not sure how an installation to anywhere else would be less easy to handle than an installation to `/home`, but you do you.

> Even synapse runs in its own home directory (and the official guide even says to do so).

I think the instructions (for installing synapse from source) install in *your* home directory, not it's own. Packaged versions of synapse most definitely don't install to a home directory.

Half-Shot:
Sorry, I think we let this rot a bit too long. On the whole I don't think we have our systemd story figured out yet. I think the format we'd likely accept is something like https://github.com/matrix-org/synapse/tree/develop/contrib/systemd (in that directory structure, documentation format).

The bridge team are currently a bit limited on what environments support too. We effectively support:
- Install via git checkout
- Install via npm
- Install via Docker

with third party support for things like Debian, Arch being provided by upstream maintainers. We'd typically expect them to maintain their own versions of a service file. We're happy to accept a generic example one, but I don't think this one is actually that generic.

Thanks for having a go though!

PR #528, opened 2017-12-04
Merge release 0.8.0 into the develop branch

14mRh4X0r:
This pulls in the dependency fix, and improves the output of `git describe`


14mRh4X0r:
The build failure cannot be due to this merge request, since it does not change code.

Half-Shot:
Fixed, via doing a new release from develop as it should be.

PR #529, opened 2017-12-04
Show a preview of long messages on IRC

14mRh4X0r:
This gives users on the IRC side an idea what the long message is about, without flooding the channel with text. This solves an issue some people have where IRC users [aren't willing to click to content they can't see](https://matrix.to/#/!SudviOJlimDvrGdFCY:matrix.org/$1512407915217926ULUZv:matrix.org), and is more [in line with other services](https://matrix.to/#/!SudviOJlimDvrGdFCY:matrix.org/$1512409886253094NDwNA:matrix.org).

### DIFF
--- a/lib/bridge/MatrixHandler.js
+++ b/lib/bridge/MatrixHandler.js
@@ -12,6 +12,7 @@ var MatrixUser = require("matrix-appservice-bridge").MatrixUser;
 var BridgeRequest = require("../models/BridgeRequest");
 var toIrcLowerCase = require("../irc/formatting").toIrcLowerCase;
 var StateLookup = require('matrix-appservice-bridge').StateLookup;
+var ContentRepo = require('matrix-appservice-bridge').ContentRepo;
 
 function MatrixHandler(ircBridge) {
     this.ircBridge = ircBridge;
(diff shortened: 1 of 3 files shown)


silkeh:
How does this handle pure code blocks?

Half-Shot:
This will need rebasing onto the current develop because we dropped node 4 :(.

Also, unit tests on the behavior would be good.

Half-Shot:
Thanking you, will review.

spec/util/irc-client-mock.js
```diff
@@ -133,6 +133,56 @@ function Client(addr, nick, opts) {
         return this.chans[channel];
     }
 
+    // Copied over from irc.js and modified slightly
```
Half-Shot: Not a huge fan of this, but it looks like the IRC lib for some reason contains code that really aught to be split out. I'll allow it.

Half-Shot: Actually, crazy idea. Could we put these two functions in a new class in util/MessageFormatting.js or something rather than using the unsafeClient version. I think it would be better than reimplementing them in a mock.

lib/bridge/MatrixHandler.js
```diff
@@ -1242,21 +1243,31 @@ MatrixHandler.prototype._sendIrcAction = Promise.coroutine(
...
+            };
+        } else {
+            event.content = {
+                msgtype: "m.text",
+                body: messagePreview + truncatedMsg
+            };
+        }
     }
     else {
         req.log.warn("Sending truncated message");
```
Half-Shot: Small complaint: This should probably be info rather than warn

Half-Shot:
I think this is looking very promising, if we fix these up I'll take another look and hopefully merge :)

Half-Shot:
Since our move to Typescript, this PR will no longer cleanly merge into the codebase. I'm going to close this rather than request changes. Feel free to re-open the PR if you want to continue to work on it.

tadzik:
Bringing this back in https://github.com/matrix-org/matrix-appservice-irc/pull/1430 since it's still an annoying issue for IRC users. Thanks for contributing it, @14mRh4X0r!

PR #537, opened 2018-01-06
fix bug where you can't kill users who are on no channels

ara4n:


lib/bridge/MatrixHandler.js
```diff
@@ -661,40 +661,40 @@ MatrixHandler.prototype.quitUser = Promise.coroutine(function*(req, userId, clie
                 `Bridged client for ${userId} is not in any channels ` +
                 `on ${bridgedClient.server.domain}`
             );
-            msg = "You are not in any channels on ${bridgedClient.server.domain}. ";
-            continue;
+            msg = `You are not in any channels on ${bridgedClient.server.domain}.`;
```
dbkr: You probably don't want to return this error message any more since it's now doing what you wanted it to do. This is the only thing that `msg` is used for so I think it can go away (it seems a bit broken anyway since if you were in no rooms on multiple networks, it will only tell you about one).

ara4n:
thanks :)

PR #544, opened 2018-02-06
Update to latest ESLint

aidalgol:
Depend upon latest ESLint, migrate our `.eslintrc` file, and fix new lint errors.


Half-Shot:
I wonder if this still works..

PR #545, opened 2018-02-06
Remove unused dependency

aidalgol:
jayschema is declared in package.json as a dependency, but is never used.


Half-Shot:
Good catch, validating the config was made part of appservice-bridge quite a while ago.

PR #549, opened 2018-02-07
Bump Node version to current LTS release

aidalgol:
Node 4 was two LTS releases ago now, and there's little value in supporting the previous LTS for this project.  (It's unlikely to be used in environments that are tied to very old versions of Node, and supporting more versions will demand more effort from an already stretched project.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: node_js
 node_js:
-  - "4"
+  - "8"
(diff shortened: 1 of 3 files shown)


Half-Shot:
@aidalgol Apologies, I didn't see this until it was too late. Thanks for trying to fix us up anyway :)

Issue #496, opened 2017-08-28
!whois is case-sensitive


Spydar007:
The `!whois` command is case-sensitive, and gives `Cannot find nick on whois` when the exact case doesn't exist. Given that all nicks and masks are considered case-insensitive, the matrix `!whois` command should also be.

aidalgol:
Is this a duplicate of #273?

Mikaela:
Yes.

PR #551, opened 2018-02-10
whois case-sensitivity fix

aidalgol:
Since the whois command is case insensitive, display the nick string from the
server's reply, instead of the string given by the user, so that we display the
nick with correct casing.

Goes with PR matrix-org/node-irc#26

Addresses matrix-org/matrix-appservice-irc#496


Half-Shot:
Since our move to Typescript, this PR will no longer cleanly merge into the codebase. I'm going to close this rather than request changes. Feel free to re-open the PR if you want to continue to work on it.

Issue #467, opened 2017-07-02
Why do nicks require beginning with a letter?


Spydar007:
[BridgedClient.js dictates that nicks must start with a letter](https://github.com/matrix-org/matrix-appservice-irc/blob/master/lib/irc/BridgedClient.js#L489). Is there a technical reason for this? It doesn't appear so, since commenting it out seems to work fine. This block means we cannot have nicks beginning with any of the special characters.

Any of the characters in the block above (dictated in RFC 2812 §2.3.1) can be used to begin a nick, except `-` and 0-9.

Kegsay:
Ah, the earlier RFC 1459 didn't include special characters, but I see that RFC 2812 does include special characters, so that can be changed. Can't start with `-` or numbers though as you state.

Kegsay:
There is no technical reason for it.

mweinelt:
I wouldn't say that this is a quick win per se. It requires handling of the 432 numeric, if the network does not support it.

```
NICK 4foo                                                                                                                                                                                                                                                               :guybrush.hackint.org 432 somenick 4foo :Erroneous Nickname
```

Kegsay:
We already handle that numeric when people issue `!nick`.

TheButterZone:
One of my grouped Freenode nicks starts with a {

"must start with a letter"

PR #552, opened 2018-02-11
Correct validation of nicks

aidalgol:
Correct the regular expression for validating the starting character of a nick
to conform to RFC 2812 § 2.3.1.  It was considering only letters to be valid
starting characters, whereas the standard allows letters or special characters.

Fixes matrix-org/matrix-appservice-irc#467

### DIFF
--- a/lib/irc/BridgedClient.js
+++ b/lib/irc/BridgedClient.js
@@ -495,8 +495,8 @@ BridgedClient.prototype._getValidNick = function(nick, throwOnInvalid) {
         throw new Error(`Nick '${nick}' contains illegal characters.`);
     }
 
-    // nicks must start with a letter
-    if (!/^[A-Za-z]/.test(n)) {
+    // nicks must start with a letter or special character
+    if (!/^[A-Za-z\]\[\^\\\{\}\-`_\|]/.test(n)) {
         if (throwOnInvalid) {
             throw new Error(`Nick '${nick}' must start with a letter.`);
         }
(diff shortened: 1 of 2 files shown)


Half-Shot:
Hiya, thanks for contributing the PR :). Could you write an appropriate test file and case in https://github.com/matrix-org/matrix-appservice-irc/tree/master/spec/unit please just so we can make sure the behaviour is tested 👼

aidalgol:
I studied the existing specs and started to write one for `BridgedClient` following their form, and I came up against the `BridgedClient` constructor requiring an `IrcServer`, an `IrcClientConfig`, an `IrcEventBroker`, an `IdentGenerator`, and an `Ipv6Generator`.  I see there is already an example of mocking `IrcClientConfig` in the `IdentGenerator` spec, but I'm having trouble figuring out what would be the bare minimum required for mocking the other classes.

0x4A6F:
@aidalgol, thanks for your work.
Any progress on this?

aidalgol:
@0x4A6F No, I never got anywhere with writing tests for this.  Still keen to see this through, but I will need some guidance.

dalcde:
Is this still being worked on? I would be happy to take it up. For the tests, it is probably easier to turn `getValidNick` into a static method, which it really ought to be imo, so that we don't have to construct a mock `BridgeClient` instance.

Half-Shot:
@dalcde You're welcome to work on it, but since we moved to Typescript it might be better to start over.

Half-Shot:
Thanks for kicking this change off @aidalgol. We've done the equivalent TS work in #1114 so closing this.

PR #554, opened 2018-02-13
Set is_direct flag when doing a PM I->M

lukebarnard1:
Fixes https://github.com/vector-im/riot-web/issues/6088


lukebarnard1:
The test failure looks like flakiness unfortunately :/

lukebarnard1:
Restarting the test has resulted in it passing, so it is indeed flaky.

ara4n:
lgtm - thanks

PR #587, opened 2018-06-06
Update winston to 2.4.2

### LINKED ISSUE #578: Bridge crashes under high logging load
The bridge uses a logging library called winston
winston.logger.log calls transport.log with a callback
the transport is DailyRotateFile. DailyRotateFile.log is backed by an fs.WriteStream and implements the callback with _stream.once
fs.writestream is backed by an EventListener, which gets expensive when there are lots of listeners
so if we log lots of stuff at once, we end up spending all our time adding and removing callbacks to the log stream, and the problem keeps getting worse

Unclear if this is our fault for doing too much logging, winston's fault for using eventlistener.once, or eventlistener.once's fault for being generally inefficient

### LINKED ISSUE #579: Winston logger is very out of date
Current Winston is pinned at 0.9.0 which was released 3 years ago, and I am reluctant to start snooping into logging issues knowing full well it's likely been improved in a later release. 

Upgrading will require making a few modifications to the logger setup, but the log statements can be left intact. I'd be curious to see if this resolves some of the issues in #578.

Half-Shot:
Fixes #579 and potentially fixes #578 (pending testing). Due to the investigation in #579 it would make sense to upgrade the library version.

### DIFF
--- a/lib/logging.js
+++ b/lib/logging.js
@@ -4,6 +4,7 @@
 "use strict";
 
 var winston = require("winston");
+require('winston-daily-rotate-file');
 
 var loggerConfig = {
     level: "debug", //debug|info|warn|error
--- a/lib/logging.js
+++ b/lib/logging.js
@@ -52,7 +52,7 @@ var makeTransports = function() {
             timestamp: timestampFn,
             formatter: formatterFn,
             maxFiles: loggerConfig.maxFiles,
-            datePattern: ".yyyy-MM-dd",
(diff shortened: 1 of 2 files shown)


Half-Shot:
Due to the entirely reasonable request of ``winston-daily-log-rotate`` requiring node 6+, I'll move the versions back slightly.

Half-Shot:
And since we were fine to move to node 6+, we can now unpin \o/

PR #590, opened 2018-06-09
Make ident reliable by delaying negative response

heftig:
The IRC server typically sends the ident request as soon as the TCP
connection is established. However, since we only add the ident mapping
after we get the established connection socket, the response is
extremely racy and often fails to find the mapping.

This commit follows [the scheme implemented by Quassel][1] and delays
giving a negative response while there are still connections in the
process of being opened and are older than the request. This should make
the response reliable.

[1]: https://github.com/quassel/quassel/pull/356

### DIFF
--- a/src/irc/BridgedClient.js
+++ b/src/irc/BridgedClient.js
@@ -127,6 +127,8 @@ BridgedClient.prototype.connect = Promise.coroutine(function*() {
             `Connecting to the IRC network '${this.server.domain}' as ${this.nick}...`
         );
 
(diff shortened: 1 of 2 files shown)


Half-Shot:
Thanks for the PR! This seems quite reasonable on the face of it so I will take a look sometime today.

heftig:
Any progress? This is still mergable.

heftig:
Review, please?

heftig:
I've rebased it to resolve the conflicts. Review, please?

PR #591, opened 2018-06-09
Allow configuring channel keys for static mappings

heftig:
In use by the archlinux.org bridge for quite some time now.

### DIFF
--- a/HOWTO.md
+++ b/HOWTO.md
@@ -60,7 +60,8 @@ ircService:
     # This is the IRC server url to connect to.
     irc.example.com:
       mappings:
-        "#some-channel": ["!someroomid:here"]
+        "#some-channel":
+          roomIds: ["!someroomid:here"]
 ```
 This would set up a simple mapping from ``#some-channel`` on 
 ``irc.example.com`` to ``!someroomid:here``, and that's it. Dynamic mappings
--- a/changelog.d/591.feature
+++ b/changelog.d/591.feature
@@ -1,0 +1,19 @@
+**Breaking Change** - Static mappings can now set a channel key:
+ - This changes the config schema, even if you do not make use of this feature. You MUST update your existing `mappings` to use the new `roomIds`:
+   ```yaml
+   old:
+     mappings:
+       "#thepub": ["!kieouiJuedJoxtVdaG:localhost"]
+
+   new:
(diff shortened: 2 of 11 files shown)


Half-Shot:
Thanks for the PR.

Couple of things:

* When updating the config, please also update the schema and the sample config to match.
* I suspect it would be nice to keep mappings and keys inside the same object, is there no way we could reuse ``mappings`` for storing the keys.

heftig:
>is there no way we could reuse mappings for storing the keys.

`mappings` is a `"#channel"`→ `"!roomid:matrix.org"` map so there's no way to add a channel key unless you propose somehow putting it into one of the strings, which would be weird.

Half-Shot:
I'm proposing making ``mappings`` a key->object. ``"#channel" -> {roomid, key}``

heftig:
@Half-Shot Is this better?

Half-Shot:
Looks good, final thing is to add a changelog as per https://github.com/matrix-org/matrix-appservice-irc/blob/master/CONTRIBUTING.md#making-pull-requests. Ensure you highlight the change in schema prominently.

heftig:
Is this supposed to be a one-liner or a multi-paragraph explanation?

Half-Shot:
> Is this supposed to be a one-liner or a multi-paragraph explanation?

I recommend you put migration instructions for the config change into the changelog (probably a paragraph or so).

heftig:
I added the following fragment:
<blockquote>

**Breaking Change** - Static mappings can now set a channel key:
 - This changes the config schema, even if you do not make use of this feature. You MUST update your existing `mappings` to use the new `roomIds`:
   ```yaml
   old:
     mappings:
       "#thepub": ["!kieouiJuedJoxtVdaG:localhost"]

   new:
     mappings:
       "#thepub":
         roomIds: ["!kieouiJuedJoxtVdaG:localhost"]
   ```
 - The key is automatically used to join Matrix users to the mapped channel. They do not need to know the key. For example, you can bridge password-protected IRC channels to invite-only Matrix rooms:
   ```yaml
   mappings:
     "#viplounge":
       roomIds: ["!xKtieojhareASOokdc:localhost"]
       key: "vip-pass"
   ```

</blockquote>

PR #594, opened 2018-06-14
Check we are joined on mapped rooms before trying to join

### LINKED ISSUE #593: We dimly try to rejoin every static mapped room as the bot user on startup
Probably because we don't know any better, however we could look at getJoinedRooms and skip anything we don't feel the need to rejoin. Given matrix.org is currently on fire while trying to join all da rooms.

Half-Shot:
Because synapse will attempt to join regardless of the actual join state, and one call to  ``joined_rooms`` is faster than x number of ``join`` calls.

This fixes #593

### DIFF
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -807,7 +807,7 @@ IrcBridge.prototype._loginToServer = Promise.coroutine(function*(server) {
             });
         }, 500 * num);
         num += 1;
-    });yield
+    });
 });
 
 IrcBridge.prototype.checkNickExists = function(server, nick) {
@@ -905,8 +905,13 @@ IrcBridge.prototype.getBotClient = function(server) {
         return this._clientPool.getBot(server);
     });
 }
-
 IrcBridge.prototype._fetchJoinedRooms = Promise.coroutine(function*() {
+    /** Fetching joined rooms is quicker on larger homeservers than trying to
+     * /join each room in the mappings list. To ensure we start quicker,
+     * the bridge will block on this call rather than blocking on all join calls.
+     * On the most overloaded servers even this call may take several attempts,
+     * so it will block indefinitely.
+     */
     const bot = this._bridge.getBot();
     if (bot.getJoinedRooms == undefined) {
         return;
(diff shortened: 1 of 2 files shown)


Half-Shot:
Tests are failing because AppServiceBot.getJoinedRooms times out

lib/bridge/IrcBridge.js
```diff
@@ -798,7 +807,7 @@ IrcBridge.prototype._loginToServer = Promise.coroutine(function*(server) {
             });
         }, 500 * num);
         num += 1;
-    });
+    });yield
```
lukebarnard1: This looks.. unnecessary.

lib/bridge/IrcBridge.js
```diff
@@ -897,6 +906,26 @@ IrcBridge.prototype.getBotClient = function(server) {
     });
 }
 
+IrcBridge.prototype._fetchJoinedRooms = Promise.coroutine(function*() {
+    const bot = this._bridge.getBot();
+    let gotRooms = false;
+    while (!gotRooms) {
+        try {
+            const roomids = yield bot.getJoinedRooms();
```
lukebarnard1: Probably want to stick to camelCase here, e.g. `roomIds`

lib/bridge/IrcBridge.js
```diff
@@ -897,6 +906,26 @@ IrcBridge.prototype.getBotClient = function(server) {
...
+            gotRooms = true;
+            this.joinedRoomList = roomids;
+            log.info(`ASBot is in ${roomids.length} rooms!`);
+        }
+        catch (ex) {
+            // We should retry this, because it's still quicker than rejoining everything.
```
lukebarnard1: *Quicker to fail and succeed once you mean? The caveat with this is that it could block the bridge from ever starting.

What would be the typical circumstances for failing here and does it make sense to retry? Or block the bridge starting on retrying? (If it fails to request the joined rooms then that implies it won't be able to do much else.)

Half-Shot:
(Moving discussion outside because GH has a tendency to hide comments)
@lukebarnard1 : 

>*Quicker to fail and succeed once you mean? The caveat with this is that it could block the bridge from ever starting.

>What would be the typical circumstances for failing here and does it make sense to retry? Or block the bridge starting on retrying? (If it fails to request the joined rooms then that implies it won't be able to do much else.)

On startup of the Freenode bridge on matrix.org, we try to join every room we have mapped (statically in the file, not provisioned rooms) to ensure we are in it. The bridge doesn't know which rooms it's in by default so we will block either on joining those rooms, or fetching a list of rooms. For some reason even trying to join a room we are in will block forever on mx.org so I figured it would be better to just get the room list once and then we can skip the joins.

lukebarnard1:
Ah ok. Thanks for explanation and comments! :tada: :+1:

PR #600, opened 2018-06-26
Announce and leave DM rooms if not supported

Half-Shot:


lib/bridge/MatrixHandler.js
```diff
@@ -13,6 +13,9 @@ var BridgeRequest = require("../models/BridgeRequest");
 var toIrcLowerCase = require("../irc/formatting").toIrcLowerCase;
 var StateLookup = require('matrix-appservice-bridge').StateLookup;
 
+const MSG_PMS_DISABLED = "[Bridge] Sorry, DMs are disabled on this bridge.";
```
anoadragon453: Seems like the codebase is calling these PMs instead of DMs?

lib/bridge/MatrixHandler.js
```diff
@@ -83,10 +86,17 @@ MatrixHandler.prototype._handleInviteFromUser = Promise.coroutine(function*(req,
                                                                   invitedIrcUser) {
     req.log.info("Handling invite from user directed at %s on %s",
         invitedIrcUser.server.domain, invitedIrcUser.nick);
+    const intent = this.ircBridge.getAppServiceBridge().getIntent(invitedUser.getId());
```
anoadragon453: `invitedUser` -> `invitedIrcUser`

Half-Shot:
>Is the intention that every bridge requires this functionality? Or should it be configurable?

It's one I've been mulling over and I'm really of the opinion we should be trying to spec a way for bridges to declare this programatically rather than joining a room, telling the client to shove off and leave.

The obvious solution would be profiles that specify whether they can accept PMs so Riot can be proactive about it...but that would mean profile support.

PR #604, opened 2018-07-05
Limit concurrent reconnects

Half-Shot:
This is designed so that we do not try to reconnect all clients at once in the case of a catastrophic connection loss. There are a few issues with this implementation as it stands but I want to see what people think of this first:

- We are now limiting reconnects, which means recovery will take longer (though potentially less crashy)
- ~~If we receive a "net_error" i.e. a socket issue, we throttle for **20s** which may be unacceptably long.~~
- The limit is currently 50 connections a second (which is now configurable), so if every connection takes one second, and we are dealing with a mass disconnect from freenode: 22000 connections / 20 = 1100s = 18m

Given this could very well make or break some of the matrix.org hosted networks, I'd like to get reviews from a few folks :)

### DIFF
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -295,6 +295,14 @@ ircService:
         # a FIFO queue.
         # Default: 5000 (5 seconds)
         reconnectIntervalMs: 5000
+        # The number of concurrent reconnects if a user has been disconnected unexpectedly
+        # (e.g. a netsplit). You should set this to a reasonably high number so that
+        # bridges are not waiting an eternity to reconnect all it's users if 
+        # we see a massive number of disconnect. This is unrelated to the reconnectIntervalMs
+        # setting above which is for connecting on restart of the bridge. Set to 0 to 
+        # immediately try to reconnect all users.
+        # Default: 50
+        concurrentReconnectLimit: 50
         # The number of lines to allow being sent by the IRC client that has received
         # a large block of text to send from matrix. If the number of lines that would
         # be sent is > lineLimit, the text will instead be uploaded to matrix and the
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -297,7 +297,7 @@ ircService:
         reconnectIntervalMs: 5000
         # The number of concurrent reconnects if a user has been disconnected unexpectedly
         # (e.g. a netsplit). You should set this to a reasonably high number so that
-        # bridges are not waiting an eternity to reconnect all it's users if 
+        # bridges are not waiting an eternity to reconnect all its clients if 
         # we see a massive number of disconnect. This is unrelated to the reconnectIntervalMs
         # setting above which is for connecting on restart of the bridge. Set to 0 to 
         # immediately try to reconnect all users.
--- a/lib/irc/BridgedClient.js
+++ b/lib/irc/BridgedClient.js
@@ -142,7 +142,6 @@ BridgedClient.prototype.connect = Promise.coroutine(function*() {
 
         this.inst = connInst;
         this.unsafeClient = connInst.client;
-        this.log.debug("connected!");
         this.emit("client-connected", this);
         // we may have been assigned a different nick, so update it from source
         this.nick = connInst.client.nick;
--- a/lib/irc/ClientPool.js
+++ b/lib/irc/ClientPool.js
@@ -4,11 +4,12 @@
  * and may be closed for a variety of reasons.
  */
 "use strict";
-var stats = require("../config/stats");
-var log = require("../logging").get("ClientPool");
-var Promise = require("bluebird");
+const stats = require("../config/stats");
+const log = require("../logging").get("ClientPool");
+const Promise = require("bluebird");
+const QueuePool = require("../util/QueuePool");
 
-const RECONNECT_TIME_MS = 1000;
+const RECONNECT_CLIENT_QUEUE_SIZE = 20;
 
 function ClientPool(ircBridge) {
     this._ircBridge = ircBridge;
@@ -35,6 +36,13 @@ function ClientPool(ircBridge) {
     this._virtualClientCounts = {
         // server_domain: number
     };
+    this._reconnectQueue = new QueuePool( 
+        RECONNECT_CLIENT_QUEUE_SIZE,
+        (item) => {
+            log.info(`Reconnecting client. ${this._reconnectQueue.waitingItems} left.`);
+            return this._reconnectClient(item);
+        }
+    );
 }
 
 ClientPool.prototype.killAllClients = function() {
@@ -276,25 +284,27 @@ ClientPool.prototype._onClientDisconnected = function(bridgedClient) {
     // remove ref to the disconnected client so it can be GC'd. If we don't do this,
     // the timeout below holds it in a closure, preventing it from being GC'd.
     bridgedClient = undefined;
-    setTimeout(function() {
-        cli.connect().then(function() {
-            log.info(
-                "<%s> Reconnected %s@%s", cli._id, cli.nick, cli.server.domain
-            );
-            if (chanList.length > 0) {
-                log.info("<%s> Rejoining %s channels", cli._id, chanList.length);
-                chanList.forEach(function(c) {
-                    cli.joinChannel(c);
-                });
-            }
-            self._sendConnectionMetric(cli.server);
-        }, function(e) {
-            log.error(
-                "<%s> Failed to reconnect %s@%s", cli._id, cli.nick, cli.server.domain
-            );
-        });
-    }, RECONNECT_TIME_MS);
+    this._reconnectQueue.enqueue(cli._id, {cli, chanList});
 };
+
+ClientPool.prototype._reconnectClient = function({cli, chanList}) {
+    return cli.connect().then(() => {
+        log.info(
+            "<%s> Reconnected %s@%s", cli._id, cli.nick, cli.server.domain
+        );
+        if (chanList.length > 0) {
+            log.info("<%s> Rejoining %s channels", cli._id, chanList.length);
+            chanList.forEach(function(c) {
+                cli.joinChannel(c);
+            });
+        }
+        this._sendConnectionMetric(cli.server);
+    }, function(e) {
+        log.error(
+            "<%s> Failed to reconnect %s@%s", cli._id, cli.nick, cli.server.domain
+        );
+    });
+}
 
 ClientPool.prototype._onNickChange = function(bridgedClient, oldNick, newNick) {
     this._virtualClients[bridgedClient.server.domain].nicks[oldNick] = undefined;
--- a/lib/irc/ClientPool.js
+++ b/lib/irc/ClientPool.js
@@ -77,7 +77,7 @@ ClientPool.prototype.getOrCreateReconnQueue = function(server) {
         q = this._reconnectQueues[server.domain] = new QueuePool(
             server.getConcurrentReconnectLimit(),
             (item) => {
-                log.info(`Reconnecting client. ${this._reconnectQueue.waitingItems} left.`);
+                log.info(`Reconnecting client. ${q.waitingItems} left.`);
                 return this._reconnectClient(item);
             }
         );
@@ -308,7 +308,6 @@ ClientPool.prototype._onClientDisconnected = function(bridgedClient) {
         });
         return;
     }
-
     queue.enqueue(cli._id, {
         cli: cli,
         chanList: chanList,
--- a/lib/irc/ClientPool.js
+++ b/lib/irc/ClientPool.js
@@ -68,7 +68,7 @@ ClientPool.prototype.killAllClients = function() {
     );
 }
 
-ClientPool.prototype.getOrCreateReconnQueue = function(server) {
+ClientPool.prototype.getOrCreateReconnectQueue = function(server) {
     if (server.getConcurrentReconnectLimit() === 0) {
         return null;
     }
@@ -300,7 +300,7 @@ ClientPool.prototype._onClientDisconnected = function(bridgedClient) {
         log.info(`Dropping ${cli._id} ${cli.nick} because they are not joined to any channels`);
         return;
     }
-    let queue = this.getOrCreateReconnQueue(cli.server);
+    let queue = this.getOrCreateReconnectQueue(cli.server);
     if (queue === null) {
         this._reconnectClient({
             cli: cli,
--- a/lib/irc/ClientPool.js
+++ b/lib/irc/ClientPool.js
@@ -284,7 +284,7 @@ ClientPool.prototype._onClientDisconnected = function(bridgedClient) {
     // the timeout below holds it in a closure, preventing it from being GC'd.
     bridgedClient = undefined;
 
-    if(chanList.length === 0) {
+    if (chanList.length === 0) {
         log.info(`Dropping ${cli._id} ${cli.nick} because they are not joined to any channels`);
         return;
     }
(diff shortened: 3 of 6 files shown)


lib/irc/ConnectionInstance.js
```diff
@@ -355,9 +355,16 @@ ConnectionInstance.create = Promise.coroutine(function*(server, opts, onCreatedC
             log.error(
                 `ConnectionInstance.connect failed after ${connAttempts} attempts (${err.message})`
             );
+
+            if (err.message === "net_error") {
```
anoadragon453: This is a strange way of backing off. Should this be an `if (... && err.message === "net_error")` with the if statement below?

lib/util/QueuePool.js
```diff
@@ -24,8 +24,12 @@ class QueuePool {
             this.queues.push(new Queue(fn));
         }
         this.overflow = new Queue(this._overflow.bind(this));
+        this._overflowCount = 0;
     }
 
+    get waitingItems() { return this._overflowCount; };
+    
+
```
anoadragon453: Two newlines?? You absolute madman.

anoadragon453:
Looks alright overall. If that count is just used for logging then you shouldn't need to do too many check on it.

lib/util/QueuePool.js
Half-Shot: You gotta give functions their space :)

lib/util/QueuePool.js
```diff
@@ -28,8 +28,8 @@ class QueuePool {
     }
 
     // Get number of items waiting to be inserted into a queue.
-    get waitingItems() { return this._overflowCount; };
-    
+    get waitingItems() { return this._overflowCount; }
```
anoadragon453: forgot the `;`

lib/irc/ClientPool.js
```diff
@@ -280,14 +280,18 @@ ClientPool.prototype._onClientDisconnected = function(bridgedClient) {
...
+    this._reconnectQueue.enqueue(cli._id, {
+        cli: cli,
+        chanList: chanList,
+    });
 };
 
-ClientPool.prototype._reconnectClient = function({cli, chanList}) {
+ClientPool.prototype._reconnectClient = function(cliChan) {
+    const cli = cliChan.cli;
+    const chanList = cliChan.chanList;
```
anoadragon453: What's the point of this?

lib/util/QueuePool.js
Half-Shot: It's a class, it doesn't need one :)

lib/irc/ClientPool.js
Half-Shot: eslint 1 doesn't understand the syntax and I don't want to bump it and break more things in *this* PR.

anoadragon453: okie dokie

lib/util/QueuePool.js
```diff
@@ -24,8 +24,12 @@ class QueuePool {
             this.queues.push(new Queue(fn));
         }
         this.overflow = new Queue(this._overflow.bind(this));
+        this._overflowCount = 0;
     }
 
+    get waitingItems() { return this._overflowCount; };
```
lukebarnard1: This is unused within the context of this commit, it's generally clearer to include usage of a thing with its creation within the same commit. (But don't worry about it this time around).

lib/irc/BridgedClient.js
```diff
@@ -142,7 +142,6 @@ BridgedClient.prototype.connect = Promise.coroutine(function*() {
 
         this.inst = connInst;
         this.unsafeClient = connInst.client;
-        this.log.debug("connected!");
```
lukebarnard1: Well, it's only logged for `debug`. Why remove it exactly?

lib/irc/ConnectionInstance.js
```diff
@@ -355,9 +355,16 @@ ConnectionInstance.create = Promise.coroutine(function*(server, opts, onCreatedC
...
+                // This is the kernel complaining, we should back off.
```
lukebarnard1: Why is the kernel complaining? Why do we back off?

lib/irc/ClientPool.js
```diff
@@ -261,10 +261,6 @@ ClientPool.prototype._onClientDisconnected = function(bridgedClient) {
         return;
     }
     // Reconnect this user
-    log.debug(
-        "onClientDisconnected: <%s> Reconnecting %s@%s in %sms",
-        bridgedClient._id, bridgedClient.nick, bridgedClient.server.domain, RECONNECT_TIME_MS
-    );
```
lukebarnard1: Is this not a useful thing to have?

lib/irc/ClientPool.js
```diff
@@ -4,11 +4,12 @@
  * and may be closed for a variety of reasons.
  */
 "use strict";
-var stats = require("../config/stats");
-var log = require("../logging").get("ClientPool");
-var Promise = require("bluebird");
+const stats = require("../config/stats");
+const log = require("../logging").get("ClientPool");
+const Promise = require("bluebird");
+const QueuePool = require("../util/QueuePool");
```
lukebarnard1: @Kegsay will be pleased that the legend of QueuePool lives on

lib/irc/ClientPool.js
```diff
@@ -276,26 +284,28 @@ ClientPool.prototype._onClientDisconnected = function(bridgedClient) {
...
-    }, RECONNECT_TIME_MS);
+    this._reconnectQueue.enqueue(cli._id, {cli, chanList});
 };
 
+ClientPool.prototype._reconnectClient = function({cli, chanList}) {
+    return cli.connect().then(() => {
+        log.info(
+            "<%s> Reconnected %s@%s", cli._id, cli.nick, cli.server.domain
+        );
+        if (chanList.length > 0) {
```
lukebarnard1: If the `chanList` is empty, perhaps we could bail out before even adding this to the QueuePool?

lib/irc/ClientPool.js
```diff
@@ -4,11 +4,12 @@
...
-var stats = require("../config/stats");
-var log = require("../logging").get("ClientPool");
-var Promise = require("bluebird");
+const stats = require("../config/stats");
+const log = require("../logging").get("ClientPool");
+const Promise = require("bluebird");
+const QueuePool = require("../util/QueuePool");
 
-const RECONNECT_TIME_MS = 1000;
+const RECONNECT_CLIENT_QUEUE_SIZE = 20;
```
lukebarnard1: This means we are limited to 20 concurrent ongoing reconnections, not a queue length of 20. Maybe `QUEUE_POOL_SIZE` would better describe this.

lib/irc/ConnectionInstance.js
```diff
@@ -18,6 +18,10 @@ const PING_TIMEOUT_MS = 1000 * 60 * 10;
 // due to throttling.
 const THROTTLE_WAIT_MS = 20 * 1000;
 
+// The minimum time to wait between connection attempts if the kernel has given us
+// a net_error.
+const NET_ERROR_WAIT_MS = 5 * 1000;
```
lukebarnard1: This suggests we are likely to have success if we retry sooner than if we are throttled. It should be explained in the comments why we retry on net error, and why it is done sooner than if we are throttled (for example).

lib/irc/ClientPool.js
```diff
@@ -9,7 +9,7 @@ const log = require("../logging").get("ClientPool");
 const Promise = require("bluebird");
 const QueuePool = require("../util/QueuePool");
 
-const RECONNECT_CLIENT_QUEUE_SIZE = 20;
+const RECONNECT_CLIENT_QUEUE_SIZE = 50;
```
lukebarnard1: But why?

lib/irc/ClientPool.js
lukebarnard1: ouch.. eslint 1??

lib/irc/ConnectionInstance.js
Half-Shot: This is more of a thing I found when trying to run enormous amounts of IRC connections from my box:
```
{
  "errno": "EBUSY",
  "code": "EBUSY",
  "syscall": "getaddrinfo",
  "hostname": "localhost",
  "host": "localhost",
  "port": 6767
}
```

lib/irc/ClientPool.js
Half-Shot: The problem with it is that if we lose connection on a large box we spam this once per client connection, which is 22k times on freenode which tends to be deadly.

lib/irc/ClientPool.js
Half-Shot: Good idea :thinking:

lib/irc/BridgedClient.js
Half-Shot: We already log "Connecting" and if it were to fail, we'd see an error. Plus, it's not actually logging anything about *which* connection is connected so it's not much use on it's own.

lukebarnard1: Fair enough!

lib/irc/ConnectionInstance.js
lukebarnard1: Right, so the host is presumably running out of connections to give? Backing off only makes sense if expect the resource to be available in the future (at this point, it's `EBUSY`). I'm not sure we can expect the server to have space for us in 20s, especially if it's your devbox and the only thing consuming the resource *is* the IRC bridge.

lib/irc/ClientPool.js
lukebarnard1: Fair. So long as there is an indication that 22k clients are reconnecting.

lib/irc/ClientPool.js
```diff
@@ -284,7 +284,7 @@ ClientPool.prototype._onClientDisconnected = function(bridgedClient) {
     // the timeout below holds it in a closure, preventing it from being GC'd.
     bridgedClient = undefined;
 
-    if(chanList.length === 0) {
+    if (chanList.length === 0) {
```
lukebarnard1: Boo, unrelated to the commit.

lib/irc/ConnectionInstance.js
```diff
@@ -19,7 +19,9 @@ const PING_TIMEOUT_MS = 1000 * 60 * 10;
 const THROTTLE_WAIT_MS = 20 * 1000;
 
 // The minimum time to wait between connection attempts if the kernel has given us
-// a net_error.
+// a net_error. Usually these come about due to a network being unavaliable or resource
+// starvation. Unlike throttling which is ircd based, we should retry sooner to see if
+// the kernel is happy now.
```
lukebarnard1: But why? :innocent:

Half-Shot: Can I not just say "please consult your kernel's manual"?

lib/irc/ClientPool.js
Half-Shot: I know :disappointed: . It was just a space :s

lib/irc/ConnectionInstance.js
Half-Shot: I'm not sure what else to do, we can't really exit this func without a connected client but equally we can't harass the kernel for a slot if it's saying no.

lib/irc/ClientPool.js
Half-Shot: The hope is the queue func that logs waiting items above is good enough of an indicator.

Half-Shot: (I'm aware it will also spam a lot, but hopefully not quite as quickly so as to run out of resources)

lib/irc/ClientPool.js
lukebarnard1: No probs. `git add -p` is your friend.

lib/irc/ConnectionInstance.js
lukebarnard1: If we can justify backing off more than the base ~2s, then let's do it. Otherwise, I'm not sure there's a problem.

lib/irc/ClientPool.js
Half-Shot: ``CONCURRENT_RECONNECT_LIMIT`` sounds okay?

Half-Shot:
Looks good to you?

lib/util/QueuePool.js
```diff
@@ -68,11 +72,13 @@ class QueuePool {
                 p: queue.enqueue(req.id, req.item)
             });
         }
+        this._overflowCount++;
```
Kegsay: Is this variable meant to be counting the size of the overflow queue (aka "number of items waiting to be inserted into a queue.")? If so, why isn't this incremented when `this.overflow.enqueue` is called? At present, this will only be incremented "when a request is at the front of the overflow queue", which isn't the same thing.

lib/irc/ClientPool.js
```diff
@@ -4,11 +4,12 @@
...
-var stats = require("../config/stats");
-var log = require("../logging").get("ClientPool");
-var Promise = require("bluebird");
+const stats = require("../config/stats");
+const log = require("../logging").get("ClientPool");
+const Promise = require("bluebird");
+const QueuePool = require("../util/QueuePool");
 
-const RECONNECT_TIME_MS = 1000;
+const CONCURRENT_RECONNECT_LIMIT = 50;
```
Kegsay: You almost certainly want this to be configurable in the YAML file, alongside `reconnectIntervalMs`.

Kegsay:
> This is designed so that we do not try to reconnect all clients at once in the case of a catastrophic connection loss.

The intention is that the bridge will only reconnect all clients at once if [this config value](https://github.com/matrix-org/matrix-appservice-irc/blob/master/config.sample.yaml#L297) is 0. Otherwise it will wait at least the amount of ms specified between connections, but it'll be limited to 1 concurrent request.

> We are now limiting reconnects, which means recovery will take longer (though potentially less crashy)

Assuming a non-zero reconnect interval, I would expect this PR to *speed up* recovery time and it will allow >1 concurrent connection attempt. YMMV

config.sample.yaml
```diff
@@ -295,6 +295,14 @@ ircService:
         # a FIFO queue.
         # Default: 5000 (5 seconds)
         reconnectIntervalMs: 5000
+        # The number of concurrent reconnects if a user has been disconnected unexpectedly
+        # (e.g. a netsplit). You should set this to a reasonably high number so that
+        # bridges are not waiting an eternity to reconnect all it's users if
```
anoadragon453: its :>

lib/irc/ClientPool.js
```diff
@@ -276,30 +292,47 @@ ClientPool.prototype._onClientDisconnected = function(bridgedClient) {
...
+    return cli.connect().then(() => {
+        log.info(
+            "<%s> Reconnected %s@%s", cli._id, cli.nick, cli.server.domain
+        );
+        log.info("<%s> Rejoining %s channels", cli._id, chanList.length);
+        chanList.forEach(function(c) {
+            cli.joinChannel(c);
+        });
+        this._sendConnectionMetric(cli.server);
+    }, function(e) {
```
turt2live: why not use an arrow function here too?

lib/irc/ConnectionInstance.js
```diff
@@ -355,9 +355,11 @@ ConnectionInstance.create = Promise.coroutine(function*(server, opts, onCreatedC
             log.error(
                 `ConnectionInstance.connect failed after ${connAttempts} attempts (${err.message})`
             );
+
             if (err.message === "throttled") {
                 retryTimeMs += THROTTLE_WAIT_MS;
             }
+
```
turt2live: These 2 blank lines are the only changes in the file

Half-Shot:
That's enough ticks for me, merging!

PR #606, opened 2018-07-08
Removing end space when changing names by migrate users script

sash13:
After removing '(IRC)' in 'test (IRC)' we get 'test '. 
So remove last space symbol.


Half-Shot:
LGTM, thanks! :+1:

PR #610, opened 2018-07-10
Don't rejoin users we know are in the room on a NAMES request.

Half-Shot:
This is a dumb lookup of the list of ``/joinedMembers`` for rooms to determine if we have this person joined. If so, we just carry on. This doesn't affect ``join`` only ``names``, so any massive flash of joins will still need to be worked through but this at least counters the issue on restart of bridges. 

Flaws:
* We sync matrix membership once on start, but if we encounter more *NAMES* calls then we do not try to resync membership so the information becomes out of date over the bridges lifetime.


Half-Shot:
Yes, but unsure if that matters if we only get one "NAMES" update.

PR #611, opened 2018-07-10
Unbridge API

Half-Shot:
An endpoint on the DebugAPI to unbridge a portal room (alias). You can call ``/killPortal`` with the body:
```
{
    room_id: // room_id to kill
    domain // Domain of the irc server
    channel: // Channel of the irc chan
    leave_notice: true // Do you want to put a notice in the room. Def true.
    remove_alias: true // Remove the alias of the room assigned by the bridge. Def true.
}
```

Kill was chosen so to be similar to `/killClient` but I might change that one.

### DIFF
--- a/lib/DebugApi.js
+++ b/lib/DebugApi.js
@@ -209,14 +209,19 @@ DebugApi.prototype.run = function() {
 DebugApi.prototype.killPortal = Promise.coroutine(function*(req, response) {
     const result = {error: [], stages: []};
     const body = yield this._wrapJsonReq(req, response);
+
     // Room room_id to lookup and delete the alias from.
     const roomId = body["room_id"];
+
     // IRC server domain
     const domain = body["domain"];
+
     // IRC channel
     const channel = body["channel"];
+
     // Should we tell the room about the deletion. Defaults to true.
     const notice = !(body["leave_notice"] === false);
+
     // Should we remove the alias from the room. Defaults to true.
     const remove_alias = !(body["remove_alias"] === false);
 
--- a/lib/DebugApi.js
+++ b/lib/DebugApi.js
@@ -207,7 +207,11 @@ DebugApi.prototype.run = function() {
 }
 
 DebugApi.prototype.killPortal = Promise.coroutine(function*(req, response) {
-    const result = {error: [], stages: []};
+    const result = {
+        error: [], // string|[string] containing a fatal error or minor errors.
+        stages: [] // stages completed for removing the room. It's possible it might only
+                   // half complete, and we should make that obvious.
+    };
     const body = yield this._wrapJsonReq(req, response);
 
     // Room room_id to lookup and delete the alias from.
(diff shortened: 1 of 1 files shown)


anoadragon453:
Is the DebugAPI something on Synapse, or the IRC bridge?

Half-Shot:
@anoadragon453 Specific to this bridge to basically allow administrators to fix things while it's running.

anoadragon453:
Technically this could unlink one room and send a message in the other. We don't seem to check if they're linked. Is that fine?

lib/DebugApi.js
```diff
@@ -201,4 +206,143 @@ DebugApi.prototype.run = function() {
...
+            result.error.push("Failed to remove alias");
+        }
+    }
+
+
+    // Drop clients from room. We are being naughty and using the provisioners method for this.
+    try {
+        yield this.ircBridge.getProvisioner()._leaveIfUnprovisioned({
+            log: log
+        },
```
turt2live: this is a little hard to read. Maybe something like this for formatting?

```
yield this.ircBridge.getProvisioner()._leaveIfUnprovisioned(
    { log: log },
    roomId,
    server,
    channel,
);
```

lib/DebugApi.js
```diff
@@ -201,4 +206,143 @@ DebugApi.prototype.run = function() {
...
+        });
+        req.on("end", () => {
+            if (body === "") {
+                reject({"error": "Body missing"});
+            }
+            try {
+                body = JSON.parse(body);
+                resolve(
+                    body
+                );
```
turt2live: can this not be a single line?

Half-Shot:
>Technically this could unlink one room and send a message in the other

Do we not check https://github.com/matrix-org/matrix-appservice-irc/pull/611/files#diff-66fea70aa58dc47efcf7321a335ad423R241 ?

anoadragon453:
@Half-Shot so that would fail if the fields didn't match up?

anoadragon453:
Everyone on Team Bridge being happy.

Half-Shot:
@anoadragon453 @turt2live is delegating his part of the review process to you, so *are you happy* :innocent: :innocent: :innocent:

lib/DebugApi.js
```diff
@@ -201,4 +206,140 @@ DebugApi.prototype.run = function() {
...
+Remove Alias: ${remove_alias}`);
+
+    // Find room
+    let room = yield this.ircBridge.getStore().getRoom(
+        roomId,
+        domain,
+        channel,
+        "alias"
+    );
+    if (room === null) {
```
anoadragon453: Would this also catch room being `undefined`?

lib/DebugApi.js
```diff
@@ -201,4 +206,140 @@ DebugApi.prototype.run = function() {
...
+    );
+    result.stages.push("Removed room from store");
+
+    if (notice) {
+        try {
+            yield this.ircBridge.getAppServiceBridge().getIntent().sendEvent(roomId, "notice",
+            {
+                body: `This room has been unbridged from ${channel} (${server.getReadableName()})`
+            });
+            result.stages.push("Left notice in room.");
```
anoadragon453: Punctuation inconsistent :P

anoadragon453:
Rest looks fine, though your lack of comments disturbs me.

lib/DebugApi.js
Half-Shot: The function only returns Room or null. :)

Half-Shot:
Better?

PR #612, opened 2018-07-11
Bump Jasmine and use NYC

Half-Shot:
* ``istanbul`` doesn't understand ``async`` syntax
* ``nyc`` does
* Running jasmine 2.X with ``nyc`` produces OOMs
* Updating to 3.X and using ``nyc`` = async :tada: , no OOMs


Half-Shot:
Turns out shifting to 3.X will cause a OOM regardless. Oh dear.

Half-Shot:
(The OOMs were fixed by us splitting things out for travis, and this change gives us a nice speed boost)

PR #615, opened 2018-07-17
Option to allow expired certs

Half-Shot:


turt2live:
Link to docs that says this works?

Half-Shot:
https://github.com/matrix-org/node-irc/blob/master/lib/irc.js#L867

turt2live:
Generally it looks safe, but I think it needs to be added to the schema and example config.

Half-Shot:
That it does. Halfy does so.

config.sample.yaml
```diff
@@ -58,6 +58,9 @@ ircService:
       ssl: true
       # Whether or not IRC server is using a self-signed cert or not providing CA Chain
       sslselfsign: false
+      # Whether to allow expired certs when connecting to the IRC server.
+      # Usually this should be off. Default: false.
```
turt2live: a comment about how this interacts with/against `sslselfsign` would be nice.

Half-Shot: I think I'm failing to see how they intersect. This is about expired certs, not ones that might be invalid due to self signing?

turt2live: because I'm an idiot and read the wrong words. Ignore me on this comment - LGTM overall.

Half-Shot: :) *pats travis*

PR #616, opened 2018-07-17
Show groups in newly created portals

Half-Shot:


config.sample.yaml
```diff
@@ -59,6 +59,7 @@ ircService:
       # Whether or not IRC server is using a self-signed cert or not providing CA Chain
       sslselfsign: false
       # Whether to allow expired certs when connecting to the IRC server.
+      # This protects explitially against expired certs, not invalid ones.
```
turt2live: this probably should be in that other PR that was merged ;)

Half-Shot: This was the sentence I wrote before I changed my mind :p

Half-Shot:
>One thing I'm wondering is if the IRC bridge should also try and set the related groups for plumbed rooms, preserving whatever is currently there. It shouldn't be catastrophic if it fails though.

I feel more like we should just make people aware that the groups exist rather than try to force them into the room state for plumbed, since we don't really *own* the rooms in the same way.

turt2live:
Can has an option? Might not be strictly included in this PR, but something for future consideration.

lib/irc/IrcServer.js
```diff
@@ -146,6 +146,15 @@ IrcServer.prototype.getJoinRule = function() {
     return this.config.dynamicChannels.joinRule;
 };
 
+IrcServer.prototype.areGroupsEnabled = function() {
+    return this.config.dynamicChannels.groupId !== undefined && 
+    this.config.dynamicChannels.groupId !== "";
```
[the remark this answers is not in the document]
Half-Shot: I'm wondering if we should fail fast somewhere rather than letting definitely invalid groupids make it's way into room state.

Half-Shot:
I guess we could make it part of the provisioner steps somehow.

Half-Shot:
@turt2live Will add the issue, I've added some group id verification. Could you check that?

lib/irc/IrcServer.js
```diff
@@ -6,6 +6,8 @@ var logging = require("../logging");
 var IrcClientConfig = require("../models/IrcClientConfig");
 var log = logging.get("IrcServer");
 
+const GROUP_ID_REGEX = /^\+\S*:\S*$/;
```
turt2live: This is over-inclusive. Synapse has a fairly well defined set of allowed characters for group IDs though: https://github.com/matrix-org/synapse/blob/2c3e995f38a411647ca08ae2959add577041ca97/synapse/types.py#L195-L214

Issue #408, opened 2017-03-29
Change default power levels for portal rooms


ara4n:
Salvaged from https://github.com/matrix-org/matrix-appservice-irc/issues/387:

Currently if you get ops on a portal room you still can't change room name, history visibility, or power levels in general.  These all push users to creating plumbed rooms, and then suffering double bridging ugliness.  We should just change the power levels of the existing rooms to be more generous (and fix up the existing ones).

Mikaela:
What are more generous power levels? I think those actions need 100 and the issue with 100 is that it cannot be reversed (it requires PL100 user to demote themselves).
I think the issue with being allowed to set PL with less rhan PL100 was that you can promote yourself or something, it was once discussed on #irc:matrix.org, but I am unsure as I am on mobile.

Mikaela:
I repeated my question at Matrix some time ago:

> https://github.com/matrix-org/matrix-appservice-irc/issues/408 was left open on does it want ops syncing to provide a higher power level or decrease m.power_levels or what how?

And Kegan replied:

> something like that, either of them
> just to make it so they can do more than currently
> so people aren't tempted to try to provision
> we could lower the level needed for those actions (probably not power levels)
> or we could raise the amount that +o gives (but I think things like the name are 100 which obviously  won't work since we can't demote them then)
> so probably lower the level
> which the bridge will have to do retrospectively

* https://matrix.to/#/!SudviOJlimDvrGdFCY:matrix.org/%2414950381881741813fdZnT:matrix.org

Looking at what I think to be current defaults...

```
{
  "events_default": 0,
  "invite": 0,
  "state_default": 50,
  "redact": 50,
  "ban": 50,
  "users_default": 0,
  "events": {
    "m.room.avatar": 50,
    "m.room.name": 50,
    "m.room.power_levels": 100,
    "m.room.canonical_alias": 50,
    "m.room.history_visibility": 100
  },

}
```

...I think the issue that drives people to try provisioning would be `m.room.history_visibility` which should be lowered from 100 to 50.

Possibly unrelated to this specific issue, I believe that the event `m.encryption` should need PL100 rather than 50 (comes from `state_default`?) also outside of the IRC bridge, but it's even more important with the IRC bridge as it stops working if encryption is enabled and that button button just is attractive especially to new users for some reason.

*I also think I saw something about listing in room directory requiring PL 100, but as I cannot find it anywhere anymore or what the event name was, I think I am imagining it.

[...]

PR #619, opened 2018-07-18
Set some default power levels for portal rooms

Half-Shot:
This attempts to address some of the concerns brought forward by #408.

I'd like to have at least @Mikaela's feedback on this if possible.


Mikaela:
I think it does what #408 wishes, but it probably won't affect existing rooms?

Is oPowerfulOne the appservice MXID like I think it is, and not random user?

PR #621, opened 2018-07-20
WIP: A prometheus metric of reconnecting clients

michaelkaye:
#604 removed logging for pending reconnect users. Something like this would give visibility to the size of that reconnection queue in prometheus, which would be a better place for it than the log files.

I'm not sure if this is the right way to provide this sort of metric so happy for it to exist in another way.

I'm not even quite sure if we want to look at the number who are waiting to reconnect or the total of waiting to connect + currently reconnecting.

### DIFF
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -126,6 +126,7 @@ IrcBridge.prototype._initialiseMetrics = function() {
             matrixRoomConfigs: 0,
             remoteRoomConfigs: 0,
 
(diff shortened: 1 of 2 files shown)


danopia:
>I'm not even quite sure if we want to look at the number who are waiting to reconnect or the total of waiting to connect + currently reconnecting.

Surely you'd want to have both numbers reported independently, so you have the freedom of stacking them in a graph or splitting out into two different graphs?

michaelkaye:
Closing in favour of #622

PR #622, opened 2018-07-23
Additional metrics

Half-Shot:
This adds the following metrics: **EDITED**
- ``clientpool_reconnect_queue{serverdomain}``
  Number of IRC clients waiting to reconnect after losing connection.
- ``user_leave_queue{serverdomain}``
  Number of leaves queued I->M after receiving the /joined_users member set.
- ``user_join_queue{serverdomain}``
  Number of joins queued I->M after receiving the /joined_users member set.
- ``irchandler_calls{method}``
  Functions called in the IrcHandler. This includes things like ``onJoin`` which handles new IRC users appearing and onPart. Might be particularly useful in monitoring netsplits.

### DIFF
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -165,6 +165,10 @@ IrcBridge.prototype._initialiseMetrics = function() {
         help: "Number of leave requests queued up for virtual users on the bridge.",
         labels: ["server"]
     });
+
+    const memberListJoinQueue = metrics.addGauge({
+        name: "user_join_queue",
+        help: "Number of join requests queued up for virtual users on the bridge.",
         labels: ["server"]
     });
 
@@ -184,6 +188,10 @@ IrcBridge.prototype._initialiseMetrics = function() {
             memberListLeaveQueue.set(
                 {server},
                 this.memberListSyncers[server].getUsersWaitingToLeave()
+            );
+            memberListJoinQueue.set(
+                {server},
+                this.memberListSyncers[server].getUsersWaitingToJoin()
             );
         });
 
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -116,9 +116,9 @@ function IrcBridge(config, registration) {
 }
 
 IrcBridge.prototype._initialiseMetrics = function() {
-    var zeroAge = new AgeCounters();
+    const zeroAge = new AgeCounters();
 
-    var metrics = this._bridge.getPrometheusMetrics();
+    const metrics = this._bridge.getPrometheusMetrics();
 
     this._bridge.registerBridgeGauges(() => {
         return {
@@ -140,7 +140,7 @@ IrcBridge.prototype._initialiseMetrics = function() {
         };
     });
 
-    var timers = this._timers = {};
+    const timers = this._timers = {};
 
     timers.matrix_request_seconds = metrics.addTimer({
         name: "matrix_request_seconds",
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -165,17 +165,30 @@ IrcBridge.prototype._initialiseMetrics = function() {
         help: "Number of leaves queued up for virtual users from initial sync",
         labels: ["server"]
     });
+
+    const ircHandlerCalls = metrics.addCounter({
+        name: "irchandler_calls",
+        help: "Track calls made to the IRC Handler",
+        labels: ["method"]
+    });
     metrics.addCollector(() => {
         this.ircServers.forEach((server) => {
             reconnQueue.set({server: server.domain},
                 this._clientPool.totalReconnectsWaiting(server.domain)
             );
         });
+
         Object.keys(this.memberListSyncers).forEach((server) => {
             memberListLeaveQueue.set(
                 {server},
                 this.memberListSyncers[server].getUsersWaitingToLeave()
             );
+        });
+
+        const ircMetrics = this.ircHandler.getMetrics();
+        Object.keys(ircMetrics).forEach((method) => {
+            const value = ircMetrics[method];
+            ircHandlerCalls.inc({method}, value);
         });
     });
 };
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -159,10 +159,22 @@ IrcBridge.prototype._initialiseMetrics = function() {
         help: "Number of irc connections waiting to reconnect",
         labels: ["server"]
     });
+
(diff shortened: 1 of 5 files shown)


lib/bridge/IrcHandler.js
```diff
@@ -538,14 +550,16 @@ IrcHandler.prototype.onJoin = Promise.coroutine(function*(req, server, joiningUs
...
+                    ) {
                     req.log.error(`Not retrying join for ${room.getId()}.`);
                     return Promise.reject(err);
                 }
                 attempts++;
                 const delay = JOIN_DELAY_MS*attempts;
-                req.log.warn(`Failed to join ${room.getId()} attempts:${attempts} ${err}.delaying for ${delay}ms`);
-                req.log.debug(`Failed with: ${err.errcode} ${err.message}`);
+                req.log.warn(
+                `Failed to join ${room.getId()} attempts:${attempts} Delaying for ${delay}ms`
```
turt2live: Is there a reason why this is on a different line (and if so, why isn't it tabbed in?)

lib/bridge/IrcHandler.js
```diff
@@ -696,6 +712,7 @@ IrcHandler.prototype.onPart = Promise.coroutine(function*(req, server, leavingUs
 
 IrcHandler.prototype.onMode = Promise.coroutine(function*(req, server, channel, by,
                                                 mode, enabled, arg) {
+    this._callCountMetrics["mode"]++;
```
turt2live: for safety reasons, I think it might be best to use a function instead. It'd be dead simple, but probably worth it given the fairly wide range of possible values.

```
function _incrementCallCount(metric) {
  if (!this._callCountMetrics[metric]) this._callCountMetrics[metric] = 0;
  this._callCountMetrics[metric]++;
}
```

Half-Shot: Safety for what? Is it possible for any of these properties to be undefined?

spec/integ/irc-to-matrix.spec.js
```diff
@@ -616,7 +616,7 @@ describe("IRC-to-Matrix name bridging", function() {
         cli.joinRoom.and.callFake(function(r, opts) {
             if (count > 0) {
                 count--;
-                return Promise.reject({httpStatus: 504});
```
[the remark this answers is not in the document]
Half-Shot: It was more to fix the failing test :(

lib/bridge/IrcHandler.js
turt2live: Developer safety. At some point someone will forget to add the default to the object and be met with angry stack traces. Might as well mitigate it while we can.

lib/bridge/IrcHandler.js
Half-Shot: I suspect the linter was screaming at me. It should be tabbed, though.

lib/bridge/IrcHandler.js
```diff
@@ -538,14 +550,16 @@ IrcHandler.prototype.onJoin = Promise.coroutine(function*(req, server, joiningUs
         const joinRetry = (attempts) => {
             req.log.debug(`Joining room (attempts:${attempts})`);
             return intent.join(room.getId()).catch((err) => {
-                if (attempts > MAX_JOIN_ATTEMPTS || err.httpStatus === undefined || err.httpStatus <= 500) {
```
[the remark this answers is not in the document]
Half-Shot: Same as above :(

lib/bridge/IrcBridge.js
```diff
@@ -152,6 +152,45 @@ IrcBridge.prototype._initialiseMetrics = function() {
         help: "Histogram of processing durations of received remote messages",
         labels: ["outcome"],
     });
+
+    // Custom IRC metrics
+    const reconnQueue = metrics.addGauge({
+        name: "clientpool_reconnect_queue",
```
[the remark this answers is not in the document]
Half-Shot: I'm worried that we might be trying to over generalize this though. The concept of an IRC reconnection pool where TCP connections are trying to regain consciousness and a twitter/gitter/discord connection pool where we don't really handle connections in the same way.

But I guess from a raw reporting metric standpoint, we don't really care because the context should be obvious?

lib/bridge/IrcBridge.js
```diff
@@ -152,6 +152,45 @@ IrcBridge.prototype._initialiseMetrics = function() {
...
+        help: "Number of irc connections waiting to reconnect",
+        labels: ["server"]
+    });
+
+    const memberListLeaveQueue = metrics.addGauge({
+        name: "memberlistsync_leave_queue",
+        help: "Number of leaves queued up for virtual users from initial sync",
```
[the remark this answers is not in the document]
Half-Shot: The name is more or less transposed from the handler code. In this case I think it *would* make sense to generalise it.

lib/bridge/IrcHandler.js
Half-Shot: fiiine

lib/bridge/IrcBridge.js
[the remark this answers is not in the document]
michaelkaye: I was just mentally creating the dashboards for this, if we can have metrics that share names across bridges then :+1: as we don't need to have per-type dashboards if they are bridge-specific then that's what they are.

Half-Shot:
FYI it occured to me that we also might care about users who are waiting to be joined ``user_join_queue`` so that's done now.

lib/bridge/IrcHandler.js
```diff
@@ -953,4 +970,27 @@ IrcHandler.prototype._setMatrixRoomAsInviteOnly = function(room, isInviteOnly) {
...
+    this._callCountMetrics = {
+        "join.names": 0,
+        "join": 0,
+        "part": 0,
+        "pm": 0,
+        "invite": 0,
+        "topic": 0,
+        "message": 0,
+        "kick": 0,
+        "mode": 0,
```
anoadragon453: Extra ,

Half-Shot: [it's a personal preference thing](https://medium.com/@nikgraf/why-you-should-enforce-dangling-commas-for-multiline-statements-d034c98e36f8)

lib/bridge/IrcHandler.js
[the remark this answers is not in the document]
turt2live: If it's not part of this PR, please take it out :/

lib/bridge/IrcBridge.js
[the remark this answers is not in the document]
turt2live: This feels fairly IRC specific. Some bridges have a reconnect concept, although there's still too many that don't for me to feel comfortable.

PR #623, opened 2018-07-23
Configurable join retry

Half-Shot:
New "joinAttempts" config option which allows admins to configure how many attempts a join can take before giving up in``IrcHandler``. Setting to 0 switches the feature off.

### DIFF
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -258,6 +258,10 @@ ircService:
         # insert the IRC domain.
         # Optional. Default: "$NICK (IRC)". Example: "Alice (IRC)"
         displayName: "$NICK (IRC)"
+        # Number of tries a client can attempt to join a room before the request
+        # is discarded. Set to 0 to never retry.
+        # Optional. Default: 0
+        joinAttempts: 5
 
       # Configuration for virtual IRC users. The following variables are exposed:
       # $LOCALPART => The user ID localpart ("alice" in @alice:localhost)
--- a/lib/config/schema.yml
+++ b/lib/config/schema.yml
@@ -256,6 +256,9 @@ properties:
                                 displayName:
                                     type: "string"
                                     pattern: "\\$NICK"
+                                joinAttempts:
+                                    type: "integer"
+                                    minimum: 0
                         ircClients:
                             type: "object"
                             properties:
--- a/lib/irc/IrcServer.js
+++ b/lib/irc/IrcServer.js
@@ -248,6 +248,10 @@ IrcServer.prototype.getLineLimit = function() {
     return this.config.ircClients.lineLimit;
 };
 
+IrcServer.prototype.getJoinAttempts = function() {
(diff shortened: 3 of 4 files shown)


config.sample.yaml
```diff
@@ -258,6 +258,10 @@ ircService:
         # insert the IRC domain.
         # Optional. Default: "$NICK (IRC)". Example: "Alice (IRC)"
         displayName: "$NICK (IRC)"
+        # Number of tries a client can attempt to join a room before the request
+        # is discarded. Set to 0 to never retry.
```
turt2live: Can we make it 0 to be infinite, -1 to disable? Default to 5 (which is what is implied by the sample, but not listed as the default).

Half-Shot: Sure. Makes sense to me :)

Half-Shot:
Done

lib/irc/IrcServer.js
```diff
@@ -526,7 +530,8 @@ IrcServer.DEFAULT_CONFIG = {
     mappings: {},
     matrixClients: {
         userTemplate: "@$SERVER_$NICK",
-        displayName: "$NICK (IRC)"
+        displayName: "$NICK (IRC)",
+        joinAttempts: -1,
```
anoadragon453: No , at the end here.

Half-Shot: https://github.com/matrix-org/matrix-appservice-irc/pull/622#discussion_r204739927 :)

lib/config/schema.yml
```diff
@@ -256,6 +256,9 @@ properties:
                                 displayName:
                                     type: "string"
                                     pattern: "\\$NICK"
+                                joinAttempts:
+                                    type: "integer"
+                                    minimum: 0
```
turt2live: -1

spec/integ/irc-to-matrix.spec.js
```diff
@@ -616,7 +617,7 @@ describe("IRC-to-Matrix name bridging", function() {
         cli.joinRoom.and.callFake(function(r, opts) {
             if (count > 0) {
                 count--;
-                return Promise.reject({httpStatus: 504});
+                return Promise.reject({});
```
turt2live: What is this change?

Half-Shot: It's removing ``httpStatus: 504``, as it is redundant and probably misleading to keep it in.

turt2live: ftr this is the kind of change that should be in it's own PR: It's unrelated and it's impact is unclear on it's own.

Half-Shot: I think it's unfair to say unrelated, it's the same feature and all we are doing here is unbreaking it. I don't think it's going to be productive to start making sub feature branches unless it gets difficult to read the diffs.

PR #624, opened 2018-07-25
Set rawResponse=true in line with current default behaviour of matrix-js-sdk

michaelkaye:
```
Returning the raw JSON from uploadContent(). Future versions of the js-sdk will change this default, to return the parsed object. Set opts.rawResponse=false to change this behaviour now.
```
Currently the logs are full of this error.

Setting rawResponse=true is smallest change that will remove the warning, which reduces noise in the logs.

This was changed in the matrix-js-sdk nearly two years ago: https://github.com/matrix-org/matrix-js-sdk/pull/230


michaelkaye:
(I should note that moving to the already-parsed version may be a better solution, this is just coming from the point of view of reducing noise in logs).

Half-Shot:
LGTM.

PR #630, opened 2018-07-30
Cache mode internally

Half-Shot:
This PR adds mode caching to rooms (which are actually the channels modes, but we set them per room anyway). The modes that will be stored are ``m,i,k and s``.

This will currently only help with tracking modes across restarts, but will help us work out why a room has become desynced when it happens.

### DIFF
--- a/lib/DataStore.js
+++ b/lib/DataStore.js
@@ -325,23 +325,22 @@ DataStore.prototype.getModesForChannel = function (server, channel) {
     });
 };
 
-DataStore.prototype.setModeForRoom = Promise.coroutine(function*(roomId, mode, remove=False) {
-    log.info("setModeForRoom (mode=%s, roomId=%s, remove=%s)",
-        mode, roomId, remove
+DataStore.prototype.setModeForRoom = Promise.coroutine(function*(roomId, mode, enabled=True) {
+    log.info("setModeForRoom (mode=%s, roomId=%s, enabled=%s)",
+        mode, roomId, enabled
     );
     this._roomStore.getEntriesByMatrixId(roomId).then((entries) => {
         entries.forEach((entry) => {
             const modes = entry.remote.get("modes") || [];
             const hasMode = modes.includes(mode);
 
-            if ((!hasMode && remove) || (hasMode && !remove)) {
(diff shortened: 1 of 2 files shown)


lib/DataStore.js
```diff
@@ -311,6 +311,45 @@ DataStore.prototype.getMappingsForChannelByOrigin = function(server, channel, or
...
+DataStore.prototype.setModeForRoom = Promise.coroutine(function*(roomId, mode, enabled=True) {
+    log.info("setModeForRoom (mode=%s, roomId=%s, enabled=%s)",
+        mode, roomId, enabled
+    );
+    this._roomStore.getEntriesByMatrixId(roomId).then((entries) => {
+        entries.forEach((entry) => {
+            const modes = entry.remote.get("modes") || [];
+            const hasMode = modes.includes(mode);
+
+            if ((!hasMode && !enabled) || (hasMode && enabled)) {
```
turt2live: A comment on this would be great, given it took me a while to figure out what is going on. Also, this could be shortened to `hasMode === enabled` if you wanted.

lib/bridge/IrcHandler.js
```diff
@@ -882,12 +892,16 @@ IrcHandler.prototype._onPrivateMode = Promise.coroutine(function*(req, server, c
...
-                log.warn(`onMode: Unhandled channel mode ${mode}`);
+                req.log.warn(`onMode: Unhandled channel mode ${mode}`);
                 return Promise.resolve();
         }
     });
 
-    yield Promise.all(promises);
+    yield Promise.all(promises).then(() => {
+        matrixRooms.forEach((room) => {
+            this.ircBridge.getStore().setModeForRoom(room.getId(), mode, enabled);
```
turt2live: `k` and `i` aren't caught here.

turt2live: why not have the block set the modes on the room as early as ~line 860 (after the "no mapped rooms" check)?

lib/DataStore.js
Half-Shot: ...yes, it could couldn't it.

PR #631, opened 2018-07-31
Delay part/kick mirroring until inflight messages are completed.

Half-Shot:
This PR will count the number of inflight requests for a given room+user pair, sent via ``sendMatrixAction()``. Calls to leave/kick by the IrcHandler will be delayed until the count is 0. Note that a failed request is still counted as a completed request, so we won't spin forever.

This does not block on events we have yet to start sending, so it's possible for this to still race if a user is sufficiently unlucky enough. Steps to mitigate that would include counting an inflight request from the point of receiving a message which might be more difficult.

### DIFF
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -191,7 +191,7 @@ IrcBridge.prototype._initialiseMetrics = function() {
         help: "Track membership changes which are delayed behind inflight messages",
         labels: []
     });
-    
+
     const matrixHandlerConnFailureKicks = metrics.addCounter({
         name: "matrixhandler_connection_failure_kicks",
         help: "Track IRC connection failures resulting in kicks",
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -204,7 +204,7 @@ IrcBridge.prototype._initialiseMetrics = function() {
                 this._clientPool.totalReconnectsWaiting(server.domain)
             );
             let mxMetrics = this.matrixHandler.getMetrics(server.domain);
-            matrixHandlerConnFailureKicks.set(
+            matrixHandlerConnFailureKicks.inc(
                 {server: server.domain},
                 mxMetrics["connection_failure_kicks"] || 0
             );
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -572,6 +572,13 @@ IrcBridge.prototype.userHasMessagesInFlight = function(room, virtualUser) {
     return val > 0;
 }
 
+IrcBridge.prototype.getUserMessagesInFlightCount = function(room, virtualUser) {
+    // Undefined == 0 here, thanks JavaScript.
+    const key = `${room.getId()}:${virtualUser.getId()}`;
+    const val = this._messagesInFlight[key];
+    return val || 0;
+}
+
 IrcBridge.prototype.uploadTextFile = function(fileName, plaintext, req) {
     return this._bridge.getIntent().getClient().uploadContent({
         stream: new Buffer(plaintext),
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -185,6 +185,12 @@ IrcBridge.prototype._initialiseMetrics = function() {
         help: "Track calls made to the IRC Handler",
         labels: ["method"]
     });
+
+    const ircHandlerMembershipDelayed = metrics.addGauge({
+        name: "irchandler_delayed_membership",
+        help: "Track membership changes which are delayed behind inflight messages",
+        labels: []
+    });
     metrics.addCollector(() => {
         this.ircServers.forEach((server) => {
             reconnQueue.set({server: server.domain},
@@ -204,10 +210,11 @@ IrcBridge.prototype._initialiseMetrics = function() {
         });
 
         const ircMetrics = this.ircHandler.getMetrics();
-        Object.keys(ircMetrics).forEach((method) => {
+        Object.keys(ircMetrics.callCounters).forEach((method) => {
             const value = ircMetrics[method];
             ircHandlerCalls.inc({method}, value);
         });
+        ircHandlerMembershipDelayed.set(ircMetrics.delayedMembership);
     });
 };
 
--- a/lib/bridge/IrcBridge.js
+++ b/lib/bridge/IrcBridge.js
@@ -120,7 +120,7 @@ function IrcBridge(config, registration) {
 
     this._messagesInFlight = {
         //"roomId:userId" => number
-    }
+    };
 }
 
 IrcBridge.prototype._initialiseMetrics = function() {
@@ -501,24 +501,26 @@ IrcBridge.prototype.sendMatrixAction = function(room, from, action, req) {
     let msgtype = ACTION_TYPE_TO_MSGTYPE[action.type];
     let intent = this._bridge.getIntent(from.userId);
     if (msgtype) {
-        this._incrementMessageInFlight(room, from);
+        this._incrementMessagesInFlight(room, from);
+        let messageContent;
         if (action.htmlText) {
-            return intent.sendMessage(room.getId(), {
+            messageContent = {
                 msgtype: msgtype,
                 body: (
                     action.text || action.htmlText.replace(/(<([^>]+)>)/ig, "") // strip html tags
                 ),
                 format: "org.matrix.custom.html",
                 formatted_body: action.htmlText
-            }).finally(() => {
-                this._decrementMessageInFlight(room, from);
-            })
+            };
         }
-        return intent.sendMessage(room.getId(), {
-            msgtype: msgtype,
-            body: action.text
-        }).finally(() => {
-            this._decrementMessageInFlight(room, from);
+        else {
+            messageContent = {
+                msgtype: msgtype,
+                body: action.text
+            };
+        }
+        return intent.sendMessage(room.getId(), messageContent).finally(() => {
+            this._decrementMessagesInFlight(room, from);
         });
     }
     else if (action.type === "topic") {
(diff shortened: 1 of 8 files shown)


lib/bridge/IrcBridge.js
```diff
@@ -505,11 +510,15 @@ IrcBridge.prototype.sendMatrixAction = function(room, from, action, req) {
                 ),
                 format: "org.matrix.custom.html",
                 formatted_body: action.htmlText
-            });
+            }).finally(() => {
+                this._decrementMessageInFlight(room, from);
+            })
```
turt2live: semicolon

lib/bridge/IrcBridge.js
```diff
@@ -117,6 +117,10 @@ function IrcBridge(config, registration) {
     this._provisioner = null;
 
     this.publicitySyncer = new PublicitySyncer(this);
+
+    this._messagesInFlight = {
+        //"roomId:userId" => number
+    }
```
turt2live: semicolon

lib/bridge/IrcBridge.js
```diff
@@ -518,6 +527,32 @@ IrcBridge.prototype.sendMatrixAction = function(room, from, action, req) {
     return Promise.reject(new Error("Unknown action: " + action.type));
 };
 
+IrcBridge.prototype._incrementMessageInFlight = function(room, virtualUser) {
```
turt2live: This should be plural: incremementMesage**s**InFlight

lib/bridge/IrcBridge.js
```diff
@@ -518,6 +527,32 @@ IrcBridge.prototype.sendMatrixAction = function(room, from, action, req) {
...
+    const key = `${room.getId()}:${virtualUser.getId()}`;
+    if (this._messagesInFlight[key] === undefined) {
```
turt2live: minor thing: might as well make this a `!this_messagesInFlight[key]` rather than explicit type checking.

lib/bridge/IrcBridge.js
```diff
@@ -518,6 +527,32 @@ IrcBridge.prototype.sendMatrixAction = function(room, from, action, req) {
...
+    }
+}
+
+IrcBridge.prototype._decrementMessageInFlight = function(room, virtualUser) {
+    const key = `${room.getId()}:${virtualUser.getId()}`;
+    if (this._messagesInFlight[key] === undefined) {
+        this._messagesInFlight[key] = 0;
+    }
+    else {
+        this._messagesInFlight[key]--;
```
turt2live: I'd feel a bit better if this was capped at zero.

lib/bridge/IrcHandler.js
```diff
@@ -705,15 +709,26 @@ IrcHandler.prototype.onPart = Promise.coroutine(function*(req, server, leavingUs
...
+            return this.ircBridge.getAppServiceBridge().getIntent(
+                matrixUser.getId()
+            ).leave(room.getId());
+        });
     });
     stats.membership(true, "part");
     yield Promise.all(promises);
 });
 
+IrcHandler.prototype.delayWhileInflight = function(room, matrixUser) {
```
turt2live: Capital F on flight please

lib/bridge/IrcHandler.js
```diff
@@ -652,10 +654,12 @@ IrcHandler.prototype.onKick = Promise.coroutine(function*(req, server,
...
         let promises = matrixRooms.map((room) => {
-            req.log.info("Leaving (due to kick) room %s", room.getId());
-            return this.ircBridge.getAppServiceBridge().getIntent(
-                matrixUser.getId()
-            ).leave(room.getId());
+            return this.delayWhileInflight(room, matrixUser).then(() => {
+                req.log.info("Leaving (due to kick) room %s", room.getId());
+                return this.ircBridge.getAppServiceBridge().getIntent(
+                    matrixUser.getId()
+                ).leave(room.getId());
```
turt2live: The column limit is really driving me crazy. Not something for this PR, but we should consider raising it or removing it.

lib/bridge/IrcHandler.js
```diff
@@ -243,7 +246,6 @@ IrcHandler.prototype.onPrivateMessage = Promise.coroutine(function*(req, server,
     req.log.info("Relaying PM in room %s", pmRoom.getId());
     yield this.ircBridge.sendMatrixAction(pmRoom, virtualMatrixUser, mxAction, req);
 });
-
```
turt2live: we should keep this in

lib/bridge/IrcHandler.js
```diff
@@ -15,6 +15,9 @@ var QuitDebouncer = require("./QuitDebouncer.js");
 const JOIN_DELAY_MS = 250;
 const JOIN_DELAY_CAP_MS = 30 * 60 * 1000; // 30 mins
 
+// Delay for when messages are inflight and we want to leave.
+const LEAVE_INFLIGHT_DELAY_MS = 500;
```
turt2live: IN_FLIGHT (two words)

lib/bridge/IrcHandler.js
```diff
@@ -705,15 +709,26 @@ IrcHandler.prototype.onPart = Promise.coroutine(function*(req, server, leavingUs
...
+        });
     });
     stats.membership(true, "part");
     yield Promise.all(promises);
 });
 
+IrcHandler.prototype.delayWhileInflight = function(room, matrixUser) {
+    if (this.ircBridge.userHasMessagesInFlight(room, matrixUser)) {
+        return Promise.delay(LEAVE_INFLIGHT_DELAY_MS).then(() => {
+            return this.delayWhileInflight(room, matrixUser);
```
turt2live: It's entirely possible that this loops forever. Might be worth having either a maximum number of attempts or maximum delay time. Something like 30 minutes would be fine.

Only reason is to force an exit condition to avoid having the thing spin forever.

lib/bridge/IrcHandler.js
```diff
@@ -705,15 +709,26 @@ IrcHandler.prototype.onPart = Promise.coroutine(function*(req, server, leavingUs
...
+            ).leave(room.getId());
+        });
     });
     stats.membership(true, "part");
     yield Promise.all(promises);
 });
 
+IrcHandler.prototype.delayWhileInflight = function(room, matrixUser) {
+    if (this.ircBridge.userHasMessagesInFlight(room, matrixUser)) {
+        return Promise.delay(LEAVE_INFLIGHT_DELAY_MS).then(() => {
```
turt2live: There's a bit of a thundering herd problem here: Timers started at approximately the same time for the same duration are all going to terminate at the same time, leading to quite a lot of blocked operations unlocking all at once. Some jitter of ~100ms on each helps spread out the timer firing load, and therefore the potential for large blocking operations.

spec/integ/kicking.spec.js
```diff
@@ -88,6 +88,34 @@ describe("Kicking", function() {
             ircUserCli.emit("kick", config._chan, ircUser.nick, "KickerNick", "Reasons");
             yield kickPromise;
         }));
+        it("should not leave until all messages are sent", test.coroutine(function*() {
+            let SENT_MESSAGES = -1; // We send a message to in the beforeEach
```
turt2live: -to?

spec/integ/kicking.spec.js
```diff
@@ -88,6 +88,34 @@ describe("Kicking", function() {
...
+            const kickPromise = new Promise(function(resolve, reject) {
+                const ircUserSdk = env.clientMock._client(ircUser.id);
+                ircUserSdk.leave.and.callFake(function(roomId) {
+                    expect(roomId).toEqual(config._roomid);
+                    expect(SENT_MESSAGES).toEqual(3);
+                    resolve();
+                    return Promise.resolve();
```
turt2live: do we really have to resolve twice? Can we not `return resolve();` ?

spec/integ/kicking.spec.js
```diff
@@ -88,6 +88,34 @@ describe("Kicking", function() {
             ircUserCli.emit("kick", config._chan, ircUser.nick, "KickerNick", "Reasons");
             yield kickPromise;
         }));
+        it("should not leave until all messages are sent", test.coroutine(function*() {
+            let SENT_MESSAGES = -1; // We send a message to in the beforeEach
+            const kickPromise = new Promise(function(resolve, reject) {
+                const ircUserSdk = env.clientMock._client(ircUser.id);
+                ircUserSdk.leave.and.callFake(function(roomId) {
+                    expect(roomId).toEqual(config._roomid);
+                    expect(SENT_MESSAGES).toEqual(3);
```
turt2live: I'm a little concerned about racing here, but I think it should be okay. Paranoia would be adding a lock on the variable, potentially even an order of operations check. I'm really not sure that's needed here though.

Ralith:
Instead of using a brute-force polling timeout, why not set a `leaving` flag on the IRC user, and have every successful message delivery check whether `messagesInFlight == 0 && leaving`, and if so, emit the leave event immediately?

Half-Shot:
>Does the IRC bridge resync all room memberships on startup?

It does, yes.

lib/bridge/IrcBridge.js
Half-Shot: I'd rather make it more explicit than `!` ?

lib/bridge/IrcHandler.js
Half-Shot: How would this loop forever?

Half-Shot:
>Instead of using a brute-force polling timeout, why not set a leaving flag on the IRC user, and have every successful message delivery check whether messagesInFlight == 0 && leaving, and if so, emit the leave event immediately?

That's actually not a bad idea. I'll rethink this

Half-Shot:
So the new edition of this change doesn't include a timeout and only checks to see if membership is delayed after the messages are sent (failed or not), which should hopefully vastly reduce time taken. We also log more.

Any thoughts on metrics travis?

lib/bridge/IrcHandler.js
```diff
@@ -52,6 +52,8 @@ function IrcHandler(ircBridge) {
         //'$fromUserId $toUserId' : Promise
     };
 
+    this.membershipWaiting = new Map(); //userId => join|leave|kick|ban
```
turt2live: the comment is wrong as per line 774

spec/integ/kicking.spec.js
turt2live: still applies - the `to` is extraneous

spec/integ/kicking.spec.js
turt2live: *bump*

Half-Shot:
Closing this because our codebase has diverged a fair amount since the move to Typescript. Will re-open a new PR if we want to go back to this.

PR #639, opened 2018-08-07
Kick users if we fail to create a IRC connection on join

Half-Shot:
This also actually smells out iline disconnects too.

### DIFF
--- a/lib/bridge/MatrixHandler.js
+++ b/lib/bridge/MatrixHandler.js
@@ -17,6 +17,8 @@ const MSG_PMS_DISABLED = "[Bridge] Sorry, PMs are disabled on this bridge.";
 const MSG_PMS_DISABLED_FEDERATION = "[Bridge] Sorry, PMs are disabled on " +
 "this bridge over federation.";
 
+const KICK_RETRY_DELAY_MS = 15000;
+
 function MatrixHandler(ircBridge) {
     this.ircBridge = ircBridge;
     // maintain a list of room IDs which are being processed invite-wise. This is
@@ -853,9 +855,32 @@ MatrixHandler.prototype._onJoin = Promise.coroutine(function*(req, event, user) 
         }
         // get the virtual IRC user for this user
         promises.push(Promise.coroutine(function*() {
-            let bridgedClient = yield self.ircBridge.getBridgedClient(
-                room.server, user.getId(), (event.content || {}).displayname
-            );
+            let bridgedClient;
+            let kickUser = false;
+            try {
+                bridgedClient = yield self.ircBridge.getBridgedClient(
+                    room.server, user.getId(), (event.content || {}).displayname
+                );
+            } catch (e) {
+                // We need to kick on failure to get a client.
+                req.log.info(`${user.getId()} failed to get a IRC connection. Kicking from room.`);
+                kickUser = true;
+
+            }
+
+            const intent = this.ircBridge.getAppServiceBridge().getIntent();
+            while (kickUser) {
+                try {
+                    yield intent.kick(
+                        event.room_id, user.getId(),
+                        `Connection limit has been reached for ${room.server.domain}. Please try again later.`
+                    );
+                    break;
+                } catch (err) {
+                    req.log.warn(`User was not kicked. Retrying in ${KICK_RETRY_DELAY_MS}ms. ${err}`);
+                    yield Promise.delay(KICK_RETRY_DELAY_MS);
+                }
+            }
 
             // Check for a displayname change and update nick accordingly.
             if (event.content.displayname !== bridgedClient.displayName) {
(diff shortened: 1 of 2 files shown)


lib/bridge/MatrixHandler.js
```diff
@@ -853,9 +856,36 @@ MatrixHandler.prototype._onJoin = Promise.coroutine(function*(req, event, user)
...
+            try {
+                bridgedClient = yield self.ircBridge.getBridgedClient(
+                    room.server, user.getId(), (event.content || {}).displayname
+                );
+            }
+            catch (e) {
+                // We need to kick on failure to get a client.
+                req.log.info(`${user.getId()} failed to get a IRC connection. Kicking from room.`);
+                kickIntent = self.ircBridge.getAppServiceBridge().getIntent();
+
```
ara4n: spurious blank line

lib/bridge/MatrixHandler.js
```diff
@@ -17,6 +17,9 @@ const MSG_PMS_DISABLED = "[Bridge] Sorry, PMs are disabled on this bridge.";
 const MSG_PMS_DISABLED_FEDERATION = "[Bridge] Sorry, PMs are disabled on " +
 "this bridge over federation.";
 
+const KICK_RETRY_DELAY_MS = 15000;
```
ara4n: trying to kick every 15s feels a bit aggressive
and 500ms of jitter is nothing
i'd suggest 30+/-15 secs or something

ara4n:
lgtm. i wouldn't rush this out though, until the next freenode restart happens

ara4n:
(modulo comments)

lib/bridge/MatrixHandler.js
```diff
@@ -853,9 +856,36 @@ MatrixHandler.prototype._onJoin = Promise.coroutine(function*(req, event, user)
...
+                req.log.info(`${user.getId()} failed to get a IRC connection. Kicking from room.`);
+                kickIntent = self.ircBridge.getAppServiceBridge().getIntent();
+
+            }
+
+            while (kickIntent) {
+                try {
+                    yield kickIntent.kick(
+                        event.room_id, user.getId(),
+                    `Connection limit reached for ${room.server.domain}. Please try again later.`
```
turt2live: can we indent this (which pushes it on another line, which is annoying)?

Half-Shot:
(Next freenode restart is soonish, on the agenda for https://github.com/matrix-org/internal-config/issues/372)

lib/bridge/MatrixHandler.js
```diff
@@ -18,7 +18,7 @@ const MSG_PMS_DISABLED_FEDERATION = "[Bridge] Sorry, PMs are disabled on " +
 "this bridge over federation.";
 
 const KICK_RETRY_DELAY_MS = 15000;
-const KICK_DELAY_JITTER = 500;
+const KICK_DELAY_JITTER = 30000;
```
turt2live: err, delay_ms 30s, jitter 15s instead?

Half-Shot: I thought we wanted 30+/- 15?

lib/bridge/MatrixHandler.js
Half-Shot: We can't, because we go over the silly 100 char limit :(

lib/bridge/MatrixHandler.js
Half-Shot: I think this is is mathematically equivalent, and easier to read.

Half-Shot: Travis has explained why I am wrong..but also why this approach works better.

ara4n: delay 15, jitter 30 (if jitter is additive here, which it is) is fine.

ara4n:
btw, the lgtm is contingent on this having been tested and confirmed working.

Half-Shot:
>btw, the lgtm is contingent on this having been tested and confirmed working.

I did test on my local box, it works fine :)

Half-Shot:
(The failures are dodgy provisoning tests that are too slow under node 6 and die)

lib/bridge/MatrixHandler.js
turt2live: > Travis has explained why I am wrong..but also why this approach works better.

For curious onlookers: the delay doesn't matter too much beyond giving a static offset to when kicks start. Having a 30s jitter means the spread of requests is much greater, which is a Good Thing for this. It has to be additive otherwise with a 15s delay a quarter of the values are negative. A 15s jitter instead gives too much concentration, making the problem worse. Having a 30s delay just means everything takes longer, so we shouldn't do that.

or TLDR: 30s is bigger than 15s for jitter, and we should do that. Additive because negative time is hard.

PR #643, opened 2018-08-10
SASL support

### LINKED ISSUE #470: SASL support
Attempting to connect to a freenode from a AS running in AWS leads to this error:

` {"command":"ERROR","rawCommand":"ERROR","commandType":"normal","args":["Closing Link: ec2-34-192-170-23.compute-1.amazonaws.com (SASL access only)"]}`

Edit: Should probably just rename this to a feature request asking for SASL support in reality.  Please feel free to do so if that is the case! :)

Half-Shot:
This should hopefully fix #470

### DIFF
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -63,7 +63,7 @@ ircService:
       ssl: true
       # Whether or not IRC server is using a self-signed cert or not providing CA Chain
       sslselfsign: false
-      # Should the connection attempt to identify via SASL (if a server or user password is given?)
+      # Should the connection attempt to identify via SASL (if a server or user password is given)
       # If false, this will use PASS instead. If SASL fails, we do not fallback to PASS.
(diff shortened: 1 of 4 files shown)


config.sample.yaml
```diff
@@ -63,6 +63,9 @@ ircService:
       ssl: true
       # Whether or not IRC server is using a self-signed cert or not providing CA Chain
       sslselfsign: false
+      # Should the connection attempt to identify via SASL (if a server or user password is given?)
```
turt2live: question should be answered

Half-Shot: Not sure why there is a ?. It reads fine without.

eqyiel:
@Half-Shot would you happen to know anything about this (kind of related) issue? https://github.com/matrix-org/matrix-appservice-irc/issues/689

PR #645, opened 2018-08-11
Add err_nononreg so we can announce PMs that failed

Half-Shot:
This adds a dependency on the latest version of matrix-org/node-irc


Half-Shot:
@turt2live Made some changes, now it even works \o/

lib/irc/BridgedClient.js
```diff
@@ -175,12 +175,13 @@ BridgedClient.prototype.connect = Promise.coroutine(function*() {
             }
         });
         connInst.client.addListener("error", (err) => {
+            const ERRORS_TO_FORCE = ["err_nononreg"]
```
turt2live: force what? (please add comment)

PR #647, opened 2018-08-13
Formatting of replies

### LINKED ISSUE #636: Replies are really mangled over the bridge
I suspect we need to think of a smart way to handle this, because the large text pastebin-ifier grabs it and makes for a poor UX.

Half-Shot:
Fixes #636

The format turns it into something like `<source sender "source text"> reply text`

Depends on https://github.com/matrix-org/matrix-appservice-bridge/pull/79

### DIFF
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -431,5 +431,5 @@ ircService:
 
   # Config for Matrix -> IRC bridging
   matrixHandler:
-    # Cache this many matrix events in memory to be used for m.relates_to messages.
+    # Cache this many matrix events in memory to be used for m.relates_to messages (usually replies).
     eventCacheSize: 4096
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -428,3 +428,8 @@ ircService:
   # `!storepass server.name passw0rd. When a connection is made to IRC on behalf of
   # the Matrix user, this password will be sent as the server password (PASS command).
   passwordEncryptionKeyPath: "passkey.pem"
+
+  # Config for Matrix -> IRC bridging
+  matrixHandler:
+    # Cache this many matrix events in memory to be used for m.relates_to messages.
+    eventCacheSize: 4096
--- a/lib/bridge/MatrixHandler.js
+++ b/lib/bridge/MatrixHandler.js
@@ -1263,22 +1263,25 @@ MatrixHandler.prototype._sendIrcAction = Promise.coroutine(
         yield this.ircBridge.sendIrcAction(ircRoom, ircClient, ircAction);
         return;
     }
+
+    let text = event.content.body;
+    let cacheText = text;
+    if (event.content["m.relates_to"] && event.content["m.relates_to"]["m.in_reply_to"]) {
+        const reply = yield this._textForReplyEvent(event, ircRoom);
+        ircAction.text = text = reply.formatted;
+        cacheText = reply.reply;
+    }
+    this._eventCache.set(event.event_id, {
+        body: cacheText.substr(0, REPLY_SOURCE_MAX_LENGTH),
+        sender: event.sender
+    });
+
     /**
      * Cache events in here so we can refer to them for replies.
-     */
-    this._eventCache.set(event.event_id, {
-        body: event.content.body.substr(0, REPLY_SOURCE_MAX_LENGTH),
-        sender: event.sender }
-    );
+    */
     if (this._eventCache.size > this._eventCacheMaxSize) {
         const delKey = this._eventCache.entries().next().value[0];
         this._eventCache.delete(delKey);
-    }
-
-    let text = event.content.body;
-    if (event.content["m.relates_to"] && event.content["m.relates_to"]["m.in_reply_to"]) {
-        text = yield this._textForReplyEvent(event, ircRoom);
-        ircAction.text = text;
     }
 
     // Check for the existance of the getSplitMessages method.
@@ -1527,7 +1530,10 @@ MatrixHandler.prototype._textForReplyEvent = Promise.coroutine(function*(event, 
                 this._eventCache.set(eventId, {sender: rplName, body: rplSource});
             } catch (err) {
                 // If we couldn't find the event, then frankly we can't trust it and we won't treat it as a reply.
-                return rplText;
+                return {
+                    formatted: rplText,
+                    reply: rplText,
+                };
             }
         }
 
@@ -1560,7 +1566,10 @@ MatrixHandler.prototype._textForReplyEvent = Promise.coroutine(function*(event, 
             );
         }
 
-        return `<${rplName}${rplSource}> ${rplText}`;
+        return {
+            formatted: `<${rplName}${rplSource}> ${rplText}`,
+            reply: rplText,
+        };
     }
 });
 
--- a/lib/bridge/MatrixHandler.js
+++ b/lib/bridge/MatrixHandler.js
@@ -1262,6 +1262,29 @@ MatrixHandler.prototype._sendIrcAction = Promise.coroutine(
     }
 
     let text = event.content.body;
+    const REPLY_SOURCE_MAX_LENGTH = 32;
+    const REPLY_NAME_MAX_LENGTH = 12;
+    if (event.content["m.relates_to"] && event.content["m.in_reply_to"]) {
+        // This is a reply.
+        const match = /> <(@.*:.*)>(.*)\n\n(.*)/.exec(text);
+        if (match.length === 4) {
+            let rplName;
+            let sourceLength = REPLY_SOURCE_MAX_LENGTH - "...".length;
+            const rplSource = match[2].substr(0, sourceLength) + (match[2].length > sourceLength ? "..." : "");
+            const rplText = match[3];
+            
+            // Fetch the sender's IRC nick.
+            const sourceClient = this.ircBridge.getIrcUserFromCache(ircRoom.server, match[1]);
+            if (sourceClient) {
+                rplName = sourceClient.nick;
+            }
+            else {
+                // Somehow we failed, so fallback to userid.
+                rplName = match[1].substr(1, Math.min(REPLY_NAME_MAX_LENGTH, userid.indexOf(":") - 1));
+            }
+            text = `<reply to ${rplName} "${rplSource}"> ${rplText}`;
+        }
+    }
 
     // Generate an array of individual messages that would be sent
     let potentialMessages = ircClient.unsafeClient.getSplitMessages(ircRoom.channel, text);
--- a/lib/bridge/MatrixHandler.js
+++ b/lib/bridge/MatrixHandler.js
@@ -19,6 +19,7 @@ const MSG_PMS_DISABLED_FEDERATION = "[Bridge] Sorry, PMs are disabled on " +
 
 const KICK_RETRY_DELAY_MS = 15000;
 const KICK_DELAY_JITTER = 30000;
+const EVENT_CACHE_SIZE = 4096;
 
 function MatrixHandler(ircBridge) {
     this.ircBridge = ircBridge;
@@ -30,6 +31,7 @@ function MatrixHandler(ircBridge) {
     };
 
     this._memberTracker = null;
+    this._eventCache = new Map(); //eventId => {body, sender}
 }
 
 // ===== Matrix Invite Handling =====
@@ -1259,6 +1261,16 @@ MatrixHandler.prototype._sendIrcAction = Promise.coroutine(
         !(ircClient.unsafeClient && ircClient.unsafeClient.getSplitMessages)) {
         yield this.ircBridge.sendIrcAction(ircRoom, ircClient, ircAction);
         return;
+    }
+
+    /**
+     * Cache events in here so we can refer to them for replies.
+     * We deliberately avoid fetching events from the homeserver due to speed concerns.
+     */
+    this._eventCache.set(event.event_id, { body: event.content.body, sender: event.sender });
+    if (this._eventCache.size > EVENT_CACHE_SIZE) {
+        const delKey = this._eventCache.entries().next().value[0];
+        this._eventCache.delete(delKey);
     }
 
     let text = event.content.body;
--- a/lib/bridge/MatrixHandler.js
+++ b/lib/bridge/MatrixHandler.js
@@ -1284,9 +1284,7 @@ MatrixHandler.prototype._sendIrcAction = Promise.coroutine(
         sender: event.sender
     });
 
-    /**
-     * Cache events in here so we can refer to them for replies.
-    */
+    // Cache events in here so we can refer to them for replies.
     if (this._eventCache.size > this._eventCacheMaxSize) {
         const delKey = this._eventCache.entries().next().value[0];
         this._eventCache.delete(delKey);
@@ -1571,11 +1569,11 @@ MatrixHandler.prototype._textForReplyEvent = Promise.coroutine(function*(event, 
         // Cut to a maximum length.
         rplSource = lines[0].substr(0, REPLY_SOURCE_MAX_LENGTH);
         // Ellipsis if needed.
-        if (rplSource.length > REPLY_SOURCE_MAX_LENGTH) {
+        if (lines[0].length > REPLY_SOURCE_MAX_LENGTH) {
             rplSource = rplSource + "...";
         }
         // Wrap in formatting
-        rplSource = ` "${lines[0]}"`;
+        rplSource = ` "${rplSource}"`;
     }
     else {
         // Don't show a source because we couldn't format one.
@@ -1588,7 +1586,7 @@ MatrixHandler.prototype._textForReplyEvent = Promise.coroutine(function*(event, 
         rplName = sourceClient.nick;
     }
     else {
-        // Somehow we failed, so fallback to userid.
+        // Somehow we failed, so fallback to localpart.
         rplName = rplName.substr(1,
             Math.min(REPLY_NAME_MAX_LENGTH, rplName.indexOf(":") - 1)
         );
--- a/lib/bridge/MatrixHandler.js
+++ b/lib/bridge/MatrixHandler.js
@@ -1282,7 +1282,7 @@ MatrixHandler.prototype._sendIrcAction = Promise.coroutine(
                 // Somehow we failed, so fallback to userid.
                 rplName = match[1].substr(1, Math.min(REPLY_NAME_MAX_LENGTH, userid.indexOf(":") - 1));
             }
-            text = `<reply to ${rplName} "${rplSource}"> ${rplText}`;
+            text = `<${rplName} "${rplSource.trim()}"> ${rplText}`;
             ircAction.text = text;
         }
     }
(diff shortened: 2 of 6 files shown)


lib/bridge/MatrixHandler.js
```diff
@@ -1262,6 +1262,33 @@ MatrixHandler.prototype._sendIrcAction = Promise.coroutine(
...
+            if (sourceClient) {
+                rplName = sourceClient.nick;
+            }
+            else {
+                // Somehow we failed, so fallback to userid.
+                rplName = match[1].substr(1,
+                    Math.min(REPLY_NAME_MAX_LENGTH, userid.indexOf(":") - 1)
+                );
+            }
+            text = `<${rplName} "${rplSource.trim()}"> ${rplText}`;
```
turt2live: If we're going to trim, surely we should trim ahead of getting to this point?

lib/bridge/MatrixHandler.js
```diff
@@ -1262,6 +1262,33 @@ MatrixHandler.prototype._sendIrcAction = Promise.coroutine(
     }
 
     let text = event.content.body;
+    const REPLY_SOURCE_MAX_LENGTH = 32;
+    const REPLY_NAME_MAX_LENGTH = 12;
+    if (event.content["m.relates_to"] && event.content["m.relates_to"]["m.in_reply_to"]) {
+        // This is a reply.
+        const match = /> <(@.*:.*)>(.*)\n\n(.*)/.exec(text);
```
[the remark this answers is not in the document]
Half-Shot: Would love to, but I'm not sure I can risk the perf of doing so.

lib/bridge/MatrixHandler.js
```diff
@@ -1262,6 +1262,33 @@ MatrixHandler.prototype._sendIrcAction = Promise.coroutine(
...
+        if (match.length === 4) {
+            let rplName;
+            let sourceLength = REPLY_SOURCE_MAX_LENGTH - "...".length;
```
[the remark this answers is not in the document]
Half-Shot: Would be tempted to..but there is the risk of blowing up non UTF8 supporting clients.

lib/bridge/MatrixHandler.js
Half-Shot: Point

lib/bridge/MatrixHandler.js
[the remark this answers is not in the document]
t3chguy: the fallback you're relying on is intended to be removed so relying on it will eventually even for non-malicious events will screw things up.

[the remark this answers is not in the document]
t3chguy: also the `/event/` API unlike the context API should be rather cheap

Half-Shot:
@turt2live I've added a recent event cache containing events we've seen so you can pull the source from it, if it exists.

lib/bridge/MatrixHandler.js
```diff
@@ -19,6 +19,7 @@ const MSG_PMS_DISABLED_FEDERATION = "[Bridge] Sorry, PMs are disabled on " +
 
 const KICK_RETRY_DELAY_MS = 15000;
 const KICK_DELAY_JITTER = 30000;
+const EVENT_CACHE_SIZE = 4096;
```
turt2live: I doubt this is enough for Freenode. Matrix.org is handling ~25Hz of messages, so we'll round up to 30 for math+future capabilities. At 30Hz, this cache is overturned every 3 minutes. Given most replies I've seen (note: not actually tested) happen 5-10+ minutes after the original message, your cache is too small.

Having a large cache means you're more likely to run out of memory though. Combined with the fact that the cache is not self-expiring, you'll never reclaim that memory during low volume periods.

lib/bridge/MatrixHandler.js
```diff
@@ -1261,7 +1263,52 @@ MatrixHandler.prototype._sendIrcAction = Promise.coroutine(
         return;
     }
 
+    /**
+     * Cache events in here so we can refer to them for replies.
+     * We deliberately avoid fetching events from the homeserver due to speed concerns.
+     */
+    this._eventCache.set(event.event_id, { body: event.content.body, sender: event.sender });
```
turt2live: There's no protection from massive events landing in the cache.

lib/bridge/MatrixHandler.js
```diff
@@ -1261,7 +1263,52 @@ MatrixHandler.prototype._sendIrcAction = Promise.coroutine(
...
+        if (match.length === 4) {
+            let rplName = match[1];
+            let rplSource = match[2].trim();
+            if (this._eventCache.has(eventId)) {
+                rplName = this._eventCache.get(eventId).sender;
+                rplSource = this._eventCache.get(eventId).body;
+            }
+            // Replace newlines so we can fit this in the message body.
+            // It's not a perfect solution but it's legible.
+            rplSource = rplSource.replace(/\v/, '. ');
```
turt2live: or just grab the first line and call it a day. Modifying user's text like this feels too invasive for a bridge.

lib/bridge/MatrixHandler.js
```diff
@@ -1261,7 +1263,52 @@ MatrixHandler.prototype._sendIrcAction = Promise.coroutine(
...
+     * Cache events in here so we can refer to them for replies.
+     * We deliberately avoid fetching events from the homeserver due to speed concerns.
+     */
+    this._eventCache.set(event.event_id, { body: event.content.body, sender: event.sender });
+    if (this._eventCache.size > EVENT_CACHE_SIZE) {
+        const delKey = this._eventCache.entries().next().value[0];
+        this._eventCache.delete(delKey);
+    }
+
     let text = event.content.body;
```
turt2live: there's no read from the cache?

turt2live: also if we cache miss, we should do a hit to `/event`. Synapse's cache is likely going to be faster and more useful than ours anyways.

Half-Shot: >Synapse's cache is likely going to be faster and more useful than ours anyways.

I doubt it's going to be faster.

lib/bridge/MatrixHandler.js
Half-Shot: This is assuming a lot of things.
> Matrix.org is handling ~25Hz of messages.

You're assuming the bridge can *see* 25Hz of events. Even freenode isn't in that many rooms and is excluding things like PM traffic and the ~90% of rooms that aren't in a bridge. Let's not also forget that some of that traffic is IRC, which won't hit the cache. I seriously doubt the bridge is seeing an *average* of 30Hz M->I traffic a second. 

>likely to run out of memory though

We could up this to 16384 and be laughing to be honest, it's not a lot of memory.

>cache is not self-expiring

It's really not something I want to waste CPU cycles on.

lib/bridge/MatrixHandler.js
Half-Shot: True, we should trim this to how much of the body we actually need.

lib/bridge/MatrixHandler.js
Half-Shot: Grabbing the first line makes sense...until someone does some really strange paragraphing. I'm not overly worried about editing out some newlines given the original text will be in IRC scrollback, and this version isn't too far removed from the original.

lib/bridge/MatrixHandler.js
Half-Shot: https://github.com/matrix-org/matrix-appservice-irc/blob/6e0ac53ca578b1bf353e97a2d75f624b8612377a/lib/bridge/MatrixHandler.js#L1286

lib/bridge/MatrixHandler.js
turt2live: Sorry, you're right - I was thinking of a similar nightmare from a different project and started having flashbacks. Would be nice to have the cache size configurable.

lib/bridge/MatrixHandler.js
turt2live: This is going to be the thing I'm going to push back on fairly strongly: where possible we shouldn't be modifying a user's message - that's not the job of the bridge. This includes format markers which may introduce less clarity.

If someone does something weird with paragraphs, that's their own fault in my honest opinion. For a safety net, we can grab the first non-empty line of text.

lib/bridge/MatrixHandler.js
turt2live: Synapse's cache itself is faster. The HTTP overhead obviously won't be good.

Still should have a cache miss lookup. Put it behind an option if you'd like.

The event body is not looked up from the cache, therefore making the unsafe reliance on the sender's event still there.

Half-Shot: The JS SDK doesn't support the /event endpoint, which is going to really bloat this PR out. I've made the caching configurable anyway.

turt2live: It'd be great to have a js-sdk PR for it then. I'm quite surprised the endpoint doesn't already exist, given it's age and usefulness. Having a cache miss is important to deal with to prevent unexpected and unpredictable messages being sent, particularly by those who would be unaware of the problem in the first place.

lib/bridge/MatrixHandler.js
```diff
@@ -1254,60 +1257,35 @@ MatrixHandler.prototype._onMessage = Promise.coroutine(function*(req, event) {
...
      * Cache events in here so we can refer to them for replies.
-     * We deliberately avoid fetching events from the homeserver due to speed concerns.
      */
-    this._eventCache.set(event.event_id, { body: event.content.body, sender: event.sender });
-    if (this._eventCache.size > EVENT_CACHE_SIZE) {
+    this._eventCache.set(event.event_id, {
+        body: event.content.body.substr(0, REPLY_SOURCE_MAX_LENGTH),
+        sender: event.sender }
+    );
+    console.log("SET!", this._eventCache, this._eventCache.size);
```
turt2live: .

Half-Shot:
So to drag comments out from inline for a second, the caching thing is a problem. Frankly the reply format sucks as is, because I can't actually distinguish what is reply text and what is source text without running a regex through it so this is going to break at some point (according to @t3chguy, we are going to be dropping the source text from ``body``).

Separately, cache misses should mean we fetch from /event except the JS-SDK doesn't actually support it yet. I could do a PR, but then that is going to add more delay on a PR I'd like to make the 0.11 release. It's very tempting for cache misses to forgo the source text entirely.

lib/bridge/MatrixHandler.js
turt2live: gah, github showed me the wrong diff

Half-Shot: (oops, was dropped)

t3chguy:
> Frankly the reply format sucks as is, because I can't actually distinguish what is reply text and what is source text without running a regex through it so this is going to break at some point (according to @t3chguy, we are going to be dropping the source text from body).

Yes this is expected as it is *ONLY* there for backwards-compatibility

turt2live:
I really do appreciate that a release is imminent, however I don't want to have an incomplete feature be merged. I'm okay with pinning the js-sdk to the commit for the bridge (for now) once the PR is merged.

Half-Shot:
@turt2live I've made a few changes:
- Rejoice, we now use /event to fetch events that aren't cached. PRs for those coming soon.
- Replies of replies previously formatted awfully, now we seperate out the reply from the source guff in the cache and it looks better.
- More tests

config.sample.yaml
```diff
@@ -428,3 +428,8 @@ ircService:
   # `!storepass server.name passw0rd. When a connection is made to IRC on behalf of
   # the Matrix user, this password will be sent as the server password (PASS command).
   passwordEncryptionKeyPath: "passkey.pem"
+
+  # Config for Matrix -> IRC bridging
+  matrixHandler:
+    # Cache this many matrix events in memory to be used for m.relates_to messages.
```
turt2live: minor thing: we should probably mention that `m.relates_to` is effectively replies somewhere

lib/bridge/MatrixHandler.js
```diff
@@ -1256,16 +1264,39 @@ MatrixHandler.prototype._onMessage = Promise.coroutine(function*(req, event) {
...
+        }
+    }
+    this._eventCache.set(event.event_id, {
+        body: cacheBody.substr(0, REPLY_SOURCE_MAX_LENGTH),
+        sender: event.sender
+    });
+
+    /**
+     * Cache events in here so we can refer to them for replies.
+    */
```
turt2live: That's a large comment for a single line

lib/bridge/MatrixHandler.js
```diff
@@ -1486,6 +1517,89 @@ MatrixHandler.prototype._onUserQuery = Promise.coroutine(function*(req, userId)
     yield this.ircBridge.getMatrixUser(ircUser);
 });
 
+MatrixHandler.prototype._textForReplyEvent = Promise.coroutine(function*(event, ircRoom) {
+    const REPLY_REGEX = /> <(@.*:.*)>(.*)\n\n(.*)/;
```
turt2live: I'm not sure if we actually need to perform this regex at all - the existence of `m.relates_to/m.in_reply_to` should be enough for us to consider it a reply, regardless of how silly the sender's client is.

lib/bridge/MatrixHandler.js
```diff
@@ -1486,6 +1517,89 @@ MatrixHandler.prototype._onUserQuery = Promise.coroutine(function*(req, userId)
...
+        rplSource = this._eventCache.get(eventId).body;
+    }
+
+    // Get the first non-blank line from the source.
+    const lines = rplSource.split('\n').filter((line) => !/^\s*$/.test(line))
+    if (lines.length > 0) {
+        // Cut to a maximum length.
+        rplSource = lines[0].substr(0, REPLY_SOURCE_MAX_LENGTH);
+        // Ellipsis if needed.
+        if (rplSource.length > REPLY_SOURCE_MAX_LENGTH) {
```
turt2live: This condition won't be activated because the instruction directly above trims it to a safe length.

lib/bridge/MatrixHandler.js
```diff
@@ -1486,6 +1517,89 @@ MatrixHandler.prototype._onUserQuery = Promise.coroutine(function*(req, userId)
...
+            rplSource = rplSource + "...";
+        }
+        // Wrap in formatting
+        rplSource = ` "${lines[0]}"`;
```
turt2live: this throws away the trimmed body

lib/bridge/MatrixHandler.js
```diff
@@ -1486,6 +1517,89 @@ MatrixHandler.prototype._onUserQuery = Promise.coroutine(function*(req, userId)
...
+        rplSource = "";
+    }
+
+    // Fetch the sender's IRC nick.
+    const sourceClient = this.ircBridge.getIrcUserFromCache(ircRoom.server, rplName);
+    if (sourceClient) {
+        rplName = sourceClient.nick;
+    }
+    else {
+        // Somehow we failed, so fallback to userid.
```
turt2live: s/userid/localpart (minor thing)

lib/bridge/MatrixHandler.js
turt2live: nevermind, we need the actual text the user sent... I wonder if it's worth using the html instead as one common thing amoung reply implementers is mixing up how the fallback actually works.

spec/integ/matrix-to-irc.spec.js
```diff
@@ -240,6 +240,165 @@ describe("Matrix-to-IRC message bridging", function() {
...
+                done();
+            });
+
+            env.mockAppService._trigger("type:m.room.message", {
+                content: {
+                    body: "> <@somedude:bar.com> This is the fake message\n\nReply Text",
+                    msgtype: "m.text",
+                    "m.relates_to": {
+                        "m.in_reply_to": {
+                        "event_id": "$original:bar.com"
```
turt2live: indent

spec/integ/matrix-to-irc.spec.js
```diff
@@ -240,6 +240,165 @@ describe("Matrix-to-IRC message bridging", function() {
...
+                expect(channel).toEqual(roomMapping.channel);
+                expect(text).toEqual('<friend "This"> Reply Text');
+                done();
+            });
+
+            env.mockAppService._trigger("type:m.room.message", {
+                content: {
+                    body: "> <@somedude:bar.com> This is the fake message\n\nReply Text",
+                    msgtype: "m.text",
+                    "m.relates_to": {
```
turt2live: technically this event isn't spec compliant because it doesn't populate the html component (which through this review I'm starting to think is a better idea to use anyways). Spec coming Soon™️

spec/integ/matrix-to-irc.spec.js
```diff
@@ -240,6 +240,165 @@ describe("Matrix-to-IRC message bridging", function() {
...
+            done();
+        });
+
+        env.mockAppService._trigger("type:m.room.message", {
+            content: {
+                body: "> <@somedude:bar.com> This message is possibly fake\n\nReply Text",
+                msgtype: "m.text",
+                "m.relates_to": {
+                    "m.in_reply_to": {
+                    "event_id": "$original:bar.com"
```
turt2live: indent

spec/integ/matrix-to-irc.spec.js
turt2live: applies to other tests too

spec/integ/matrix-to-irc.spec.js
```diff
@@ -240,6 +240,165 @@ describe("Matrix-to-IRC message bridging", function() {
...
+            event_id: "$first:bar.com",
+            type: "m.room.message"
+        }).then(() => {
+            return env.mockAppService._trigger("type:m.room.message", {
+                content: {
+                    body: "> <@friend:bar.com> Message#1\n\nMessage #2",
+                    msgtype: "m.text",
+                    "m.relates_to": {
+                        "m.in_reply_to": {
+                          "event_id": "$first:bar.com"
```
turt2live: half the indent is here, but not all of it

Half-Shot:
>This is mostly because things that implement the feature inevitably mess up the formatting on the fallback, leading to regex not matching.

My reason for not doing HTML is because I'd have to strip away all the HTML (and also parse the HTML). I'm fairly confident this will be fine on Riot, and really this is something that needs speccing asap. I'd **really** love it if we either dropped the source text & sender from the body, or put the reply in it's own key. 

Going to fix the other bits now!

turt2live:
yea, that's all fair. Having the reply text somewhere in the event body without having to parse it out would be nice - have filed https://github.com/matrix-org/matrix-doc/issues/1541

lib/bridge/MatrixHandler.js
Half-Shot: oh wow, that's pretty silly

Half-Shot:
Also, looks like my replies totally break M->I formatting so I will be fixing that.

Half-Shot:
TACO MERGE

turt2live:
ftr if you have to pin the dependencies to get the bridge sdk all happy and working, by all means do so. Would be nice to have it in a bridge sdk release that coincides with an IRC bridge release though :)

Half-Shot:
Honestly once the js-sdk release goes I might bump both.

PR #650, opened 2018-08-14
Replace nicks with user pill mentions

### LINKED ISSUE #573: Support notifications in Matrix based on your IRC nick
Currently you have to manually set up keywords to alert you to people mentioning your current IRC nick.  It'd be nice if this happened automatically somehow (either with extensible profiles or custom keyword notifs of some kind).

Half-Shot:
Which should hopefully ping the right user. This PR will fetch a list of nick->userIds based off the collection of clients connected and try to match on that. Includes a free set of unit tests.

Fixes #573 and makes @ara4n happy.

### DIFF
--- a/lib/bridge/IrcHandler.js
+++ b/lib/bridge/IrcHandler.js
@@ -468,7 +468,7 @@ IrcHandler.prototype.onMessage = Promise.coroutine(function*(req, server, fromUs
     );
 
     let mxAction = MatrixAction.fromIrcAction(action);
-    action.formatMentions(
+    mxAction.formatMentions(
         this.ircBridge.getClientPool().getNickUserIdMappingForChannel(server, channel)
     );
 
--- a/lib/bridge/IrcHandler.js
+++ b/lib/bridge/IrcHandler.js
@@ -21,6 +21,8 @@ const MODES_TO_WATCH = [
     "s"
 ];
 
+const NICK_USERID_CACHE_MAX = 512;
+
 function IrcHandler(ircBridge) {
     this.ircBridge = ircBridge;
     // maintain a map of which user ID is in which PM room, so we know if we
@@ -50,6 +52,8 @@ function IrcHandler(ircBridge) {
         //'$fromUserId $toUserId' : Promise
     };
 
+    this.nickUserIdMapCache = new Map(); // server:channel => mapping
+    
     // Map<string,bool> which contains nicks we know have been registered/has display name
     this._registeredNicks = Object.create(null);
     this.getMetrics();
@@ -468,8 +472,22 @@ IrcHandler.prototype.onMessage = Promise.coroutine(function*(req, server, fromUs
     );
 
     let mxAction = MatrixAction.fromIrcAction(action);
+    let mapping;
+    if (this.nickUserIdMapCache.has(`${server.domain}:${channel}`)) {
+        mapping = this.nickUserIdMapCache.get(`${server.domain}:${channel}`);
+    }
+    else {
+        mapping = this.ircBridge.getClientPool().getNickUserIdMappingForChannel(
+            server, channel
+        );
+        this.nickUserIdMapCache.set(`${server.domain}:${channel}`, mapping);
+        if (this.nickUserIdMapCache.size > NICK_USERID_CACHE_MAX) {
+            this.nickUserIdMapCache.delete(this.nickUserIdMapCache.keys()[0]);
+        }
+    }
+
     mxAction.formatMentions(
-        this.ircBridge.getClientPool().getNickUserIdMappingForChannel(server, channel)
+        mapping
     );
 
     if (!mxAction) {
@@ -521,6 +539,8 @@ IrcHandler.prototype.onJoin = Promise.coroutine(function*(req, server, joiningUs
     else { // Let's avoid any surprises
         this.incrementMetric("join");
     }
+
+    this._invalidateNickUserIdMap(server, chan);
 
     let nick = joiningUser.nick;
     let syncType = kind === "names" ? "initial" : "incremental";
@@ -681,6 +701,7 @@ IrcHandler.prototype.onKick = Promise.coroutine(function*(req, server,
  */
 IrcHandler.prototype.onPart = Promise.coroutine(function*(req, server, leavingUser, chan, kind) {
     this.incrementMetric("part");
+    this._invalidateNickUserIdMap(server, chan);
     // parts are always incremental (only NAMES are initial)
     if (!server.shouldSyncMembershipToMatrix("incremental", chan)) {
         req.log.info("Server doesn't mirror parts.");
@@ -1005,6 +1026,10 @@ IrcHandler.prototype._setMatrixRoomAsInviteOnly = function(room, isInviteOnly) {
         }, ""
     );
 };
+
+IrcHandler.prototype._invalidateNickUserIdMap = function(server, channel) {
+    this.nickUserIdMapCache.delete(`${server.domain}:${channel}`);
+}
 
 IrcHandler.prototype.incrementMetric = function(metric) {
     if (this._callCountMetrics[metric] === undefined) {
--- a/lib/bridge/IrcHandler.js
+++ b/lib/bridge/IrcHandler.js
@@ -53,7 +53,7 @@ function IrcHandler(ircBridge) {
     };
 
     this.nickUserIdMapCache = new Map(); // server:channel => mapping
-    
+
     // Map<string,bool> which contains nicks we know have been registered/has display name
     this._registeredNicks = Object.create(null);
     this.getMetrics();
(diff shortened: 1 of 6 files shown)


lib/models/MatrixAction.js
```diff
@@ -26,6 +30,46 @@ function MatrixAction(type, text, htmlText, timestamp) {
     this.htmlText = htmlText;
     this.ts = timestamp || 0;
 }
+
+MatrixAction.prototype.formatMentions = function(nickUserIdMap) {
+    // Get people this message could be mentioning.
```
Half-Shot: This is wholly unhelpful.

lib/irc/ClientPool.js
```diff
@@ -248,6 +248,17 @@ ClientPool.prototype.totalReconnectsWaiting = function (serverDomain) {
     return 0;
 }
 
+ClientPool.prototype.getNickUserIdMappingForChannel = function(server, channel) {
```
turt2live: How fast is this function on a channel with tens of thousands of nicks? This seems like a good thing to have a live cache for.

lib/models/MatrixAction.js
```diff
@@ -26,6 +30,47 @@ function MatrixAction(type, text, htmlText, timestamp) {
...
+
+MatrixAction.prototype.formatMentions = function(nickUserIdMap) {
+    const regexString = "(" +
+        Object.keys(nickUserIdMap).map((value) => escapeStringRegexp(value)).join("|")
+        + ")";
+    const userRegex = new RegExp(regexString, "igm");
+    const matched = new Set(); // lowercased nicknames we have matched already.
+    let match;
+    let i = 0;
+    while ((match = userRegex.exec(this.text)) !== null && i < MAX_MATCHES) {
```
turt2live: It'd be nice to have this loop not require so much setup. A `for` loop could work too:

```
for (let i = 0; i < MAX_MATCHES && (match = userRegex.exec(this.text)) !== null) {
```

lib/models/MatrixAction.js
```diff
@@ -26,6 +30,47 @@ function MatrixAction(type, text, htmlText, timestamp) {
...
+        i++;
+        let matchName = match[1];
+        // Deliberately have a minimum length to match on, so we don't match smaller nicks accidentally.
+        if (matchName.length < MIN_LENGTH_TO_MATCH || matched.has(matchName.toLowerCase())) {
+            continue;
+        }
+        let userId = nickUserIdMap[matchName];
+        if (userId === undefined) {
+            // We might need to search case-insensitive.
+            const nicks = Object.keys(nickUserIdMap).filter((nick) =>
```
turt2live: can't we just do `userId = nickUserIdMap[matchName.toLowerCase()];` ?

lib/models/MatrixAction.js
```diff
@@ -26,6 +30,47 @@ function MatrixAction(type, text, htmlText, timestamp) {
...
+            );
+            if (nicks.length === 0) {
+                continue;
+            }
+            userId = nickUserIdMap[nicks[0]];
+            matchName = nicks[0];
+        }
+        // If this is not HTML text already, make it so.
+        if (this.htmlText === undefined) {
+            this.htmlText = this.text;
```
turt2live: This is unsafe

spec/unit/MatrixAction.spec.js
```diff
@@ -0,0 +1,57 @@
+"use strict";
+const MatrixAction = require("../../lib/models/MatrixAction");
+
+describe("MatrixAction", function() {
+    it("should not highlight mentions to text without mentions", () => {
+        let action = new MatrixAction("message", "Some text", "Some text", null);
+        action.formatMentions({
+            "Some Person": "@foobar:localhost"
+        });
+        expect(action.text).toEqual("Some text");
```
turt2live: it's also expected that the function doesn't needlessly generate an HTML component.

spec/unit/MatrixAction.spec.js
```diff
@@ -0,0 +1,57 @@
...
+    });
+    it("should highlight a user", () => {
+        let action = new MatrixAction("message", "JCDenton, it's a bomb!", "JCDenton, it's a bomb!", null);
+        action.formatMentions({
+            "JCDenton": "@jc.denton:unatco.gov"
+        });
+        expect(action.htmlText).toEqual("<a href=\"https://matrix.to/#/@jc.denton:unatco.gov\">JCDenton</a>, it's a bomb!");
```
turt2live: also verify the plain text hasn't been altered

spec/unit/MatrixAction.spec.js
```diff
@@ -0,0 +1,57 @@
...
+    });
+    it("should highlight a user, with plain text", () => {
+        let action = new MatrixAction("message", "JCDenton, it's a bomb!");
+        action.formatMentions({
+            "JCDenton": "@jc.denton:unatco.gov"
+        });
+        expect(action.htmlText).toEqual("<a href=\"https://matrix.to/#/@jc.denton:unatco.gov\">JCDenton</a>, it's a bomb!");
+    });
+    it("should highlight a user, with weird characters", () => {
+        let action = new MatrixAction("message", "JCDenton[m], it's a bomb!");
```
turt2live: just for fun, should make the nick weirder: `JCDenton|$m|` or something, plus a backtick at the beginning or end of the nick.

lib/models/MatrixAction.js
```diff
@@ -26,6 +30,47 @@ function MatrixAction(type, text, htmlText, timestamp) {
...
+        }
+        this.htmlText = this.htmlText.replace(
+            new RegExp(escapeStringRegexp(matchName), "igm"),
+            `<a href="https://matrix.to/#/${userId}">${matchName}</a>`
```
turt2live: Unsafe HTML (either userId or matchName could inject HTML)

lib/irc/ClientPool.js
Half-Shot: I'm thinking that...I guess we only need to invalidate this on join.

lib/models/MatrixAction.js
```diff
@@ -17,6 +18,9 @@ const MSGTYPE_TO_TYPE = {
     "m.file": "file"
 };
 
+const MIN_LENGTH_TO_MATCH = 4;
```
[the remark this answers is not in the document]
Half-Shot: It's kinda commented on line 45, unless you want me to duplicate it?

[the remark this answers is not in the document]
turt2live: Duplicate/improve or something. Just kinda needs a note that these are for pills.

lib/models/MatrixAction.js
Half-Shot: Nicks are case sensitive, so we'd need to maintain two maps.

lib/models/MatrixAction.js
Half-Shot: This is escaped in L115. If we think it's HTML (dangerous) then htmlText will be defined first.

spec/unit/MatrixAction.spec.js
Half-Shot: Now tested for

lib/bridge/IrcHandler.js
```diff
@@ -21,6 +21,8 @@ const MODES_TO_WATCH = [
     "s"
 ];
 
+const NICK_USERID_CACHE_MAX = 512;
```
turt2live: not worth blocking this PR on it, but it would be interesting to have metrics to see how often this cache is overrun. 512 channels for freenode doesn't sound like a lot, but might be given most of them are probably quiet.

lib/bridge/IrcHandler.js
```diff
@@ -468,6 +472,23 @@ IrcHandler.prototype.onMessage = Promise.coroutine(function*(req, server, fromUs
...
+        );
+        this.nickUserIdMapCache.set(`${server.domain}:${channel}`, mapping);
+        if (this.nickUserIdMapCache.size > NICK_USERID_CACHE_MAX) {
+            this.nickUserIdMapCache.delete(this.nickUserIdMapCache.keys()[0]);
+        }
+    }
+
+    mxAction.formatMentions(
+        mapping
+    );
```
turt2live: this looks like a lot of extra space

lib/models/MatrixAction.js
turt2live: Oh I see now, I'm blind. Should we use `find` instead to avoid going over the entire list in most cases?

lib/models/MatrixAction.js
turt2live: The escaping is fairly far away path-wise, although I'm fine with it being commented to say why it's safe :)

lib/bridge/IrcHandler.js
Half-Shot: Given the cost of generating this list is small (relatively), I can live with it. Remember we are dropping things from this cache whenever someone joins or leaves so I'd expect it to stay fairly low.

turt2live: Yea, more of a thing where I love to see graphs for things. I don't suspect it'll actually be a problem.

Half-Shot:
Happy with this now @turt2live ? :innocent:

PR #655, opened 2018-08-20
Fix invalidchar nick

### LINKED ISSUE #654: Mysteriously some users get bridged as "M[m]"
One poor user was `@kaylons:matrix.org` who never made it onto IRC because their bridgedclient couldn't connect. However, it was trying to bridge them as M which sounds...wrong.

Half-Shot:
Fixes #654. This checks for invalid characters much earlier during templating and throws more aggressively if the nickname looks wrong. This change comes with a nice set of unit tests, too.

### DIFF
--- a/lib/bridge/MatrixHandler.js
+++ b/lib/bridge/MatrixHandler.js
@@ -911,7 +911,7 @@ MatrixHandler.prototype._onJoin = Promise.coroutine(function*(req, event, user) 
                         bridgedClient.changeNick(newNick, false);
                     }
                     catch (e) {
-                        req.log.warn(`Didn't change nick on the IRC side:${e}`);
+                        req.log.warn(`Didn't change nick on the IRC side: ${e}`);
                     }
                 }
             }
--- a/lib/bridge/MatrixHandler.js
+++ b/lib/bridge/MatrixHandler.js
@@ -904,9 +904,15 @@ MatrixHandler.prototype._onJoin = Promise.coroutine(function*(req, event, user) 
                 if (room.server.allowsNickChanges() &&
                     config.getDesiredNick() === null
                 ) {
-                    bridgedClient.changeNick(
-                        room.server.getNick(bridgedClient.userId, event.content.displayname),
-                        false);
+                    try {
+                        const newNick = room.server.getNick(
+                            bridgedClient.userId, event.content.displayname
+                        );
+                        bridgedClient.changeNick(newNick, false);
+                    }
+                    catch (e) {
+                        req.log.warn(`Didn't change nick on the IRC side:${e}`);
+                    }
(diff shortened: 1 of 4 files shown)


lib/bridge/MatrixHandler.js
```diff
@@ -904,9 +904,15 @@ MatrixHandler.prototype._onJoin = Promise.coroutine(function*(req, event, user)
...
-                        room.server.getNick(bridgedClient.userId, event.content.displayname),
-                        false);
+                    try {
+                        const newNick = room.server.getNick(
+                            bridgedClient.userId, event.content.displayname
+                        );
+                        bridgedClient.changeNick(newNick, false);
+                    }
+                    catch (e) {
+                        req.log.warn(`Didn't change nick on the IRC side:${e}`);
```
turt2live: can has space after `:`?

lib/irc/IrcServer.js
```diff
@@ -433,10 +434,16 @@ IrcServer.prototype.getAliasFromChannel = function(channel) {
...
 IrcServer.prototype.getNick = function(userId, displayName) {
-    var localpart = userId.substring(1).split(":")[0];
-    var display = displayName || localpart;
-    var template = this.config.ircClients.nickTemplate;
-    var nick = template.replace(/\$USERID/g, userId);
+    const illegalChars = BridgedClient.illegalCharactersRegex;
+    let localpart = userId.substring(1).split(":")[0];
+    localpart = localpart.replace(illegalChars, "");
+    displayName = displayName ? displayName.replace(illegalChars, "") : undefined;
+    let display = [displayName, localpart].find((n) => Boolean(n));
```
turt2live: This is a neat trick that I didn't think of. Can it be a constant though? :P

spec/unit/IrcServer.spec.js
```diff
@@ -0,0 +1,50 @@
...
+            );
+            expect(server.getNick("@💩ケ:foobar", "wiggle")).toBe("M-wiggle");
+        });
+        it("should throw if no characters could be used", function() {
+            const server = new IrcServer("irc.foobar",
+                extend(true, IrcServer.DEFAULT_CONFIG, {})
+            );
+            expect(() => {server.getNick("@💩ケ:foobar", "💩ケ")}).toThrow();
+        });
+    });
```
turt2live: should also have a test for lack of display name (ie: null/undefined)

Half-Shot: Isn't that covered by the first test?

Half-Shot: Oh, you mean invalid userid only.

Half-Shot:
*reassigns*

PR #656, opened 2018-08-22
QueuePool.waitingItems should use it's internal queue size

Half-Shot:
I suspect somehow we were racing which made the Freenode graphs have a deficit of 1.5k reconnecting users, which I believe do not exist. This PR adds some tests to ensure the sizes are correct, and that QueuePool uses Queue.size() to determine waiting items.

### DIFF
--- a/spec/unit/Queue.spec.js
+++ b/spec/unit/Queue.spec.js
@@ -115,4 +115,25 @@ describe("Queue", function() {
             done();
         });
     });
+
+    it("should have the correct size.", (done) => {
+        var thing1 = { foo: "bar"};
+        var thing2 = { bar: "baz"};
+        var things = [thing1, thing2];
+        let expectedSize = things.length;
+        procFn.and.callFake((thing) => {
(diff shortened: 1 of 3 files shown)


spec/unit/Queue.spec.js
```diff
@@ -115,4 +115,25 @@ describe("Queue", function() {
             done();
         });
     });
+
+    it("should have the correct size.", (done) => {
```
turt2live: why does this test get a full stop and the other's don't?

spec/unit/QueuePool.spec.js
```diff
@@ -135,4 +135,16 @@ describe("QueuePool", function() {
         yield nextTick();
         expect(Object.keys(itemToDeferMap).sort()).toEqual(["b"]);
     }));
+
+    it("should accurately track waiting items", test.coroutine(function*() {
+        for (let i = 0;i<10;i++) {
```
turt2live: so squishy

PR #658, opened 2018-08-22
Use localpart in plaintext so we get highlighted for pills

Half-Shot:


lib/models/MatrixAction.js
```diff
@@ -64,10 +64,18 @@ MatrixAction.prototype.formatMentions = function(nickUserIdMap) {
             this.htmlText = this.text;
         }
         userId = ircFormatting.escapeHtmlChars(userId);
+        /* Due to how Riot and friends do push notifications, we need the plain text to match something.
+           Modern client's will pill-ify and will show a displayname, so we can use the localpart so it
+           matches  and doesn't degrade too badly on limited clients.*/
```
turt2live: minor: double space after `matches`

lib/models/MatrixAction.js
```diff
@@ -64,10 +64,18 @@ MatrixAction.prototype.formatMentions = function(nickUserIdMap) {
...
+        const localpart = userId.substr(1, userId.indexOf(":")-1);
```
turt2live: we should be using the display name, as per https://github.com/matrix-org/matrix-doc/pull/1547

Half-Shot: You *really* want to do a displayname lookup for every match? The localparts already match.

turt2live: I'd prefer it if the cached display name was used, given we already know about the member event.

Half-Shot:
>We already pull members out of the room for the purpose of figuring out mentions

We don't? We could, but we don't currently.

lib/models/MatrixAction.js
```diff
@@ -64,14 +66,31 @@ MatrixAction.prototype.formatMentions = function(nickUserIdMap) {
...
+        }
+
+        if (identifier === undefined) {
+            // Fallback to userid.
+            identifier = userId.substr(1, userId.indexOf(":")-1)
+        }
+
+        const regex = MentionRegex(escapeStringRegexp(matchName));
+        this.htmlText = this.htmlText.replace(regex,
+`$1<a href="https://matrix.to/#/${userId}">${ircFormatting.escapeHtmlChars(identifier)}</a>`
```
turt2live: this should be indented

spec/unit/MatrixAction.spec.js
```diff
@@ -87,10 +88,40 @@ describe("MatrixAction", function() {
...
+            "Go to <a href='http://JCDenton.com'>my website</a>",
+            null
+        );
+        action.formatMentions({
+            "JCDenton": "@jc.denton:unatco.gov"
+        });
+        expect(action.text).toEqual("Go to http://JCDenton.com");
+        expect(action.htmlText).toEqual(
+            "Go to <a href='http://JCDenton.com'>my website</a>"
         );
```
turt2live: a test for making sure it also actually uses the profile where possible would be nice

PR #659, opened 2018-08-23
IRC Memberstate tracking

Half-Shot:
This is an experimental PR to leave users who we don't expect  to be in the room. It's pretty simple:
- We track all joins in a big ``Set``
- Any users not in that set are left after they send a message.
- This hopefully catches the case where notice bots are not left in rooms until restart, or users who have left and we've decided to keep around.

This comes with the pitfall that if a join is not caught (either via a join or names call), then that user will be stuck in limbo until someone else from matrix joins and forces a new names set containing the user. There is definitely a non zero chance of this happening, but it's *rare* for us to outright miss a join unless we were falling over in traffic which would probably be bridge death anyway.

This solves the problems where:
- Users aren't actually in the channel, but sending messages into it causing the member to be stuck there forever.
- This also solves the problem of /parts arriving ahead of the full set of messages being sent, and leaving the user stuck again. However it will also come with a bundle of join/leaves which probably means we need to set some delays up.

### DIFF
--- a/lib/bridge/IrcHandler.js
+++ b/lib/bridge/IrcHandler.js
@@ -547,14 +547,15 @@ IrcHandler.prototype.onMessage = Promise.coroutine(function*(req, server, fromUs
 });
 
 IrcHandler.prototype._incrementResyncChannelCounter = function(server, channel, req) {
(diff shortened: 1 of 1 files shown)


Half-Shot:
I think this has sufficiently bitrotted, and we've not really found a dying need to go back to this PR.

PR #661, opened 2018-08-24
Set topics with ghosts

### LINKED ISSUE #618: When the topic is changed, it doesn't tell you who made the change
Feels a bit weird that the bot always sets the topic.

Half-Shot:
Fixes #618

### DIFF
--- a/lib/bridge/IrcHandler.js
+++ b/lib/bridge/IrcHandler.js
@@ -397,8 +397,10 @@ IrcHandler.prototype._serviceTopicQueue = Promise.coroutine(function*(item) {
             },
             (err) => {
                 item.req.log.error(`Error storing room ${entry.matrix.getId()} (${err.message})`);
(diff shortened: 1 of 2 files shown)


lib/bridge/IrcHandler.js
```diff
@@ -382,18 +382,22 @@ IrcHandler.prototype._serviceTopicQueue = Promise.coroutine(function*(item) {
...
+        }).then(
             () => {
                 entry.matrix.topic = item.topic;
                 return this.ircBridge.getStore().upsertRoomStoreEntry(entry);
             },
             (err) => {
                 item.req.log.error(`Error storing room ${entry.matrix.getId()} (${err.message})`);
-            }
-        );
+        });
```
turt2live: personally I'm more of a fan of having the curly on it's own line so it lines up nicely

PR #669, opened 2018-08-29
Double bridging rule files

Half-Shot:
Depends on: https://github.com/matrix-org/matrix-appservice-bridge/pull/85

This change allows you to do rule specifying on what is allowed to be provisioned to avoid some cases of double bridging. New provision requests would check against the rule file to see if the members in a certain room were excluded from being bridged e.g. not mixing "@snoonet-irc:matrix.org" and "@gimpnet-irc:matrix.org" in the same room. 

This doesn't yet affect existing bridged rooms, which would need us to manually intervene.

### DIFF
--- a/rules.sample.yaml
+++ b/rules.sample.yaml
@@ -1,0 +1,10 @@
+# A set of regexes to match against joined members in rooms.
+# If one of the regexes matches a userId, then do not allow provisioning
+# to the room UNLESS it also matches a exempt regex.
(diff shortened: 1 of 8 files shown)


rules.sample.yaml
```diff
@@ -0,0 +1,10 @@
+# A set of regexes to match against joined members in rooms.
```
turt2live: This file should probably have a more descriptive name, like 'provisioning.rules.sample.yaml'.

PR #695, opened 2018-09-26
Correct spelling.

m00dimus:


Half-Shot:
(oops, we could probably define a language if people care)

Issue #687, opened 2018-09-12
I->M messages fail consistently after some uptime


Valodim:
After some uptime of the bridge (I don't know exactly, maybe one to three days) it seems to go into a state where [I->M] messages fail consistently, like this:

```
2018-09-12 03:26:56 INFO:req [92kmoegqu5s00] [I->M] onMessage: irc.freenode.net from=redacted (null@irc.freenode.net) to=#channel action={"type":"message","text":"redacted","ts":0}
2018-09-12 03:26:56 INFO:IrcBridge [92kmoegqu5s00] [I->M] FAILED  (2ms)
```

Joins and leaves still work I->M, and all of M->I also still works. But onMessage fails 100% of the time, until I restart the bridge then it works again for a while. There is nothing suspicious in synapse's log around those timestamps, at least not on log level WARN.

This is on 0.11.1, and I haven't had those issues until recently (couple weeks ago). I can't say exactly if this correlates with the 0.11.1 update from 0.11.0, but I'm positive it wasn't an issue at least on 0.9.1 (I think I skipped 0.10.1). Happy to provide more info or logs if that helps.

turt2live:
@Half-Shot This looks like a repeat of the instant-failure thing we saw before. Didn't a patch land for that?

Half-Shot:
Yes, but not a released patch. Valodim is rightly using 0.11.1 which is also unpatched. We should really do a 0.11.2 release.

Valodim:
Yes, please. I'd rather not put an enterprise loop around the bridge :)

Valodim:
...ping?

Valodim:
can confirm this is fixed :)

PR #698, opened 2018-10-05
Release v0.11.2

Half-Shot:


config.sample.yaml
```diff
@@ -439,4 +439,19 @@ ircService:
...
     # Cache this many matrix events in memory to be used for m.relates_to messages (usually replies).
-    eventCacheSize: 4096
\ No newline at end of file
+    eventCacheSize: 4096
+
+  ircHandler:
+    # How many /leave requests can be ongoing at a time.
+    # This is used to stem the flow of requests in case of a mass quit/leave, which might
+    # slow down the homeserver.
+    leaveConcurrency: 10
```
Valodim: no mention in the changelog?

lib/config/schema.yml
```diff
@@ -1,6 +1,11 @@
 "$schema": "http://json-schema.org/draft-04/schema#"
 type: "object"
 properties:
+    advanced:
```
Valodim: what about ircHandler->leaveConcurrency?

lib/irc/ClientPool.js
```diff
@@ -251,8 +251,9 @@ ClientPool.prototype.totalReconnectsWaiting = function (serverDomain) {
 ClientPool.prototype.updateActiveConnectionMetrics = function(server, ageCounter) {
     const clients = Object.values(this._virtualClients[server].userIds);
     clients.forEach((bridgedClient) => {
-        if (bridgedClient.isDead()) {
-            return; // We don't want to include dead ones.
+        if (!bridgedClient || bridgedClient.isDead()) {
+            // We don't want to include dead ones, or ones that don't exist.
```
Valodim:
```
var isNonexistentOrDead = !bridgedClient || bridgedClient.isDead()
if (isNonexistentOrDead) {
```

same amount of lines, expresses semantics in code instead of comments

config.sample.yaml
Half-Shot: Good catch, this is a develop feature atm.

lib/config/schema.yml
Half-Shot: Same as above.

lib/irc/ClientPool.js
Half-Shot: Good point, I'll include this in the next release since it's non-critical (and saves us having to apply it to two branches) but definitely reads better.

Half-Shot:
>You mentioned there's a fix for #687, is that actually in here?

Should be fixed by the null checking, so I'll add a statement to that effect in the changelogs.

Valodim:
That didn't actually end up in the changelog now, did it?

Half-Shot:
Oops, my bad. Sorry :(

PR #702, opened 2018-10-22
First naive Dockerfile for this bridge

kaiyou:
This PR provides a first implementation of a Dockerfile for this bridge. It is quite naive and should be improved upon, especially regarding the image layout and size; this probably needs a multi-stage build.

This is currently automatically built by the Docker Hub and available as ``tedomum/matrix-appservice-irc``.

PR #713, opened 2018-12-25
Provide a Dockerfile for the bridge

kaiyou:
This PR provides a first implementation of a Dockerfile for this bridge.

It replaces #702.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,33 @@
+# Builder
+FROM node:6-slim as builder
+
+RUN apt-get update \
+ && apt-get install -y git python build-essential libicu-dev
+
+RUN git clone https://github.com/matrix-org/freebindfree.git \
+ && cd freebindfree \
+ && make
+
+COPY ./package.json ./package.json
+RUN npm install
+
+# App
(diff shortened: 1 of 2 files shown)


kaiyou:
Still missing some documentation, I hope I can provide something in the next few days.

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+RUN git clone https://github.com/matrix-org/freebindfree.git \
+ && cd freebindfree \
+ && make
+
+COPY ./package.json ./package.json
+RUN npm install
+
+# App
+FROM node:6-slim
```
Half-Shot: Still not sure why we're targetting node 6? Node 10 is definitely going to yield better perf

simonszu:
Thank you so much for your Dockerfile. However, since it is a little unclear for me how to use the container, i want to note that a modified Readme would look good in this PR as well ;)

xNephe:
Thank you for the dockerfile, it's really useful !
However, I see no volume definition for the databases and the configuration files. From my understanding, `/config` and `/app/data` should be declared as volumes.

PR #721, opened 2019-01-26
Add Readme to scripts folder and some options to remove-idle-users.py

rubo77:
I added a Readme to the scripts folder and enhanced the comments in `remove-idle-users.py` and some more options:

## changes in `remove-idle-users.py`:

- Add description in comment and clarify help
- Add simulate and verbose option
- added some more validation to give the right error message, if an argument is wrong.

Signed-off-by: Ruben Barkow <github@r.z11.de>

PR #722, opened 2019-01-26
repair prefix selection and add simulate and verbose option

rubo77:
The function `should_ignore_user() ignored all users **with** that user_prefix instead of all users **without**! This is fixed now.

Also I added some more validation to give the right error message, if an argument is wrong.


rubo77:
I heard, the script was initionally meant for only kicking idle non-IRC Matrix Users from a Matrix room with a bridge, so those that are not visible online in IRC anymore are also kicked from the matrix room, ... hmmmm.... 

So it seems, I created a new work-case here now.

I will rename the script to `kick-inactive-irc-users.py` then

... WIP

rubo77:
I added this in a correct way to #721

PR #737, opened 2019-03-18
Fix getting the wrong userid for a PM

Half-Shot:


Half-Shot:
Ignore the failures, coverage is still running out of memory

Issue #442, opened 2017-05-25
IRC nickserv password characters filtered leads to auth failure


alick:
When I try to authenticate to freenode via NickServ, it seems that some special characters in the password are filtered, which leads to failed authentication. Shouldn't we accept the password as it is (with injection checks for safety)? Or is there a way to escape special characters?

Kegsay:
afaik messages are sent verbatim - I suspect this may be a markdown problem e.g. if your pass was `foo*bar*` then the bridge will send "bar" in italics with the corresponding IRC formatting codes.

Without further information on your end I cannot provide more help. If you wish to speak to me privately, I'm `@kegan:matrix.org`.

alick:
@Kegsay I think you are right on suspecting markdown, e.g. `*`, `_`.

Mikaela:
I think this might be more of upstream issue in the client where the solution is to disable markdown while entering the password.

14mRh4X0r:
According to the source, the bot picks `event.content.body` instead of some parsed version of `event.content.formatted_body`, and at least Riot sends the markdown source in the plain body.

Kegsay:
This is only true for `!commands` in the admin room. Normal PMs to NickServ are parsed like any other message, which looks at the formatted body in order to do IRC formatting.

14mRh4X0r:
Fair point, my mind was at the `!storepass` command when I wrote that.

PR #738, opened 2019-03-19
Actually join !storepass password with spaces

14mRh4X0r:
There's a comment at `!storepass` about allowing spaces in passwords, but the password doesn't actually get joined with spaces.
~~This might solve #442.~~

### DIFF
--- a/lib/bridge/MatrixHandler.js
+++ b/lib/bridge/MatrixHandler.js
@@ -536,7 +536,7 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
 
         try {
             // Allow passwords with spaces
-            let pass = args.join('');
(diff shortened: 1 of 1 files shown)


Half-Shot:
Thanks @14mRh4X0r :). Can you [sign off](https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.rst#sign-off) for us please?

14mRh4X0r:
I knew I was forgetting something.

14mRh4X0r:
Perhaps `CONTRIBUTING.md` should be amended to clarify synapse's contribution guidelines also apply here.

Half-Shot:
>Perhaps CONTRIBUTING.md should be amended to clarify synapse's contribution guidelines also apply here.

It definitely should be.

PR #750, opened 2019-05-10
Don't escape the user_ids incoming from matrix

### LINKED ISSUE #744: Slashes in display (or user?) name cause strange behaviour
image

Half-Shot:
Fixes #744

This is a problem because the bridge recently started escaping users (https://github.com/matrix-org/matrix-appservice-bridge/pull/84) which meant that the bridge was escaping genuine userids from Matrix.


Half-Shot:
(coverage exploded due to memory shortages, but the tests otherwise passed)

ara4n:
looks plausible

Issue #637, opened 2018-08-06
We don't retry kicks I->M


Half-Shot:
Which is dangerous because it leaves them in the room, and able to read messages and send messages to Matrix users inside.

Half-Shot:
This is fixed by #754, waiting on a deployment.

Issue #460, opened 2017-06-21
The bridge doesn't respect REMOVE (aka forced PART)


Mikaela:
`REMOVE` is a command to force user to `PART` the channel, I think it's designed for cases where the user is automatically rejoining on kick.

When I remove a Matrix user, I would expect them to be removed on the Matrix side too, but instead they silently stay in the room to rejoin when they say something next time.

```
<-- mijris[m] (mijrischat@gateway/shell/matrix.org/x-ykfippqwepiopoqo) has left ##Mikaela (requested by Mikaela (test#?))
<@Michaela> !ping
--> mijris[m] (@mijris:chat.weho.st) (mijrischat@gateway/shell/matrix.org/x-ykfippqwepiopoqo) has joined ##Mikaela
mijris[m] (Notice) pong! 2017-06-21 05:33:18
```

* * * * *

Addition: Wednesday, 21 June 2017 10:54 (UTC+0300)

It appears that REMOVE isn't documented almost anywhere, but it does exist at freenode. The only documentation that I can find is [Charybdis extensions README](https://github.com/charybdis-ircd/charybdis/blob/release/3.5/extensions/README#L44-L47) (thanks @Spydar007 for linking me to it) which only says:

```

m_remove.c     - REMOVE - Implements /REMOVE, which allows a channel op to
                          force-part a problematic user from a channel.  Most
                          likely, you don't want this, because it is non-standard,
                          and easily worked around for autojoin purposes.
```

Spydar007:
FTR, the syntax for REMOVE is `/quote REMOVE #channel nick :reason`. You only need include the colon at the start on the reason if your reason is more than a single word.

spb:
Note that this appears to the removed client as a simple PART message originating from the target, the same as when a client-initiated PART is reflected to the client that issued it.

Note also that this issue is more fundamental than not understanding REMOVE. There are a number of other scenarios, across most IRCd software, in which a user will PART, or JOIN, a channel without being initiated by that client; it's important that any client software understands that as part of basic IRC state tracking. Clients that fail to handle that situation gracefully have resulted in a great deal of annoyance and channel spam over the years, and usually end up getting banned as a result.

Half-Shot:
I am going to remove the p1 tag and I don't believe supporting a barely documented extension is worth doing for the bridge. We are working on a PR which honors PARTs aimed at Matrix-side users and kicks them appropriately.

Half-Shot:
[...]

PR #754, opened 2019-05-22
Handle IRC PARTs for Matrix users.

### LINKED ISSUE #743: `!cmd chat.freenode.net JOIN 0` keeps the Matrix user in all rooms while being invisible from IRC
1. join freenode bridged rooms
2. open direct chat with `@appservice-irc:matrix.org`
3. run `!cmd chat.freenode.net JOIN 0`

**Expected result**: I get kicked from all freenode bridged rooms as if I had ran `!quit` (while possibly still keeping my ghost connected for direct chats)

**What happens instead**: I stay in all the rooms and can watch what is being chatted while the IRC users cannot see me. I get the impression that the agreement with freenode requires all Matrix users who can see discussions to be visible on the channel on the freenode side.

When I again say something, I am just JOINed to the channel for messages to get through normally.

Half-Shot:
Fixes #743, #637, #460 

This PR changes the behaviour of PART handling so that we now act on PARTs directed at Matrix users. These parts are only handled if the `kind == "part"`.

I am unsure if this is the best approach, as I am worried that any kind of splutter from the IRC side might end up mass parting a lot of users from a bridged Matrix room. On the other hand, a room where the IRC user doesn't have power to kick won't be able to kick users who have left the channel.

However, I don't see another way to handle the situation in #743, besides some kind of behaviour guessing based off commands sent to the bridge.

### DIFF
--- a/lib/bridge/IrcHandler.js
+++ b/lib/bridge/IrcHandler.js
@@ -747,8 +747,7 @@ IrcHandler.prototype.onKick = Promise.coroutine(function*(req, server,
         yield this._leaveQueue.enqueue(chan + matrixUser.getId(), {
             rooms: matrixRooms,
             userId: matrixUser.getId(),
-            shouldKick: true,
-            reason: `${kicker.nick} has kicked this user from ${chan} (${reason})`,
+            shouldKick: false,
             retry: true,
             req,
         });
--- a/lib/bridge/IrcHandler.js
+++ b/lib/bridge/IrcHandler.js
@@ -27,6 +27,7 @@ const NICK_USERID_CACHE_MAX = 512;
 const LEAVE_CONCURRENCY = 10;
 const LEAVE_DELAY_MS = 3000;
 const LEAVE_DELAY_JITTER = 5000;
+const LEAVE_MAX_ATTEMPTS = 10;
 
 function IrcHandler(ircBridge, config) {
     config = config || {}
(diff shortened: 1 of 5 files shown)


lib/bridge/IrcHandler.js
```diff
@@ -1112,6 +1130,51 @@ IrcHandler.prototype._invalidateNickUserIdMap = function(server, channel) {
...
 
+IrcHandler.prototype._handleLeaveQueue = async function(item) {
+    const bridge = this.ircBridge.getAppServiceBridge();
+    let retryRooms = [];
+    for (const room of item.rooms) {
+        const roomId = room.getId();
+        item.req.log.info(
+            "Leaving room %s (%s)",
+            roomId,
+            item.userId
```
turt2live: trailing comma (consistency)
```suggestion
            item.userId,
```

lib/bridge/IrcHandler.js
```diff
@@ -1112,6 +1130,51 @@ IrcHandler.prototype._invalidateNickUserIdMap = function(server, channel) {
...
+                item.req.log.warn("Not retrying");
+                continue;
+            }
+            retryRooms.push(room);
+        }
+        if (retryRooms.length < 0) {
+            return;
+        }
+        await Promise.delay(LEAVE_DELAY_MS + (Math.random() * LEAVE_DELAY_JITTER));
+        item.attempts++;
```
turt2live: we don't give up based on attempts: we should do that

lib/bridge/IrcHandler.js
```diff
@@ -765,7 +759,7 @@ IrcHandler.prototype.onKick = Promise.coroutine(function*(req, server,
  * @param {IrcServer} server : The sending IRC server.
  * @param {IrcUser} leavingUser : The user who parted.
  * @param {string} chan : The channel that was left.
- * @param {string} kind : The kind of part (e.g. PART, KICK, BAN,
+ * @param {string} kind : The kind of part (e.g. PART, KICK, BAN, quit
```
turt2live: english:
```suggestion
 * @param {string} kind : The kind of part (e.g. PART, KICK, BAN, quit,
```

lib/bridge/IrcHandler.js
```diff
@@ -1168,6 +1166,10 @@ IrcHandler.prototype._handleLeaveQueue = async function(item) {
         }
         await Promise.delay(LEAVE_DELAY_MS + (Math.random() * LEAVE_DELAY_JITTER));
         item.attempts++;
+        if (item.attempts === LEAVE_MAX_ATTEMPTS) {
```
turt2live: `>=` for abundance of safety

PR #763, opened 2019-06-06
Configurable room versions

### LINKED ISSUE #761: Create rooms with a pre-configured room version
The bridge is still creating v1 rooms (as is the default in synapse). However pretty much every v1 room with lots of state changes (i.e. IRC rooms) seem to end up looking quite broken. Newer room versions seem to be more resilient and are well tested, so we should at least support the option of doing this.

Half-Shot:
Fixes #761 

This allows a bridge admin to specify what version a dynamic channel will be created with.


config.sample.yaml
```diff
@@ -182,6 +182,9 @@ ircService:
         # HS attached to this AS will be able to interact with this room.
         # Default: true.
         federate: true
+        # Force this room version when creating IRC channels. Beware if the homeserver doesn't
+        # support the room version then the request will fail. By default, no version is requested.
```
turt2live: We should clarify what this means:
```suggestion
        # support the room version then the request will fail. By default, the bridge will use whatever the homeserver's default is.
```

Half-Shot:
Changes made, but merged with #766 to ensure that tests pass. Will await outcome of #766 before pushing.

PR #764, opened 2019-06-06
Bump reqirement versions and use BuildKite

### LINKED ISSUE #762: Drop support for node 6.X
Node 6 [is dead](https://nodejs.org/en/about/releases/), so we can drop support for 6 now.

Half-Shot:
Fixes #762 

This PR does 3 things:

- We now state that we support Synapse 0.99.5.2. I feel like this might be a bit harsh in that we are somewhat excluding old servers from support, but on the other hand I'm not sure where we would draw the line. I'm **fairly** certain we don't support **0.18.5-rc3** anymore though.
- Drop support for node 6 **and** 8. Node 6 is out of LTS and should be purged, as it is actively stopping us from adopting nice features. Node 8 has about 6 months left to live, so I felt we might as well go straight to 10.
- This PR adds a buildkite pipeline, because we no longer use travisci

Following the merge of 766, this PR also does:
- Tests no longer use `var`, to hopefully avoid any scoping issues.
- All imports now use `const` rather than `var`.
- The testing environment code has been refactored.

However, I'm flexible on this if we believe that supporting 8 is still a good idea.


Half-Shot:
The plan was to make this a quick thing, but our memory usage for tests is atrocious. I'm going to look at fixing that as part of this PR.

Half-Shot:
@turt2live you might have missed, but the red is because of the obscene memory usage. #766 makes red go green, which is the only way we can fix this.

PR #766, opened 2019-06-07
Reduce memory consumption of tests

Half-Shot:
Based upon #764 (needs to be merged first)

This is a PR born out of desperation to make the tests work, and I am not convinced they are perfect yet. However this PR does the following

- Bumps eslint up to 5.16.0 and sets emcaVersion to 8, so that we can use async/await.
- In an effort to prevent leakage:
  - `var ... = require("...")` becomes `const ... = require("...")` across all files
  - `var abc = ...` becomes `let abc = ...` across all files
- envBundle is used to create a testEnv, and each env is now scoped to the container `describe` to hopefully give the GC something to eat.
- `TestEnv` is a new class that replaces the previous env objects.
- `main.js`/ `IrcMock`/`MatrixClientMock` is no longer created per test, because of the huge memory cost.

This PR is incredibly invasive to the code base, so if the reviewer wishes for me to drop one of these changes then please suggest so. However, at least the tests now work!


turt2live:
removing from review until #764 lands - the changes are too intertwined for review.

Half-Shot:
Closing as changes have been merged into #764

PR #778, opened 2019-07-22
Typescript support

Half-Shot:
This makes some changes to the bridge to be able to support typescript. Unlike the slack bridge, this isn't a full refactor of the whole bridge but rather sets up the correct environment to write TS files alongside the existing JS files. Future PRs may convert more files over to Typescript.

(@turt2live I've added you to yet another ts PR, but this one is less urgent)


turt2live:
@Half-Shot ***please*** can we not continue the trend of overhauling the entire codebase all at once? It's impossible to review this.

Half-Shot:
@turt2live I've deliberately not done this. I've removed strict mode from the js files because it's pointless but otherwise I've tried to keep them the same as `develop`. I did have to shift the files to `src` to stop them being overwritten.

turt2live:
There's failing tests, and 46 file changes that need auditing. Why can we not just do this incrementally file-by file on an as-needed basis? 

Most of my concerns with this PR are mirrored on the Slack bridge PR. Let's get one bridge through before starting another.

Half-Shot:
Closing this in favor of https://github.com/matrix-org/matrix-appservice-irc/pull/808, which does the bare minimum to support TS

PR #785, opened 2019-07-30
Powerlevel handling refactor

Half-Shot:
This change refactors mode/power level handling to it's own class for convenience, as well as the hope of being able to unit test the class individually to address our most glaring power level desync issues.

This makes the following changes to behaviour:
- IRC users which are left from a room ~on startup~ (https://github.com/matrix-org/matrix-appservice-irc/pull/785/commits/6b7d23e5bf4031dd45c421c08bf870977ad49b93 makes this work for realtime leaves) have their power levels removed, in compliance with IRC standards.
- Power level events are now checked locally before they are sent to ensure a change has actually been made (to avoid needless state spam).

### DIFF
--- a/lib/bridge/IrcHandler.js
+++ b/lib/bridge/IrcHandler.js
@@ -65,6 +65,7 @@ function IrcHandler(ircBridge, config) {
     //    kickReason: string,
     //    retry: boolean,
     //    req: Request,
+    //    deop: boolean,
     //}
     this._leaveQueue = new QueuePool(
         config.leaveConcurrency || LEAVE_CONCURRENCY,
@@ -746,6 +747,7 @@ IrcHandler.prototype.onKick = Promise.coroutine(function*(req, server,
             shouldKick: false,
             retry: true,
             req,
+            deop: true, // deop real irc users, likereal irc.
         });
     }
 });
@@ -820,6 +822,7 @@ IrcHandler.prototype.onPart = Promise.coroutine(function*(req, server, leavingUs
         req,
         kickReason: "Client PARTed from channel", // this will only be used if shouldKick is true
         retry: true, // We must retry these so that membership isn't leaked.
+        deop: !leavingUser.isVirtual, // deop real irc users, likereal irc.
     });
     stats.membership(true, "part");
     yield promise;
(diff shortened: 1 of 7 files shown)


lib/bridge/IrcHandler.js
```diff
@@ -751,6 +747,7 @@ IrcHandler.prototype.onKick = Promise.coroutine(function*(req, server,
             shouldKick: false,
             retry: true,
             req,
+            deop: true, // deop real irc users, likereal irc.
```
turt2live:
```suggestion
            deop: true, // deop real irc users, like real irc.
```

lib/bridge/IrcHandler.js
```diff
@@ -825,11 +822,20 @@ IrcHandler.prototype.onPart = Promise.coroutine(function*(req, server, leavingUs
         req,
         kickReason: "Client PARTed from channel", // this will only be used if shouldKick is true
         retry: true, // We must retry these so that membership isn't leaked.
+        deop: !leavingUser.isVirtual, // deop real irc users, likereal irc.
```
turt2live:
```suggestion
        deop: !leavingUser.isVirtual, // deop real irc users, like real irc.
```

lib/bridge/PowerlevelSyncer.js
```diff
@@ -0,0 +1,402 @@
...
+        };
+    }
+
+    /**
+     * Called when a m.room.power_levels is forwarded to the bridge. This should
+     * happen when a Matrix user or the bridge changes the power levels for a room.
+     * @param {MatrixEvent} event The matrix event.
+     */
+    onMatrixPowerlevelEvent(event) {
+        this._powerLevelsForRoom[event.room_id] = event.content;
```
turt2live: I believe this is a js-sdk Event class, so we should be using the functions to pull the content out.
```suggestion
        this._powerLevelsForRoom[event.room_id] = event.getContent();
```

lib/bridge/PowerlevelSyncer.js
```diff
@@ -0,0 +1,402 @@
...
+        const promises = matrixRooms.map((room) => {
+            switch (mode) {
+                case "k":
+                case "i":
+                    req.log.info((enabled ? "Locking room %s" :
+                        "Reverting %s back to default join_rule"),
+                        room.getId()
+                    );
+                    if (enabled) {
+                        return this._setMatrixRoomAsInviteOnly(room, true);
```
turt2live: it's a bit unfortunate that the powerlevel syncer also affects access to the room. Can we just call this a `AccessSyncer` instead or something?

lib/bridge/PowerlevelSyncer.js
Half-Shot: It's not, and it's terribly misleading. This event comes from an appservice transaction which we've unwrapped in `IrcBridge`.

turt2live: ohgod.gif

PR #790, opened 2019-08-06
Banned users should be quit from rooms

### LINKED ISSUE #789: User that is k-lined from freenode stays on the Matrix room to harass Matrix users
We had a problem on a mainly freenode channel where an abusive user was k-lined (network banned) and was saying "creepy stuff" on Matrix that went completely undetected by IRC ops until a Matrix user with frowning smiley happened to connect to IRC directly and notice this issue and thus told what is the case.

I don't think the bridge should be moving control of the channel from IRC to Matrix or require IRC ops to use Matrix for opping the channel, especially considering how this was in a portal room.

Half-Shot:
This fixes #789 by running the quit logic for banned users, removing them from matrix rooms. This PR also kicks users who are banned on join.

### DIFF
--- a/lib/bridge/MatrixHandler.js
+++ b/lib/bridge/MatrixHandler.js
@@ -757,7 +757,7 @@ MatrixHandler.prototype._onAdminMessage = Promise.coroutine(function*(req, event
 });
 
 MatrixHandler.prototype.quitUser = Promise.coroutine(function*(req, userId, clientList,
-                                                                ircServer, reason, banned=false) {
+                                                                ircServer, reason) {
     let clients = clientList;
     if (ircServer) {
         // Filter to get the clients for the [specified] server
@@ -811,11 +811,6 @@ MatrixHandler.prototype.quitUser = Promise.coroutine(function*(req, userId, clie
                     );
                 }
             }
-        }
-
-        if (banned) {
-            // We have already been disconnected.
-            return null;
         }
 
         req.log.info(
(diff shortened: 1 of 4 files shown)


lib/bridge/MatrixHandler.js
```diff
@@ -813,6 +813,11 @@ MatrixHandler.prototype.quitUser = Promise.coroutine(function*(req, userId, clie
             }
         }
 
+        if (banned) {
+            // We have already been disconnected.
```
turt2live: can we guarantee this? Not sure how killing a dead connection would impact the state machine (ie: can we just not have this condition?)

lib/bridge/MatrixHandler.js
```diff
@@ -960,7 +965,7 @@ MatrixHandler.prototype._onJoin = Promise.coroutine(function*(req, event, user)
                 try {
                     yield kickIntent.kick(
                         event.room_id, user.getId(),
-                    `Connection limit reached for ${room.server.domain}. Please try again later.`
+                    `IRC connection failure.`
```
turt2live: Not sure if we should change this: the error at least helps identify a problem to us.

Half-Shot: Well, the error is wrong in the case of banned users. I'm wondering if there is a sane way to pass it down the exception path.

turt2live: fair, we should just set up proper monitoring on the i-line usage instead.

lib/bridge/MatrixHandler.js
Half-Shot: Yeah, I actually added this line to stop some things from exploding --- we probably don't *actually* need to avoid calling disconnect if it's been called already.

turt2live: fair, just seemed weird to have a thing that stopped termination. If it breaks though, the condition is needed :D

PR #791, opened 2019-08-06
Allow more channel prefixes, fixes #278

### LINKED ISSUE #278: Cannot join rooms that start with a '&'
I'd like to bridge an instance of Bitlbee (an IRC <-> everything other protocol bridge) using matrix-appservice-ircd, which would allow me to use Matrix to talk to people on things like Skype / Google Hangouts. (While obviously going from Matrix <-> IRC <-> Hangouts isn't ideal, it does allow me to use a single client and not have to worry about the state of various other matrix bridge projects, since this one appears to be the mature). My configuration is set up to whitelist only myself (for obvious privacy reasons), and to bridge a single matrix room to the Bitlbee control channel, which is a room that you get force-joined to upon connecting to the server.

Unfortunately, the Bitlbee control channel is called '&bitlbee', and rooms that don't start with a # will be rejected in various ways by matrix-appservice-irc:

In IRC Admin Room:
!join localhost &bitlbee
> Format: '!join irc.example.com #channel [key]'

In logs:
> 2016-11-05 01:54:39 ERROR:req [4bikfj5fft0k8] [I->M] Cannot route PM to &bitlbee (null@localhost)

In the config file, upon boot:
```
Loading config file /home/tyler/.synapse/matrix-appservice-irc/config.yaml
{"instanceContext":"#/ircService/servers/localhost/mappings","resolutionScope":"anon-schema://8b4bb394aee20d180017ba591a723b4b5be6f69a/#/properties/ircService/properties/servers/additionalProperties/properties/mappings","constraintName":"additionalProperties","testedValue":"&bitlbee","desc":"property \"&bitlbee\" not allowed by \"properties\" or by \"patternProperties\" and \"additionalProperties\" is false","kind":"ObjectValidationError"}
/home/tyler/.synapse/matrix-appservice-irc/node_modules/matrix-appservice-bridge/lib/components/config-validator.js:47
        throw e;
        ^

Error: Failed to validate file
    at ConfigValidator.validate (/home/tyler/.synapse/matrix-appservice-irc/node_modules/matrix-appservice-bridge/lib/components/config-validator.js:45:17)
    at Cli._loadConfig (/home/tyler/.synapse/matrix-appservice-irc/node_modules/matrix-appservice-bridge/lib/components/cli.js:159:22)
    at Cli._assignConfigFile (/home/tyler/.synapse/matrix-appservice-irc/node_modules/matrix-appservice-bridge/lib/components/cli.js:144:23)
    at Cli.run (/home/tyler/.synapse/matrix-appservice-irc/node_modules/matrix-appservice-bridge/lib/components/cli.js:138:10)
    at Object.<anonymous> (/home/tyler/.synapse/matrix-appservice-irc/app.js:28:4)
    at Module._compile (module.js:556:32)
    at Object.Module._extensions..js (module.js:565:10)
    at Module.load (module.js:473:32)
    at tryModuleLoad (module.js:432:12)
    at Function.Module._load (module.js:424:3)
    at Module.runMain (module.js:590:10)
    at run (bootstrap_node.js:394:7)
    at startup (bootstrap_node.js:149:9)
    at bootstrap_node.js:509:3
```

The Bitlbee FAQ specifically covers this edge case (see the last question on https://wiki.bitlbee.org/FAQ), and suggests that IRC clients that have trouble connecting to rooms that start with a '&' are client bugs and should be reported to the IRC client developers.

maakuth:
It seems that the test coverage is a bit iffy regarding this functionality, but I haven't added tests to join !channels for example... I hope I will get around to do that.

### DIFF
--- a/lib/irc/BridgedClient.js
+++ b/lib/irc/BridgedClient.js
@@ -666,8 +666,8 @@ BridgedClient.prototype._joinChannel = function(channel, key, attemptCount) {
     if (Object.keys(this.unsafeClient.chans).indexOf(channel) !== -1) {
         return Promise.resolve(new IrcRoom(this.server, channel));
     }
-    if (channel.indexOf("#") !== 0) {
-        // PM room
+    if (!/^[#!&+]/.test(channel)) {
+        // Private room instead of channel
         return Promise.resolve(new IrcRoom(this.server, channel));
     }
     if (this.server.isExcludedChannel(channel)) {


Half-Shot:
Looks sane, though I wonder if we rely on the hash prefix in other parts of the code too...?

lib/irc/BridgedClient.js
```diff
@@ -666,8 +666,8 @@ BridgedClient.prototype._joinChannel = function(channel, key, attemptCount) {
     if (Object.keys(this.unsafeClient.chans).indexOf(channel) !== -1) {
         return Promise.resolve(new IrcRoom(this.server, channel));
     }
-    if (channel.indexOf("#") !== 0) {
-        // PM room
+    if (!/[#!&+]/.test(channel)) {
```
Half-Shot: This looks insufficent compared to the last test. `/[#&+]/` is checking if the channel has these characters anywhere in it's name rather than at the start, whereas the previous line checked the first character.

image

maakuth: Good point, I've fixed that.

maakuth:
> Looks sane, though I wonder if we rely on the hash prefix in other parts of the code too...?

Yeah, I'd like to have a spec to verify joining to one of these channels works correctly. I have to learn a bit more about the internal interfaces to do that.

Half-Shot:
Hello. This PR has been left a while, but I'm still happy to merge it if you want to port it to Typescript. It shouldn't be a huge amount of work to do so in this case.

maakuth:
Hi. Unfortunately I've not had time to do that, and also similar change needs to be applied to other places as well. Let's see.

Half-Shot:
I'm going to close this one for now, and we can re-open it if you find time.

PR #797, opened 2019-08-22
Fix ids on upgrades

### LINKED ISSUE #795: Room entries for upgraded rooms retain their old room id for mapping
This makes things like unprovisioning unsuccessful.

Half-Shot:
This is the IRC bridge side of https://github.com/matrix-org/matrix-appservice-bridge/pull/129

Fixes #795 

**EDIT**: This depends on ~~https://github.com/matrix-org/matrix-appservice-bridge/pull/132~~, we are being bitten by a bug.

**EDIT 2**: This also depends on ~~#803~~


lib/bridge/IrcBridge.js
```diff
@@ -20,7 +20,8 @@ var BridgeRequest = require("../models/BridgeRequest");
 var stats = require("../config/stats");
 var DataStore = require("../DataStore");
 var log = require("../logging").get("IrcBridge");
-const { Bridge, MatrixUser, PrometheusMetrics, Logging } = require("matrix-appservice-bridge");
+const { Bridge, MatrixUser, MatrixRoom,
```
anoadragon453:
```suggestion
const {
    Bridge,
    MatrixUser,
    MatrixRoom,
```

nit

lib/bridge/IrcBridge.js
```diff
@@ -20,7 +20,8 @@ var BridgeRequest = require("../models/BridgeRequest");
...
+    PrometheusMetrics, Logging } = require("matrix-appservice-bridge");
```
anoadragon453:
```suggestion
    PrometheusMetrics,
    Logging,
} = require("matrix-appservice-bridge");
```

/me guesses if this is how it works in JS land.

lib/bridge/IrcBridge.js
```diff
@@ -1048,6 +1050,19 @@ IrcBridge.prototype._fetchJoinedRooms = Promise.coroutine(function*() {
     }
 });
 
+
+IrcBridge.prototype._roomUpgradeMigrateEntry = function(entry, newRoomId) {
```
anoadragon453: nit: Might be nice to have a comment here along the lines of https://github.com/matrix-org/matrix-appservice-bridge/pull/129 for why this function is necessary.

lib/bridge/IrcBridge.js
Half-Shot: yis

PR #799, opened 2019-08-26
Clarify storepass response

auscompgeek:
We are storing the password in the bridge database,
not saving it on the IRC server.


Half-Shot:
LGTM, can you sign off as per https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.rst#sign-off which we don't actually state anywhere on this repo..

auscompgeek:
Done. Apparently signing off is mentioned in CONTRIBUTING.rst on develop, but I didn't notice on account of it not being on master.

... say, isn't Travis CI supposed to report a status on PRs?

Half-Shot:
Probably not for non-org members, since we changed the build system. Will look into it.

PR #809, opened 2019-09-16
[Typescript] Refactor DataStore

Half-Shot:
This depends on #808 

This PR refactors the `DataStore` and `IrcRoom` classes for TypeScript which will be the foundations for the PostgreSQL support to be included in a future PR.


src/DataStore.ts
```diff
@@ -0,0 +1,630 @@
...
+
+interface ChannelMappings {
+    [roomId: string]: Array<{networkId: string, channel: string}>
+}
+
+interface UserFeatures {
+    [name: string]: boolean
+}
+
+const IrcClientConfig = require("./models/IrcClientConfig");
```
turt2live: This should be near the imports, or ideally be an import itself.

turt2live:
@Half-Shot please don't forcepush - the IrcClientConfig class wasn't reviewed :/

lgtm, but please avoid doing that.

PR #814, opened 2019-09-18
Create a generic datastore interface

Half-Shot:
This is part of the work required to support PostgreSQL. This PR changes the following:

- Create an interface which defines the public functions of the existing Datastore.
- Move NeDBs implementation to it's own datastore, which implements `Datastore`.
- Convert dependencies of Datastore (IrcServer/IrcClientConfig) to typescript
- Split out encrypted password management to a smaller class


src/datastore/NedbDataStore.ts
```diff
@@ -483,39 +466,44 @@ export class DataStore {
         log.info("storeAdminRoom (id=%s, user_id=%s)", room.getId(), userId);
         room.set("admin_id", userId);
         await this.roomStore.upsertEntry({
-            id: DataStore.createAdminId(userId),
+            id: NeDBDataStore.createAdminId(userId),
+            matrix_id: room.getId(),
```
benparsons: Having `matrix_id: room.getId()` and `matrix: room` seems fragile? Is this needed?

PR #824, opened 2019-10-01
Support room upgrades on postgresql

Half-Shot:
This uses https://github.com/matrix-org/matrix-appservice-bridge/pull/135


package.json
```diff
@@ -35,7 +35,7 @@
     "iconv": "^2.3.4",
     "irc": "matrix-org/node-irc#matrix-irc-bridge",
     "js-yaml": "^3.2.7",
-    "matrix-appservice-bridge": "^1.10.3",
+    "matrix-appservice-bridge": "matrix-org/matrix-appservice-bridge#hs/option-to-disable-stores",
```
benparsons: Is this still needed if https://github.com/matrix-org/matrix-appservice-bridge/pull/135 lands?

Half-Shot: Nope. This was just for testing, will modify before release.

PR #825, opened 2019-10-01
Heftig+hs/ident

Half-Shot:
This closes #590


src/irc/Ident.ts
```diff
@@ -146,15 +139,15 @@ class IdentSrv {
     private respond(sock: net.Socket, localPort: string, remotePort: string, username?: string) {
         let response;
         if (username) {
-            response = `${localPort},${remotePort}:USERID:UNIX:${username}\r\n`;
+            response = `${localPort}, ${remotePort}:USERID:UNIX:${username}\r\n`;
         }
         else {
-            response = `${localPort},${remotePort}:ERROR:NO-USER\r\n`;
-        }    
+            response = `${localPort}, ${remotePort}:ERROR:NO-USER\r\n`;
```
heftig: You've added a space into the response strings here. Are you sure this is right?

src/irc/Ident.ts
```diff
@@ -0,0 +1,160 @@
...
+        const username = this.portMappings[localPort];
+        if (username) {
+            log.debug("Port %s is %s", localPort, username);
+            this.respond(sock, localPort, remotePort, username);
+            return;
+        }
+        // Wait for pending connections to finish first.
+        await Promise.all([...this.pendingConnections]);
+        log.debug("No user on port %s", localPort);
+        this.respond(sock, localPort, remotePort);
```
heftig: You need to recheck the `portMappings` after waiting on the pending connections.

src/irc/Ident.ts
Half-Shot: Nope, a mistake.

src/irc/Ident.ts
```diff
@@ -130,19 +130,26 @@ class IdentSrv {
             this.respond(sock, localPort, remotePort, username);
             return;
         }
+        const mappingHash = Object.keys(this.portMappings).join("|");
         // Wait for pending connections to finish first.
         await Promise.all([...this.pendingConnections]);
+        // We don't know here if anything has actually changed, so compare hashes.
+        if (Object.keys(this.portMappings).join("|") !== mappingHash) {
+            // Hash has changed, retry.
+            await this.tryRespond(sock, localPort, remotePort);
```
heftig: I don't think recursion is the way to go here. This code can recurse forever now as long as new pending connections are made.

The old code avoided recursing forever by carrying the `currentToken` through the recursion. This way it only recurses until the pending connections *at the start of recursion* are completed.

Your `await` does all that in one step, though. So I would write it like this:

```typescript
    private async respond(sock: net.Socket, localPort: string, remotePort: string) {
        let username = this.portMappings[localPort];
        if (!username) {
            // Wait for pending connections to finish first.
            await Promise.all([...this.pendingConnections]);
            username = this.portMappings[localPort];
        }

        let response;
        if (username) {
            log.debug("Port %s is %s", localPort, username);
            response = `${localPort},${remotePort}:USERID:UNIX:${username}\r\n`;
        } else {
            log.debug("No user on port %s", localPort);
            response = `${localPort},${remotePort}:ERROR:NO-USER\r\n`;
        }
        log.debug(response);
        sock.end(response);
    }
```

This also subsumes `respond` into `tryRespond` and renames the latter to the former.

Half-Shot: Yeah, I can't fault that. Will try it.

PR #826, opened 2019-10-01
Convert ClientPool to Typescript

Half-Shot:
This converts the `ClientPool` class and it's simple dependencies to Typescript. It does not convert the more complex types as they will be handled in a separate PR.

To avoid having to use `any` in all places where we do not have types, I've created `type` aliases which will be replaced when the types are written.


src/promiseutil.ts
```diff
@@ -0,0 +1,43 @@
...
+
+import Bluebird from "bluebird";
+
+export interface Defer<T> {
+    resolve: (value?: T) => void;
+    reject: (err?: unknown) => void;
+    promise: Promise<T>;
+}
+
+export function defer<T>(): Defer<T> {
```
benparsons: this & `Queue` have been explained and excused IRL

PR #834, opened 2019-10-04
Create seperate DBs for each integration test

Half-Shot:
@babolivier says this is what the synapse folks do, so it might help with collisions.


Half-Shot:
Yay, that works

PR #836, opened 2019-10-08
Typescriptify IrcBridge

Half-Shot:


src/bridge/IrcBridge.ts
```diff
@@ -0,0 +1,1199 @@
...
+
+    public getClientPool() {
+        return this.clientPool;
+    }
+
+    public getProvisioner() {
+        return this.provisioner;
+    }
+
+    public createBridgedClient(ircClientConfig: IrcClientConfig, matrixUser: MatrixUser|null, isBot: boolean) {
```
benparsons: re `matrixUser` being nullable, does this need to be a special case, maybe with another create* method? Seems strange to sometimes pass null here considering the whole goal is to get it all typed

Half-Shot: Yeah, I'm starting to think we should have a dedicated creator for bot clients.

Half-Shot: Logged as #852 because we should merge this first.

PR #839, opened 2019-10-10
Convert irc/*Syncer classes to Typescript

Half-Shot:
**depends on #836**

This converts the syncer classes into Typescript. The logic has been left alone for the most part. I recommend reviewing line by line.


src/datastore/postgres/PgDataStore.ts
```diff
@@ -505,6 +505,10 @@ export class PgDataStore implements DataStore {
         return new MatrixUser(res.rows[0].user_id, res.rows[0].data);
     }
 
+    public async roomUpgradeOnRoomMigrated(oldRoomId: string, newRoomId: string) {
+        await this.pgPool.query("UPDATE rooms SET room_id = $1 WHERE room_id = $2", [newRoomId, oldRoomId]);
```
benparsons: love how simple this is in the new pg era

Half-Shot: :)

PR #842, opened 2019-10-13
Generic connection options for node-irc

### LINKED ISSUE #840: Option to allow CA certificates with weak signature digests
for my bridge to a network with a self-signed certificate, I had to make some low-level changes to node-irc, to make sure it didn't error out with a "CA signature digest algorithm too weak"

This involves changing`node-irc`'s connectionOpts, and appending `@SECLEVEL=0`[1] to the standard cipherlist[2] in the cert key.

I'd like to upstream that, but I'm unsure if it should be exposed on the matrix-appservice-irc side as "Object to be merged with node-irc's connectionOpts", or a specific flag just for allowing weak CA signature digests

[1] https://github.com/openssl/openssl/issues/3558#issuecomment-383154880
[2] https://nodejs.org/api/tls.html#tls_modifying_the_default_tls_cipher_suite

f0x52:
This fixes #840
It moves the `ca` option to a more generic `connectionOpts` object, which can be used to directly pass options to the `tls.connect()` in node-irc. There's also an example in `config.sample.yaml` to use this for fixing the "CA signature digest algorithm too weak" error

Will need a version bump in node-irc, merged with this PR: https://github.com/matrix-org/node-irc/pull/39

### DIFF
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -62,10 +62,19 @@ ircService:
       # Whether to use SSL or not. Default: false.
       ssl: true
       # Object to merge with node-irc's connectionOpts
-      # Can be used to override keys like 'ciphers'
+      # Can be used to override keys like ca and ciphers
       # valid options in https://nodejs.org/api/tls.html#tls_tls_connect_options_callback
       # and https://nodejs.org/api/tls.html#tls_tls_createsecurecontext_options
-      connectionOpts: {}
+      #connectionOpts: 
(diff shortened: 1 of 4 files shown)


Half-Shot:
Just for history's sake, you need to make a few changes here to make this reviewable:
- Base your changes on `develop`
- The code is mostly now in `Typescript` with a few pending PRs to go.
- You need a changelog.d entry.
- You need to sign this off as per CONTRIBUTING.md

PR #844, opened 2019-10-22
Shortcircuit keepalives

Half-Shot:
This just saves us a bit of memory and event loop time, but it's also an easy one so let's do it. We do lose the logging that tells us that users are being kept alive, but that can also be inferred from the configuration file.

### DIFF
--- a/src/irc/BridgedClient.ts
+++ b/src/irc/BridgedClient.ts
@@ -656,6 +656,7 @@ export class BridgedClient extends EventEmitter {
             }, (1000 * idleTimeout));
         }
     }
+
     private removeChannel(channel: string) {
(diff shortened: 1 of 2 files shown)


src/irc/BridgedClient.ts
```diff
@@ -628,6 +628,12 @@ export class BridgedClient extends EventEmitter {
 
     private keepAlive() {
         this.lastActionTs = Date.now();
```
anoadragon453: Might as well put this below the membership/bot check.

Half-Shot: We do still use this one for metrics c:, which is a bit misleading I know.

PR #845, opened 2019-10-22
Convert MatrixHandler to Typescript

Half-Shot:
**depends on #839**


Half-Shot:
Sorry about the force pushes. Trying to refactor it to sit nicely on top of `develop`.

src/models/BridgeRequest.ts
```diff
@@ -37,8 +38,10 @@ export class BridgeRequest {
     reject(err?: unknown) {
         this.req.reject(err);
     }
+}
 
-    public static ERR_VIRTUAL_USER = "virtual-user";
-    public static ERR_NOT_MAPPED = "virtual-user";
-    public static ERR_DROPPED = "virtual-user";
+export enum BridgeRequestErr {
```
Half-Shot: We're using Typescript `enum`s here instead of string values because they are quite handy. Similar to C# enums.

spec/util/app-service-mock.js
```diff
@@ -85,6 +85,9 @@ MockAppService.prototype.listen = function(port) {
 };
 
 MockAppService.prototype._trigger = function(eventType, content) {
+    if (content.user_id) {
```
Half-Shot: Matrix used to use `user_id` instead of sender in the good ol days. Bridge transactions currently come down in both formats from the homeserver, so while `user_id` works, we should use `sender`. For the sake of not having to change all the tests and code at the same time, we've bodged in this change to ensure both keys exist in tests.

src/bridge/IrcBridge.ts
```diff
@@ -522,16 +519,17 @@ export class IrcBridge {
         }
 
         // SUCCESS
-        this.bridge.getRequestFactory().addDefaultResolveCallback((req, res: string) => {
-            if (res === BridgeRequest.ERR_VIRTUAL_USER) {
+        this.bridge.getRequestFactory().addDefaultResolveCallback((req, _res) => {
+            const res = _res as BridgeRequestErr|null;
+            if (res === BridgeRequestErr.ERR_VIRTUAL_USER) {
```
Half-Shot: We use `BridgeRequestErr` instead of `string` now to ensure more type correctness.

src/bridge/IrcBridge.ts
```diff
@@ -888,7 +891,7 @@ export class IrcBridge {
         return new IrcUser(ircInfo.server, ircInfo.nick, true);
     }
 
-    public async trackChannel(server: IrcServer, channel: string, key: string): Promise<IrcRoom> {
+    public async trackChannel(server: IrcServer, channel: string, key?: string): Promise<IrcRoom> {
```
Half-Shot: Channels do not always need a `key` to enter them.

src/bridge/RoomAccessSyncer.ts
```diff
@@ -310,7 +310,7 @@ export class RoomAccessSyncer {
                 this.ircBridge.getStore().setModeForRoom(room.getId(), "s", enabled);
             });
             // Update the visibility for all rooms connected to this channel
-            await this.ircBridge.publicitySyncer.updateVisibilityMap(
+            this.ircBridge.publicitySyncer.updateVisibilityMap(
```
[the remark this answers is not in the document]
Half-Shot: This is not an `async` function.

PR #849, opened 2019-10-28
Remove warnings about deprecated config values

Half-Shot:
These were added a long time ago (before I took over the project) and I'm going to make the broad assumption that users are no longer using the ancient format. Regardless, we will be making a breaking release anyway so let's remove these.

### DIFF
--- a/changelog.d/849.removal
+++ b/changelog.d/849.removal
@@ -1,0 +1,1 @@
+Remove warnings/hacks around `config.appservice`. Users should have upgraded to the new format by now.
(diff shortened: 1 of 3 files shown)


Half-Shot:
Some of this has already been incorporated into #836 so the changelog will look a little strange.

Half-Shot:
In previous releases, we used the option and just warned about it. This release removes it, but since we are going to be doing a major release anyway it's probably fine.

turt2live:
as long as it's already removed, this is fine to merge. If it hasn't been removed yet, I would suggest removing it here.

PR #853, opened 2019-10-30
Update Dockerfile to support Typescript

Half-Shot:
And do a few misc cleanup things while we're at it.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,11 +15,7 @@ WORKDIR /build
 
 RUN apt-get update && apt-get install -y git python3 libicu-dev build-essential
 
-COPY ./package.json /build/package.json
-COPY ./package-lock.json /build/package-lock.json
-COPY ./src /build/src
-COPY ./tsconfig.json /build/tsconfig.json
-COPY ./types /build/types
+COPY . /build/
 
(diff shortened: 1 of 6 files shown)


Dockerfile
```diff
@@ -1,36 +1,43 @@
-# Builder
-FROM node:10-slim as builder
+# Freebind build
+FROM node:12-slim as freebind
```
turt2live: this doesn't seem to need node?

Dockerfile
```diff
@@ -1,36 +1,43 @@
...
+
+WORKDIR /build
+
+RUN apt-get update && apt-get install -y git python3 libicu-dev build-essential
+
+COPY ./package.json /build/package.json
+COPY ./package-lock.json /build/package-lock.json
+COPY ./src /build/src
+COPY ./tsconfig.json /build/tsconfig.json
+COPY ./types /build/types
```
turt2live: you can avoid all of these by using a `.dockerignore` file btw (ie: make it a single COPY operation)

Dockerfile
Half-Shot: Yes this could really be anything..

PR #855, opened 2019-10-30
Tweak counter for leaving users

Half-Shot:
**depends on #845**

This change makes it so that we do not queue a leave operation if there are no users to leave. In addition, we no longer count the number of queued promises as the number of leaving users.

### DIFF
--- a/src/bridge/MemberListSyncer.ts
+++ b/src/bridge/MemberListSyncer.ts
@@ -406,6 +406,7 @@ export class MemberListSyncer {
 
         // If a userId is in remoteJoinedUsers, but not ircUserIds, intend on leaving roomId
         const promises: Promise<unknown>[] = [];
+        let totalLeavingUsers = 0;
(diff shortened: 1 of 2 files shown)


src/bridge/MemberListSyncer.ts
```diff
@@ -421,17 +422,20 @@ export class MemberListSyncer {
                     return !ircUserIds.includes(userId);
                 }
             );
-
-            this.usersToLeave += usersToLeave.length;
+            if (usersToLeave.length < 1) {
```
turt2live: which is also this?
```suggestion
            if (usersToLeave.length === 0) {
```

Half-Shot: I like to think `< 1` is for those who are extra paranoid. But point.

turt2live: I normally do`<= 0` which people hate :p

PR #857, opened 2019-11-01
Add yaml config options to set port and hostname

### LINKED ISSUE #95: Allow binding only to localhost instead of to 0.0.0.0
Currently when running the bridge, it listens on port 9999 on all interfaces.

It would be great if the bridge could listen only on 127.0.0.1:9999 when the matrix service is also running locally, so that the port isn't opened to the world.

Half-Shot:
Fixes #95 
This PR allows admins to specify the port and hostname they want to bind the appservice to in the config file. The port takes priority over the CLI provided value. The hostname has no equivalent in the CLI, but will default to 0.0.0.0 as before if not specified. This PR also bumps us up to 1.11.1.

### DIFF
--- a/app.js
+++ b/app.js
@@ -5,11 +5,11 @@ const main = require("./lib/main");
 const path = require("path");
 
 const REG_PATH = "appservice-registration-irc.yaml";
-
 new Cli({
     registrationPath: REG_PATH,
     enableRegistration: true,
     enableLocalpart: true,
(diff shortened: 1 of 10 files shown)


turt2live:
> The port takes priority over the CLI provided value. 

This should be the other way around: CLI is more specific than the config.

Half-Shot:
Hopefully fixed.

PR #863, opened 2019-11-05
Split out bot command handling to a seperate class

Half-Shot:
One thing that came up during the grand refactor of `MatrixHandler` is the size of the `onAdminMessage` function. This PR changes it so that admin command handling is separated from the general matrix handling code.


src/bridge/AdminRoomHandler.ts
```diff
@@ -0,0 +1,603 @@
+import { BridgeRequest } from "../models/BridgeRequest";
```
turt2live: missing copyright?

PR #865, opened 2019-11-05
Replace some calls to .roomState()

Half-Shot:
There are a few places when handling invites where we call roomState, and then use it to find the state of one user. This change removes those and uses more sensible alternatives. This change is intended to improve the performance of the bridge for matrix.org, where /state calls are costly.

### DIFF
--- a/changelog.d/865.bugfix
+++ b/changelog.d/865.bugfix
@@ -1,0 +1,1 @@
+Replace calls to /state with more efficent calls, where possible.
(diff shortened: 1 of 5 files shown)


changelog.d/865.bugfix
```diff
@@ -0,0 +1 @@
+Replace calls to /state with more efficent calls, where possible.
```
turt2live: please consider backticks:
```suggestion
Replace calls to `/state` with more efficient calls, where possible.
```

Half-Shot:
Yep.

PR #866, opened 2019-11-05
Fix topics echoing after setting it on Matrix

### LINKED ISSUE #817: Changing the topic on Matrix can cause the irc bot to change the topic
This is probably because the IRC bridge sees a topic change coming from IRC, and doesn't identify the irc user as a virtual one

Half-Shot:
Fixes #817 

The bug worked as follows:
- A matrix user send a m.room.topic update.
- The IRC bridge sends a topic update to IRC.
- The IRC bridge hears a topic update from a virtual IRC user.
- The IRC bridge sets the topic on Matrix which is identical to the new topic (except perhaps if the topic gets cut off).
- Because we use the actual irc user's ghost to set topics now, the real Matrix user's virtual irc ghost will appear in the room and try to set the topic.

### DIFF
--- a/changelog.d/866.bugfix
+++ b/changelog.d/866.bugfix
@@ -1,0 +1,1 @@
+Topic changes from Matrix no longer cause a ghost user to join the room.
--- a/src/bridge/IrcHandler.ts
+++ b/src/bridge/IrcHandler.ts
@@ -504,6 +504,11 @@ export class IrcHandler {
             server.domain, fromUser, channel, JSON.stringify(action).substring(0, 80)
         );
 
+        if (fromUser.isVirtual) {
+            // Don't echo our topics back.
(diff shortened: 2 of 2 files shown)


changelog.d/866.bugfix
```diff
@@ -0,0 +1 @@
+Topic changes from Matrix no longer cause a ghost user to join the room.
```
anoadragon453: I'm failing to see how this relates to the rest of the described problem?

Half-Shot: The changelog describes the fix from a user PoV. The issue mentioned is poorly explained. I'll update the description of this PR for historical reasons.

anoadragon453: Makes sense now, thanks!

PR #867, opened 2019-11-05
Ensure bot client will reconnect

### LINKED ISSUE #862: Bridge bot doesn't reconnect after a lost connection
On the IRCnet bridge the bridge bot (for some reason) tends to get disconnected after some time. 
When that happens, the bridge bot doesn't reconnect, leaving all Bot-related functionality broken.

Bot config:

```yaml
      botConfig:
        enabled: false
        nick: SNTMatrixBot
        joinChannelsIfNoUsers: false
```

My sneaking suspicion is that this is caused by  [this check](https://github.com/matrix-org/matrix-appservice-irc/blob/3c68b6f82692ef4bdfa16b00a9de13089d8a063d/lib/irc/ClientPool.js#L402), but I cannot confirm because we have the bridge set to log level `warn`.

Half-Shot:
Fixes #862 

This ensures the bot will reconnect, even if it isn't in any channels (Freenode's bot doesn't linger in channels, for ex).

### DIFF
--- a/changelog.d/867.bugfix
+++ b/changelog.d/867.bugfix
@@ -1,0 +1,1 @@
+Ensure bot clients stay connected after being disconnected
--- a/changelog.d/867.bugfix
+++ b/changelog.d/867.bugfix
@@ -1,1 +1,1 @@
-Ensure bot clients stay connected after being disconnected
+Ensure bot clients stay connected after being disconnected.
--- a/src/irc/ClientPool.ts
+++ b/src/irc/ClientPool.ts
@@ -403,6 +403,7 @@ export class ClientPool {
         }
 
         if (bridgedClient.explicitDisconnect) {
+            log.info(`Dropping ${bridgedClient.id} (${bridgedClient.nick}) because explicitDisconnect is true`);
             // don't reconnect users which explicitly disconnected e.g. client
             // cycling, idle timeouts, leaving rooms, etc.
             return;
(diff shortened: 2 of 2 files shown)


changelog.d/867.bugfix
```diff
@@ -0,0 +1 @@
+Ensure bot clients stay connected after being disconnected
```
anoadragon453: Not sure if you care, but missing a trailing period.
```suggestion
Ensure bot clients stay connected after being disconnected.
```

src/irc/ClientPool.ts
```diff
@@ -403,6 +403,7 @@ export class ClientPool {
         }
 
         if (bridgedClient.explicitDisconnect) {
+            log.info(`Dropping ${bridgedClient.id} (${bridgedClient.nick}) because explicitDisconnect is true`);
```
anoadragon453: I know the guy in the issue said his log level was only `warn`, but this sounds like a `debug` log line more than an `info` one.

src/irc/ClientPool.ts
```diff
@@ -414,21 +415,16 @@ export class ClientPool {
...
         );
         const chanList = bridgedClient.chanList;
         // remove ref to the disconnected client so it can be GC'd. If we don't do this,
         // the timeout below holds it in a closure, preventing it from being GC'd.
         (bridgedClient as unknown) = undefined;
 
-        if (chanList.length === 0) {
-            log.info(`Dropping ${cli.id} ${cli.nick} because they are not joined to any channels`);
+        if (chanList.length === 0 && !bridgedClient.isBot) { // Never drop the bot.
+            log.info(`Dropping ${cli.id} (${cli.nick}) because they are not joined to any channels`);
```
anoadragon453: likewise

src/irc/ClientPool.ts
Half-Shot: Debatable, we wouldn't expect to see a ton of these because not many people disconnect on purpose. I would argue the information is rather vital?

src/irc/ClientPool.ts
Half-Shot: Same deal. Not many people disconnect who have also never joined a channel. You'd expect to see both these messages possibly 50 times a day at the worst case on FN, which would put them in the info category for me.

PR #868, opened 2019-11-06
Hs/newsfile docs

Half-Shot:


Half-Shot:
We did this already :man_facepalming:

PR #873, opened 2019-11-12
s/counter/count

Half-Shot:
Fixing a typo.


Half-Shot:
We've never really adopted a standard. I guess I might go through and do the dots before release.

PR #876, opened 2019-11-14
Track last seen time across restarts

Half-Shot:
This means the connection reaper stuff should actually be usable, since we can trust that after 28+ days of soaking up timings, we are actually evicting people who are idle.

### DIFF
--- a/spec/integ/connection-reaping.spec.js
+++ b/spec/integ/connection-reaping.spec.js
@@ -63,7 +63,7 @@ describe("Connection reaping", function() {
         expect(res.inactiveMs).toBeLessThanOrEqual(50);
     });
 
-    it("users should appear online if they have sent a message", async () => {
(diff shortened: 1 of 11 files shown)


spec/integ/connection-reaping.spec.js
```diff
@@ -0,0 +1,81 @@
...
+            sender: testUser.id,
+            room_id: roomMapping.roomId,
+            type: "foo.notamessage",
+        });
+        const res = await env.ircBridge.activityTracker.isUserOnline(testUser.id, 1000);
+        expect(res.online).toBeTruthy();
+        expect(res.inactiveMs).toBeLessThanOrEqual(50);
+    });
+
+    it("users should appear online if they have sent a message", async () => {
```
benparsons: need to differentiate these test names?

PR #882, opened 2019-11-15
Store room visibility cache in the DB

### LINKED ISSUE #880: We should cache directory PUTs
Rather than PUTing all the rooms on startup each time, the database should keep track of it's state locally and only make the call if the room's status has changed. Currently, we absolutely hammer the homeserver on startup and most of the requests fail anyway.

Half-Shot:
Soft dependency on #881 

Fixes #880 

This is a simple change which caches the room directory visibility in the database rather than just in memory, so that restarts do not spam the homeserver with requests.

### DIFF
--- a/src/bridge/PublicitySyncer.ts
+++ b/src/bridge/PublicitySyncer.ts
@@ -202,8 +202,8 @@ export class PublicitySyncer {
         const cli = this.ircBridge.getAppServiceBridge().getBot().getClient();
 
         // Update rooms to correct visibilities
-        const promises = roomIds.map((roomId) => {
-            const currentState = this.visibilityMap.roomVisibilities[roomId];
+        const promises = roomIds.map(async (roomId) => {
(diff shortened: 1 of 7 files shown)


src/bridge/PublicitySyncer.ts
```diff
@@ -202,25 +202,25 @@ export class PublicitySyncer {
         const cli = this.ircBridge.getAppServiceBridge().getBot().getClient();
 
         // Update rooms to correct visibilities
-        const promises = roomIds.map((roomId) => {
-            const currentState = this.visibilityMap.roomVisibilities[roomId];
+        const promises = roomIds.map(async (roomId) => {
+            const currentState = await this.ircBridge.getStore().getRoomVisibility(roomId);
```
benparsons: Possible perf issue - what is the expected size of `roomIds` here? Could end up making loads of tiny `select`s when you could group them together then iterate.

Issue #887, opened 2019-11-23
Autopastebinned messages are missing URLs on IRC


Half-Shot:
This is broken on `develop` (so what we have currently deployed to Freenode). We need some tests on this.

PR #889, opened 2019-11-25
Fix uploaded long message URL's not sent to IRC side

jaywink:
The call was changed in #845 to instead of getting the raw response back, to get just the content
uri back. Due to wrong type definition of the uploadContent
method, the change actually made the raw response still get back
due to legacy defaults in uploadContent.

Refactor the call to pass the buffer as first argument and opts
as second. This has been preferred in matrix-js-sdk since 2016.

Refs: #887
Signed-off-by: Jason Robinson <jasonr@matrix.org>

### DIFF
--- a/changelog.d/889.bugfix
+++ b/changelog.d/889.bugfix
@@ -1,0 +1,1 @@
+Fix uploaded long message URL's not sent to IRC side.
--- a/src/bridge/IrcBridge.ts
+++ b/src/bridge/IrcBridge.ts
@@ -602,13 +602,15 @@ export class IrcBridge {
     }
 
     public uploadTextFile(fileName: string, plaintext: string) {
-        return this.bridge.getIntent().getClient().uploadContent({
-            stream: new Buffer(plaintext),
-            name: fileName,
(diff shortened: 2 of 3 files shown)


jaywink:
> LGTM but ideally would be paired with a regression test

I ... tried, for a bit. Unfortunately since `IrcBridge` has no existing tests (afaict), I'm having a hard time getting it mocked up so that I could write the test. Since this is a regression affecting users, I'm thinking we should roll the fix out.

I'm a bit worried about the buildkite tests though, any idea whether they're expected to fail or not? Locally the `npm run check` does run through ok, except for random failure as per #890

t3chguy:
> I'm thinking we should roll the fix out.

I agree, hence the Approve/LGTM

That test failure looks completely unrelated, the change is just how you interact with js-sdk so I second ignoring that test failure for this instance. Merge and deploy away.

jaywink:
> That test failure looks completely unrelated, the change is just how you interact with js-sdk so I second ignoring that test failure for this instance. Merge and deploy away.

Yep, also confirmed now with a rebuild. Merging :)

PR #897, opened 2019-11-28
Add support for sentry.io tracing

Half-Shot:
This is part of a plan to start routing errors to Sentry for review rather than trying to grep through the logs after an error has been reported to us.


src/bridge/IrcBridge.ts
```diff
@@ -523,6 +523,7 @@ export class IrcBridge {
             logMessage(req, "FAILED");
             stats.request(isFromIrc, "fail", req.getDuration());
             this.logMetric(req, "fail");
+            BridgeRequest.HandleExceptionForSentry(req, "failed");
```
benparsons: this should be `fail`, to match the others? FAILED, fail, fail, failed

src/bridge/IrcBridge.ts
```diff
@@ -536,6 +537,7 @@ export class IrcBridge {
             const isFromIrc = Boolean((req.getData() || {}).isFromIrc);
             stats.request(isFromIrc, "fail", req.getDuration());
             this.logMetric(req, "fail");
+            BridgeRequest.HandleExceptionForSentry(req, "dead");
```
benparsons: fail, fail, dead?

src/bridge/IrcBridge.ts
Half-Shot: Hmm, tricky. From an english standpoint, I liked state="failed" more than state="fail". However, it does introduce some inconsistency.

benparsons: I don't mind which, just think they should be consistent

PR #907, opened 2019-12-06
Increase default txn payload size limit and allow it to be configurable

Half-Shot:
As an alternative solution to https://github.com/matrix-org/synapse/issues/3536 for now.


src/bridge/IrcBridge.ts
```diff
@@ -165,6 +167,17 @@ export class IrcBridge {
...
+
+
+        const homeserverToken = this.registration.getHomeserverToken();
+        if (!homeserverToken) {
+            throw Error("No HS token defined");
+        }
+
+        this.appservice = new AppService({
+            homeserverToken,
+            httpMaxSizeBytes: (this.config.advanced || { }).maxTxnSize || TXN_SIZE_DEFAULT,
```
benparsons: I would rather ensure that `this.config.advanced` is always at least `{}` rather than this

Half-Shot: Yep, we generally need a default config to rely on. This will do for now.

PR #918, opened 2019-12-13
Handle replies that contain displaynames

### LINKED ISSUE #915: Messages with `m.in_reply_to` are not propagating to IRC
Stumbled upon this bug while investigating why Telegram-to-Telegram replies (generated by mautrix-telegram appservice) are not propagated to IRC ([see bug report](https://github.com/tulir/mautrix-telegram/issues/400)).

To reproduce, send the following minimal message from Matrix:

```
{
    "body": "testing",
    "msgtype": "m.text",
    "m.relates_to": {
	"event_id": "$15760700702709kXicd:hacklab.fi",
	"rel_type": "m.reference",
	"m.in_reply_to": {
            "event_id": "$15760700702709kXicd:hacklab.fi"
	}
    }
}
```

And then remove `m.in_reply_to` object and resend the message. The message is then propagating to IRC.

Half-Shot:
Fixes #915

### DIFF
--- a/src/bridge/MatrixHandler.ts
+++ b/src/bridge/MatrixHandler.ts
@@ -1097,7 +1097,7 @@ export class MatrixHandler {
 
     private async textForReplyEvent(event: MatrixMessageEvent, replyEventId: string, ircRoom: IrcRoom):
     Promise<{formatted: string; reply: string}|null> {
-        const REPLY_REGEX = /> <(@.*:.*)>(.*)\n\n(.*)/;
+        const REPLY_REGEX = /> <(.*)>(.*)\n\n(.*)/;
         const REPLY_NAME_MAX_LENGTH = 12;
         const eventId = replyEventId;
         if (!event.content.body) {
--- a/changelog.d/918.bugfix
+++ b/changelog.d/918.bugfix
@@ -1,0 +1,1 @@
+Handle replies that contain a displayname rather than a userid.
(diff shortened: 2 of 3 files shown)


src/bridge/MatrixHandler.ts
```diff
@@ -1097,7 +1097,7 @@ export class MatrixHandler {
 
     private async textForReplyEvent(event: MatrixMessageEvent, replyEventId: string, ircRoom: IrcRoom):
     Promise<{formatted: string; reply: string}|null> {
-        const REPLY_REGEX = /> <(@.*:.*)>(.*)\n\n(.*)/;
+        const REPLY_REGEX = /> <(.*)>(.*)\n\n(.*)/;
```
14mRh4X0r: Might want to change this to:

```js
const REPLY_REGEX = /> <(.*?)>(.*?)\n\n(.*)/;
```

These [non-greedy matches](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_Expressions/Quantifiers) make sure that in e.g.

```
<SomeName> if x > 2

Some reply that could, technically

Contain >2 newlines

Or even more
```

the first group will match `SomeName` instead of `SomeName> if x`, or, in this case,

```
SomeName> if x > 2

Some reply that could, technically,

Contain 
```

And the second group `if x > 2` instead of `  2` or, in this case, `2 newlines`,
and the third group 

```
Some reply that could, technically

Contain >2 newlines

Or even more
```

instead of 

```
Contain >2 newlines

Or even more
```

or, in this case,

```
Or even more
```

14mRh4X0r: I just realized that this *would* break on display names with `>` in them. To rule this out, I guess you'd need to parse the HTML.

PR #919, opened 2019-12-13
Discard old PM rooms

### LINKED ISSUE #911: duplicate key value violates unique constraint "cons_pm_rooms_matrix_irc_unique"
Seeing a fair few of these in Sentry, should investigate.

Half-Shot:
Fixes #911 

When a new PM room gets created/assigned, we want to drop the old one.

### DIFF
--- a/changelog.d/919.bugfix
+++ b/changelog.d/919.bugfix
@@ -1,1 +1,1 @@
-If a new PM room is created for a IRC user, discard the old room.
+If a new DM room is created for a IRC user, discard the old room.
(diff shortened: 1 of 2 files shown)


changelog.d/919.bugfix
```diff
@@ -0,0 +1 @@
+If a new PM room is created for a IRC user, discard the old room.
```
turt2live: Consistency with Matrix:
```suggestion
If a new DM room is created for a IRC user, discard the old room.
```

PR #920, opened 2019-12-14
fix Dockerfile

lub:
Before config.schema.yml was missing in the image, because conf.* was
listed in the .dockerignore
This commit uses explicit names instead of `COPY .` and thus makes it
more readable (because everything is contained in the Dockerfile),
better cacheable and fixes the missing config.schema.yml

Signed-off-by: Simon Koerner <git@lubiland.de>


Half-Shot:
Thanks, could you also add a changelog file as per https://github.com/matrix-org/matrix-appservice-irc/blob/develop/CONTRIBUTING.md#making-pull-requests. Apparently our CI is now failing to check for it.

lub:
I've added a changelog file

PR #921, opened 2019-12-14
fix startup without sentry config

lub:
This actually respects config.sentry.enabled and thus makes the
bridge usable without configuring sentry.

https://matrix.to/#/!BAXLHOFjvDKUeLafmO:matrix.org/$afPi3kvwcVUp6zc3Th5fEVans8SqOFx+xPo9HQUbk/Q?via=matrix.org&via=chat.weho.st&via=privacytools.io

Signed-off-by: Simon Koerner <git@lubiland.de>

### DIFF
--- a/changelog.d/921.bugfix
+++ b/changelog.d/921.bugfix
@@ -1,0 +1,1 @@
+Stop trying to use sentry when config.sentry.enabled is false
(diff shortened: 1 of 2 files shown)


Half-Shot:
Oh fun. Thanks.

Half-Shot:
Could you also add a changelog file as per https://github.com/matrix-org/matrix-appservice-irc/blob/develop/CONTRIBUTING.md#making-pull-requests. Apparently our CI is now failing to check for it.

lub:
I've added a changelog file

PR #925, opened 2019-12-18
Fix reply regex again

Half-Shot:
See https://github.com/matrix-org/matrix-appservice-irc/pull/918#discussion_r359110255

### DIFF
--- a/src/bridge/MatrixHandler.ts
+++ b/src/bridge/MatrixHandler.ts
@@ -1097,7 +1097,7 @@ export class MatrixHandler {
 
     private async textForReplyEvent(event: MatrixMessageEvent, replyEventId: string, ircRoom: IrcRoom):
     Promise<{formatted: string; reply: string}|null> {
-        const REPLY_REGEX = /> <(.*)>(.*)\n\n(.*)/;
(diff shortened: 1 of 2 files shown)


Half-Shot:
Tests pass locally. Merging.

src/bridge/MatrixHandler.ts
```diff
@@ -1097,7 +1097,7 @@ export class MatrixHandler {
 
     private async textForReplyEvent(event: MatrixMessageEvent, replyEventId: string, ircRoom: IrcRoom):
     Promise<{formatted: string; reply: string}|null> {
-        const REPLY_REGEX = /> <(.*)>(.*)\n\n(.*)/;
+        const REPLY_REGEX = /> <(.*?)>(.*?)\n\n(.*)/;
```
tulir: Although this no longer breaks with the telegram bridge, it still doesn't do what the spec says you should do:

> To strip the fallback on the `body`, the client should iterate over each line of the string, removing any lines that start with the fallback prefix ("> ", including the space, without quotes) and stopping when a line is encountered without the prefix. This prefix is known as the "fallback prefix sequence".

PR #929, opened 2019-12-20
Fix a typo when moving to typescript that broke nedb support

kaiyou:
It seems that when migrating to nedb, the comparison made when checking the database type was changed from equal to different, which effectively breaks nedb support.


kaiyou:
Sorry, will fix the branching for this one.

Half-Shot:
We subsequently fixed this

PR #930, opened 2019-12-20
Add 'defaultOnline' and 'excludeRegex' parameters to /reapUsers

Half-Shot:
This is so we can exclude certain VIPs from being culled (e.g. other bridges)


src/DebugApi.ts
```diff
@@ -201,8 +201,10 @@ export class DebugApi {
         const since = parseInt(query["since"] as string);
         const reason = query["reason"] as string;
         const dry = query["dryrun"] !== undefined && query["dryrun"] !== "false";
+        const defaultOnline = query["defaultOnline"] === "false"; // Default to true.
```
benparsons: useful comment but is there a better way of setting a default here? It took me an extra second to parse.

wouldn't `const defaultOnline = query["defaultOnline"] ?? "true";` be more readable?

Half-Shot: This must be a boolean, so how about:
```
const defaultOnline = (query["defaultOnline"] ?? "true") === "true";
```

benparsons: ok that works. looks clunkier but it's the "correct" way around

PR #932, opened 2019-12-20
Share membership queue between handlers

Half-Shot:
In the hopes that this might allow us to reuse the retry logic on joins and parts


src/util/MembershipQueue.ts
```diff
@@ -15,7 +15,7 @@ interface QueueUserItem {
     reason?: string;
     attempts: number;
     roomId: string;
-    userId: string;
+    userId?: string;
```
benparsons: Marking LGTM, but can we talk about why userId is now optional throughout the Queue?

PR #947, opened 2020-01-20
Remove old DB scripts

Half-Shot:
They are likely to confuse users, and are at this point several years old. Let's point users to use the old releases to update their ancient DBs.


changelog.d/947.removal
```diff
@@ -0,0 +1 @@
+Removed `upgrade-db-0.*.js` scripts. Use an older release to upgrade NeDB.
```
turt2live: This is not very helpful - can put a link to where people can learn about the upgrade process please?

PR #949, opened 2020-01-20
Remove statsd

### LINKED ISSUE #818: Drop support for statsd
I'm looking through the code and seeing what we can prune from the bridge that is no longer in use. I'm interested to know if anyone is still using the statsd support for metrics in the bridge. Note that this is different to Prometheus metrics, which is more featureful at this point.

Half-Shot:
Fixes #818 

We don't use this in production, and nobody else really seemed that keen either. Removing it saves us a bit of maintenance work.


changelog.d/949.removal
```diff
@@ -0,0 +1 @@
+Remove `statsd` support as per #818
```
turt2live: this will probably need a real link to work

Half-Shot:
>If you only need review from one person it's generally best to not re-request the whole team.

The person in question isn't around for a while, and the changes were sufficient that it needed a re-review.

PR #956, opened 2020-01-23
Restore use of Ident.clientBegin()

heftig:
This was lost when `BridgedClient` was ported to TypeScript.


Half-Shot:
Seems legit. Thanks!

PR #957, opened 2020-01-23
Disable !join when dynamic channels are off

heftig:
AFAICT this may lead to private, statically configured channels getting exposed to the public, as `hasInviteRooms` is `false` when `dynamicChannels.enabled` is off, thus the user is considered always "authorised to join channels".

I haven't tested if this is actually the case; I just sleep better when the `!join` command is immediately denied when `dynamicChannels.enabled` is off.

### DIFF
--- a/src/bridge/AdminRoomHandler.ts
+++ b/src/bridge/AdminRoomHandler.ts
@@ -169,7 +169,7 @@ export class AdminRoomHandler {
         if (!ircChannel || ircChannel.indexOf("#") !== 0) {
             errText = "Format: '!join irc.example.com #channel [key]'";
         }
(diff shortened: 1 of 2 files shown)


Half-Shot:
Looks sound to me. Can you sign off as per https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#sign-off. I do wonder if we should have some tests to ensure this never breaks...

heftig:
> Looks sound to me. Can you sign off as per https://github.com/matrix-org/synapse/blob/master/CONTRIBUTING.md#sign-off.

Done.

> I do wonder if we should have some tests to ensure this never breaks...

I'm not familiar with the testsuite, sorry.

Half-Shot:
Yeah, I'll do it separate from this PR.

PR #969, opened 2020-02-05
Allow configuring channel keys for static mappings (with some changes to allow the existing format)

Half-Shot:
This adds some magic to use the existing format and the new format.

(Sorry @heftig, I can't append commits onto your PR)

### DIFF
--- a/HOWTO.md
+++ b/HOWTO.md
@@ -60,7 +60,8 @@ ircService:
     # This is the IRC server url to connect to.
     irc.example.com:
       mappings:
-        "#some-channel": ["!someroomid:here"]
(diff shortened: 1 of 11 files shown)


benparsons:
only thought: can we link to an anchor/title in the changelog

heftig:
> I can't append commits onto your PR

Strange, because I do have "Allow edits from maintainers" enabled.

Half-Shot:
> only thought: can we link to an anchor/title in the changelog

Not really, because the anchor have the dates of the releases in :(

Half-Shot:
I might fix it before we release though.

PR #973, opened 2020-02-09
Recommend using net.ipv6.ip_nonlocal_bind on kernels 4.3+

f0x52:
From Linux kernel 4.3+ there's now the option to set the sysctl `net.ipv6.ip_nonlocal_bind=1`, which makes preloading the freebindfree so obsolete

### DIFF
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -319,7 +319,8 @@ ircService:
           # connected user. If not specified, all users will connect from the same
           # (default) address. This may require additional OS-specific work to allow
           # for the node process to bind to multiple different source addresses
-          # e.g IP_FREEBIND on Linux, which requires an LD_PRELOAD with the library
+          # Linux kernels 4.3+ support sysctl net.ipv6.ip_nonlocal_bind=1
+          # Older kernels will need IP_FREEBIND, which requires an LD_PRELOAD with the library
           # https://github.com/matrix-org/freebindfree as Node does not expose setsockopt.
           # prefix: "2001:0db8:85a3::"  # modify appropriately
         #


Half-Shot:
If it works, this is fantastic :)

f0x52:
I have it working on production; to both Freenode and OFTC :D

Half-Shot:
We looked into this for matrix.org, and decided not to use it because:

>One thing about this is that it allows any service on the machine to listen on any IP - so it's able to incorrectly listen on an IP that won't be routable (as opposed to the checks that normally happen when you try to bind to an IP that isn't on an interface on your machine). I don't know if we should do this as it is a level of FRV protection that services will just fail quickly rather than just not having traffic routed to them.

> Ideally all that would happen is node would set this socket option itself, as all the magic freebindfree is doing is setting it for node (because node is crippled like that).

Personally, I'm still happy to recommend this but matrix.org will not use it for the time being.

sijanec:
Hello! How do I set this up on Linux?

I enabled the sysctl and routed packets for this network to the server (I can see syn-acks arriving), but connection still times out. Should I set up this IP range with `ip(8)` as well?

I do not really know how nonlocal binding works, as this is my first time interacting with it.

tcpdump output:
```
21:09:07.718819 eno1  Out ifindex 2 fc:aa:14:2a:6f:9b ethertype IPv6 (0x86dd), length 100: (flowlabel 0x2c97d, hlim 64, next-header TCP (6) payload length: 40) 2a01:261:e77:5502:eeee:faca:0:4.34155 > 2001:6b0:e:2a18::120.6697: Flags [S], cksum 0xcbbe (incorrect -> 0xe62a), seq 2836362907, win 64800, options [mss 1440,sackOK,TS val 4050643080 ecr 0,nop,wscale 7], length 0
21:09:07.788310 eno1  In  ifindex 2 00:1a:92:4b:0e:bc ethertype IPv6 (0x86dd), length 100: (hlim 45, next-header TCP (6) payload length: 40) 2001:6b0:e:2a18::120.6697 > 2a01:261:e77:5502:eeee:faca:0:4.34155: Flags [S.], cksum 0x3c94 (correct), seq 1418806385, ack 2836362908, win 0, options [mss 1460,sackOK,TS val 498659287 ecr 4050643080,nop,wscale 7], length 0
```

node output:
```
023-08-07 21:08:33 INFO:client-connection Retrying connection for MatrixBotSijanecEu on irc.libera.chat in 1671.0623791152436ms (attempts 1)
2023-08-07 21:09:04 ERROR:client-connection MatrixBotSijanecEu@irc.libera.chat still not connected after 30000ms. Killing connection.
2023-08-07 21:09:04 INFO:client-connection disconnect()ing MatrixBotSijanecEu@irc.libera.chat - timeout
2023-08-07 21:09:34 WARN:client-connection Waited for 'end' that never came
2023-08-07 21:09:34 ERROR:client-connection ConnectionInstance.connect failed after 2 attempts (timeout)
2023-08-07 21:09:34 INFO:client-connection Retrying connection for MatrixBotSijanecEu on irc.libera.chat in 659.6168642355822ms (attempts 2)
2023-08-07 21:09:43 ERROR:client-connection Server: irc.libera.chat (MatrixBotSijanecEu) Network Error: connect ETIMEDOUT 2607:5300:205:300::ae0:6697
```

EDIT/UPDATE: Sorry for bothering you. Problem solved.
It did not work just by adding `ip addr add dev eno1 2a01:261:e77:5502::1/64`.
I also had to do `2a01:261:e77:5502:eeee::/80 dev lo`, but I still do not understand why.

PR #976, opened 2020-02-12
Make sure a killed BridgedClient is dead, even if connect was never called

### LINKED ISSUE #944: unsafeClient not ready yet error
at LCA2020 we have a few people seeing this message when they try and connect to the bridge to freenode.

The bridge reports unsafeClient not ready yet 

this issue results in riot being able to join the irc channel and see what is happening on the channel but unable to send to the channel. when you send a message it appears to send correctly, but it never arrives in the irc channel, with no visible error message.

see attached screen shot

tilosp:
If for some reason `connect()` gets never called on a `BridgedClient` you will end up in a situation described in #944. The problem is that there is no way to solve this currently.
You can use the `!quit` command which calls `kill()` on `BridgedClient`, but that does not change the state of `isDead()`. This undying `BridgedClient` will remain around until the bridge gets restarted.

Fix #944

### DIFF
--- a/changelog.d/976.bugfix
+++ b/changelog.d/976.bugfix
@@ -1,0 +1,1 @@
+Make sure a killed BridgedClient is dead, even if connect was never called
--- a/src/DebugApi.ts
+++ b/src/DebugApi.ts
@@ -24,7 +24,7 @@ import { inspect } from "util";
 import { DataStore } from "./datastore/DataStore";
 import { ClientPool } from "./irc/ClientPool";
 import { getLogger } from "./logging";
-import { BridgedClient } from "./irc/BridgedClient";
+import { BridgedClient, BridgedClientStatus } from "./irc/BridgedClient";
 import { IrcBridge } from "./bridge/IrcBridge";
 import { ProvisionRequest } from "./provisioning/ProvisionRequest";
 import { getBridgeVersion } from "./util/PackageInfo";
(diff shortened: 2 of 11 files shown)


Half-Shot:
This feels like the wrong solution to the problem :(. I feel like if we add yet-another-flag then it's going to get confusing between killed and dead. Perhaps a better solution would be an enum that tracks the state of the connection. E.g. "created", "connecting", "connected", "killed", "dead".

tilosp:
@Half-Shot sure, i have a few questions regarding this enum:
- Should the connected checks, that are currently mostly based on null checking unsafeClient, be replaced with a check that state is `connected`?
- And is there any reason to have `killed` and `dead` as two different states?

Half-Shot:
>Should the connected checks, that are currently mostly based on null checking unsafeClient, be replaced with a check that state is connected?

The null checking is mostly because it makes typescript happy, and allows us to catch the edge cases. I guess we should attempt to implement the enum as well as the type checking, if poss.

>And is there any reason to have killed and dead as two different states?

I would. `dead` can mean that the connection just died, `killed` is a explicit action from the user.

tilosp:
@Half-Shot I have added the enum and replaced unsafeClient.

The ci doesn't seem to work on fork based PRs, but the tests seem to pass locally (at least the 198 that i was able to run).

Half-Shot:
@tilosp I asked and got the CI working?

tilosp:
> @tilosp I asked and got the CI working :)

thanks, and it's all green now :)

djmitche:
This matches the circumstances I'm experiencing, which are preventing me from using the bridge.  I'd love to see that fixed!

tilosp:
@Half-Shot could you take a another look at this?

Half-Shot:
Will do, it's a little invasive so it may have to happen after this release now.

tilosp:
Rebased onto master

tilosp:
@Half-Shot is there any chance that this will ever get reviewed?
If not please tell me, so I don't waste any more time on solving merge conflicts.

jaller94:
I cannot make any promises about the time this will take me to review but it's on the top of my list for community contributions.
ETA: Next week

The code looks good and we're in favour of the change, but I need to conduct some manual tests before it gets merged. :)

Nuc1eoN:
Any news?

I would like to point out that without this PR, the IRC functionality is unfixably broken for many.
I don't understand how it does not have higher priority.

tilosp:
Superseded by #1098

Issue #692, opened 2018-09-21
Notify sender when the recipient in a PM is offline


z3ntu:
Could we at least have some indication that the message wasn't delivered? Currently I mostly connect to a channel via an IRC client to check if a person is online and only then send a message in the PM which is quite annoying.

And @Mikaela wrote at https://github.com/matrix-org/matrix-appservice-irc/issues/348#issuecomment-423552863
> @z3ntu I would recommed you to open a separate issue. I imagine that the Matrix ghosts could at least messagn m.notice nick offline as they do when they are invited to a not-bridged channel.
> 
> Of course it would also be possible for IRC users to /notice matrixuser user offline and other "fun", but that doesn't seem very risky/bad for me.

Angedestenebres:
We got the same report with Channel operators that receive PM and reply without seeing that the user left IRC.

Half-Shot:
Fixed in #978

PR #978, opened 2020-02-13
Report DM failures to the DM room

Half-Shot:
Fixes https://github.com/matrix-org/matrix-appservice-irc/issues/692


src/bridge/IrcHandler.ts
```diff
@@ -875,20 +876,42 @@ export class IrcHandler {
...
+            const room = await this.ircBridge.getStore().getMatrixPmRoom(client.userId, otherUser.userId);
+            if (room) {
+                return this.ircBridge.sendMatrixAction(
+                    room, otherUser, new MatrixAction(
+                        "notice", `User is not online or does not exist. Message not sent.`
+                    ),
+                );
+            }
+            req.log.warn(`No existing PM found for ${client.userId} <--> ${otherUser.userId}`);
+            // No room associated, fall through
```
benparsons: why don't we return here? nothing left after `err_nosuchnick` is there?

Half-Shot: We still want to report the error, but if we don't have a PM room, we report it in the admin room.

PR #979, opened 2020-02-13
Fix reconnections

Half-Shot:
A bug was introduced during the typescript port where we stopped trying to rejoin channels on reconnect. Unsure if this fixes other issues yet.


src/irc/BridgedClient.ts
```diff
@@ -143,7 +143,7 @@ export class BridgedClient extends EventEmitter {
     }
 
     public get chanList() {
-        return this._chanList;
+        return Array.from(this._chanList);
```
benparsons: better to convert all calls to this to use `Set<string>`, but fine for now

Half-Shot: Yeah, maybe

PR #993, opened 2020-02-17
Ensure irc users join new room on upgrade

Half-Shot:


spec/integ/room-upgrade.spec.js
```diff
@@ -3,10 +3,21 @@ const BridgeEventType = require("../../lib/bridge/BridgeStateSyncer").BridgeStat
 
 describe("Room upgrades", function() {
     const {env, roomMapping, botUserId, test} = envBundle();
+    const testUser = {
+        id: "@flibble:wibble",
```
benparsons: this is a valid mxid I guess? If we assume `wibble` is provided by something other than DNS

Half-Shot: It's good enough for string comparison, it shouldn't ever be resolved :)

PR #1000, opened 2020-02-25
Reconnect to the correct domain on passsword changes

palmer-dabbelt:
The appservice triggers a reconnect when handling "!storepass" commands, but
rather than reconnecting to the domain for which the password was changed the
appservice just always reconnects to the first domain.  This causes an issue
for my setup, which has multiple domains in a singe appservice.

I've tested this by running it on my instance, which is connected to two
networks, and changing my password.

Signed-off-by: Palmer Dabbelt <palmerdabbelt@google.com>

### DIFF
--- a/src/bridge/AdminRoomHandler.ts
+++ b/src/bridge/AdminRoomHandler.ts
@@ -134,7 +134,7 @@ export class AdminRoomHandler {
                 await this.handleWhois(req, args, ircServer, adminRoom, event.sender);
                 break;
             case "!storepass":
-                await this.handleStorePass(req, args, ircServer, adminRoom, event.sender, clientList[0]);
+                await this.handleStorePass(req, args, ircServer, adminRoom, event.sender, clientList);
                 break;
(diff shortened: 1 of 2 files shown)


src/bridge/AdminRoomHandler.ts
```diff
@@ -419,8 +419,10 @@ export class AdminRoomHandler {
                 notice = new MatrixAction(
                     "notice", `Successfully stored password for ${domain}. You will now be reconnected to IRC.`
                 );
-                if (client) {
-                    await client.disconnect("iwantoreconnect", "authenticating", false);
+                for (let i = 0; i < clientList.length; i++) {
```
Half-Shot: Please use `for (const client of clientList)`. You can then use `client.disconnect(` below.

Half-Shot:
@palmer-dabbelt what's the status on this one?

twouters:
@Half-Shot can this be merged or is something amiss about this PR? :)

PR #1010, opened 2020-03-18
Disallow invites to PM rooms

Half-Shot:
Sometimes users will create a PM room with a irc user and then try to invite their friends to the room, which breaks horribly on IRC and causes the irc users to get a new PM channel per matrix user, which looks like a clusterfuck.

This PR forces a IRC user to leave a PM room if a user gets invited to it, and also removes the PM room from the datastore. If a new PM room is created from the IRC side, the matrix user will now lack the permissions to invite users.

### DIFF
--- a/src/bridge/IrcHandler.ts
+++ b/src/bridge/IrcHandler.ts
@@ -148,6 +148,9 @@ export class IrcHandler {
      * @return {Promise} which is resolved when the PM room has been created.
      */
     private async createPmRoom (toUserId: string, fromUserId: string, fromUserNick: string, server: IrcServer) {
+        const users: {[userId: string]: number} = { };
+        users[toUserId] = 10;
+        users[fromUserId] = 100;
         const response = await this.ircBridge.getAppServiceBridge().getIntent(
             fromUserId
         ).createRoom({
@@ -155,12 +158,29 @@ export class IrcHandler {
             options: {
                 name: (fromUserNick + " (PM on " + server.domain + ")"),
                 visibility: "private",
-                preset: "trusted_private_chat",
+                // We deliberately set our own power levels below.
+                // preset: "trusted_private_chat",
                 invite: [toUserId],
                 creation_content: {
                     "m.federate": server.shouldFederatePMs()
                 },
                 is_direct: true,
+                initial_state: [{
+                    content: {
+                        users,
+                        events: {
+                            "m.room.avatar": 10,
+                            "m.room.name": 10,
+                            "m.room.canonical_alias": 100,
+                            "m.room.history_visibility": 100,
+                            "m.room.power_levels": 100,
+                            "m.room.encryption": 100
+                        },
+                        invite: 100,
+                    },
+                    type: "m.room.power_levels",
+                    state_key: "",
+                }],
             }
         });
         const pmRoom = new MatrixRoom(response.room_id);
--- a/src/bridge/IrcHandler.ts
+++ b/src/bridge/IrcHandler.ts
@@ -17,6 +17,8 @@ import { MembershipQueue } from "../util/MembershipQueue";
 import { IrcMessage } from "../irc/ConnectionInstance";
 
 const NICK_USERID_CACHE_MAX = 512;
+const PM_POWERLEVEL_MATRIXUSER = 10;
+const PM_POWERLEVEL_IRCUSER = 100;
 
 type MatrixMembership = "join"|"invite"|"leave"|"ban";
 
@@ -149,8 +151,8 @@ export class IrcHandler {
      */
     private async createPmRoom (toUserId: string, fromUserId: string, fromUserNick: string, server: IrcServer) {
         const users: {[userId: string]: number} = { };
-        users[toUserId] = 10;
-        users[fromUserId] = 100;
+        users[toUserId] = PM_POWERLEVEL_MATRIXUSER;
+        users[fromUserId] = PM_POWERLEVEL_IRCUSER;
         const response = await this.ircBridge.getAppServiceBridge().getIntent(
             fromUserId
         ).createRoom({
--- a/src/bridge/MatrixHandler.ts
+++ b/src/bridge/MatrixHandler.ts
@@ -401,8 +401,6 @@ export class MatrixHandler {
         const inviteeIsVirtual = !!this.ircBridge.getServerForUserId(event.state_key);
         const inviterIsVirtual = !!this.ircBridge.getServerForUserId(event.sender);
 
-        console.log(rooms, event.sender)
-
         // work out which flow we're dealing with and fork off asap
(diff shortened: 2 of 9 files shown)


src/bridge/MatrixHandler.ts
```diff
@@ -336,6 +336,30 @@ export class MatrixHandler {
...
+        // We don't support this
+        req.log.warn(`User ${inviter.getId()} tried to invite ${invitee.getId()} to a PM room. Disconnecting from room`);
+        const store = this.ircBridge.getStore();
+        const [room] = await store.getIrcChannelsForRoomId(event.room_id);
+        await store.removePmRoom(event.room_id);
+        const userId = room.server.getUserIdFromNick(room.channel);
+        const intent = this.ircBridge.getAppServiceBridge().getIntent(userId);
+        await intent.sendMessage(event.room_id, {
+            msgtype: "m.notice",
+            body: "This room has been disconnected from IRC. You cannot invite new users into a IRC PM. Please create a new PM room",
```
jaller94: Shouldn't the last sentence also end on a full stop?
```suggestion
            body: "This room has been disconnected from IRC. You cannot invite new users into a IRC PM. Please create a new PM room.",
```

Half-Shot: Yep.

src/bridge/MatrixHandler.ts
```diff
@@ -336,6 +336,33 @@ export class MatrixHandler {
...
+        );
+        const store = this.ircBridge.getStore();
+        const [room] = await store.getIrcChannelsForRoomId(event.room_id);
+        await store.removePmRoom(event.room_id);
+        const userId = room.server.getUserIdFromNick(room.channel);
+        const intent = this.ircBridge.getAppServiceBridge().getIntent(userId);
+        await intent.sendMessage(event.room_id, {
+            msgtype: "m.notice",
+            body: "This room has been disconnected from IRC. You cannot invite new users into a IRC PM. " +
+                  "Please create a new PM room",
```
jaller94:
```suggestion
                  "Please create a new PM room.",
```

src/bridge/MatrixHandler.ts
```diff
@@ -336,6 +336,33 @@ export class MatrixHandler {
...
+    private async handleInviteToPMRoom(req: BridgeRequest, event: MatrixEventInvite,
+        inviter: MatrixUser, invitee: MatrixUser): Promise<BridgeRequestErr|null> {
+        // We don't support this
+        req.log.warn(
+            `User ${inviter.getId()} tried to invite ${invitee.getId()} to a PM room. Disconnecting from room`
+        );
+        const store = this.ircBridge.getStore();
+        const [room] = await store.getIrcChannelsForRoomId(event.room_id);
+        await store.removePmRoom(event.room_id);
+        const userId = room.server.getUserIdFromNick(room.channel);
```
jaller94: How can you infer the name of a user id from a room channel?
Is that because a PM channel in IRC has the name of the user?

src/bridge/MatrixHandler.ts
```diff
@@ -367,9 +395,13 @@ export class MatrixHandler {
...
-        const hasExistingRoom = (
-            await this.ircBridge.getStore().getIrcChannelsForRoomId(event.room_id)
-        ).length > 0;
+        const rooms = await this.ircBridge.getStore().getIrcChannelsForRoomId(event.room_id);
+        const hasExistingRoom= rooms.length > 1;
+
+        const inviteeIsVirtual = !!this.ircBridge.getServerForUserId(event.state_key);
+        const inviterIsVirtual = !!this.ircBridge.getServerForUserId(event.sender);
+
+        console.log(rooms, event.sender)
```
jaller94: Is `console.log` allowed in production or is this a leftover from debugging?

Half-Shot: leftover :man_facepalming:

src/bridge/MatrixHandler.ts
Half-Shot: Yep. The nickname of the IRC user is the IRC channel name for PMs.

src/bridge/IrcHandler.ts
```diff
@@ -148,19 +148,39 @@ export class IrcHandler {
...
         const response = await this.ircBridge.getAppServiceBridge().getIntent(
             fromUserId
         ).createRoom({
             createAsClient: true,
             options: {
                 name: (fromUserNick + " (PM on " + server.domain + ")"),
                 visibility: "private",
-                preset: "trusted_private_chat",
+                // We deliberately set our own power levels below.
+                // preset: "trusted_private_chat",
```
jaller94: I don't understand this line.
Set our own power levels below *what*? The other user?
Why is the next line commented then? Do we need to keep this?

src/bridge/IrcHandler.ts
```diff
@@ -148,19 +148,39 @@ export class IrcHandler {
      * @return {Promise} which is resolved when the PM room has been created.
      */
     private async createPmRoom (toUserId: string, fromUserId: string, fromUserNick: string, server: IrcServer) {
+        const users: {[userId: string]: number} = { };
+        users[toUserId] = 10;
+        users[fromUserId] = 100;
```
jaller94: 10 and 100 are repeatedly used with the meaning of `lowPowerLevel` and `highPowerLevel` which I think should be constants to write more descriptively and remove potential errors when updating these numbers in the future.

src/bridge/MatrixHandler.ts
jaller94: Can they include any weird characters we may need to transform or escape?

src/bridge/IrcHandler.ts
Half-Shot: Mm, potentially. I think we use these powerlevels in multiple places so it might need a sensible refactor anyway.

src/bridge/MatrixHandler.ts
Half-Shot: getUserIdFromNick should take care of that.

src/bridge/IrcHandler.ts
Half-Shot: Ah, perhaps this needs some clarity. We used to use the [preset `trusted_private_chat`](https://matrix.org/docs/spec/client_server/r0.6.0#post-matrix-client-r0-createroom) for creating a PM room, which basically means both users get powerlevel 100.

Because we want to limit the matrix user's ability to invite into a PM, we deliberately make them have a lower PL and increase the required PL to invite. This is done in the state object below. I've kept this comment and the commented out preset line to indicate that this is what we would be doing, if not for the special requirements.

I tend to prefer leaving this in so as to signal that there was a deliberate reason not to use the preset, but it's a taste thing and the comment can be removed/amended if needed.

PR #1015, opened 2020-03-30
Fix kick when no config is defined for a user.

Half-Shot:
Fixes #1014

PR #1017, opened 2020-03-30
Fix kick when no config is defined for a user - correct branch

Half-Shot:


Half-Shot:
Merging based on okay-ness from #1015

PR #1018, opened 2020-04-02
On name change, inform Matrix users, if their preferred IRC name is taken

Half-Shot:
When a Matrix user "peter" writes `!nick max` to the IRC Bridge, but the name max is already in use, we used to rename them to `peter1` (or another number at the end.

This PR checks if the new name is available and aborts the change if the name is taken.

Screenshot_2020-04-03 Riot appservice-irc localhost

### DIFF
--- a/spec/integ/admin-rooms.spec.js
+++ b/spec/integ/admin-rooms.spec.js
@@ -258,6 +258,15 @@ describe("Admin rooms", function() {
             sentNickCommand = true;
         });
 
+        env.ircMock._whenClient(roomMapping.server, userIdNick, "whois", (client, whoisNick) => {
+            expect(whoisNick).toEqual(newNick);
+            client.emit("error", {
+                commandType: "error",
+                command: "err_nosuchnick",
+                args: [undefined, newNick]
+            });
+        });
+
         // make sure that when a message is sent it uses the new nick
         let sentSay = false;
         env.ircMock._whenClient(roomMapping.server, newNick, "say",
@@ -331,6 +340,15 @@ describe("Admin rooms", function() {
             expect(arg).toEqual(newNick);
             client._changeNick(userIdNick, newNick);
             sentNickCommand = true;
+        });
+
+        env.ircMock._whenClient(roomMapping.server, userIdNick, "whois", (client, whoisNick) => {
+            expect(whoisNick).toEqual(newNick);
+            client.emit("error", {
+                commandType: "error",
+                command: "err_nosuchnick",
+                args: [undefined, newNick]
+            });
         });
 
         // make sure that when a message is sent it uses the new nick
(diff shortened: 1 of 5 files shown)


jaller94:
Still need to adapt and add tests...

src/bridge/IrcBridge.ts
```diff
@@ -926,10 +926,21 @@ export class IrcBridge {
...
     }
 
+    /**
+     * Determines if a nick name already exists.
+     */
     public async checkNickExists(server: IrcServer, nick: string) {
         log.info("Querying for nick %s on %s", nick, server.domain);
         const client = await this.getBotClient(server);
-        return await client.whois(nick);
+        try {
```
Half-Shot, line 935: Hmm, any reason why this is on the `IrcBridge` monolith and not on `BridgedClient` which the user is using to change their nick?

jaller94: The function wasn't used anywhere.
I'll happily move it to `BridgedClient`.

jaller94:
Note for Monday: Task seems correctly done, but the `!nick` tests seem to be failing for some reason.

spec/integ/admin-rooms.spec.js
```diff
@@ -593,6 +647,53 @@ describe("Admin rooms", function() {
         expect(sentAckNotice).toBe(true, "sent ACK m.notice");
     }));
 
+    it("should not try to change to a nickname that is already in use", async () => {
+        const newNick = "Blurple";
+
+        // make sure that the NICK command not is sent
+        let sentNickCommand = false;
+        env.ircMock._whenClient(roomMapping.server, userIdNick, "send",
+        function() {
```
jaller94, line 656: Maybe check if the second parameter `command` equals `NICK`?
What other commands are send via `send`?

spec/integ/admin-rooms.spec.js
```diff
@@ -593,6 +647,53 @@ describe("Admin rooms", function() {
...
+            sentNickCommand = false;
```
jaller94, line 657:
```suggestion
            sentNickCommand = true;
```

src/irc/BridgedClient.ts
```diff
@@ -302,51 +308,78 @@ export class BridgedClient extends EventEmitter {
...
+            );
         }
+
+        return await this.sendNickCommand(validNick);
+    }
+
+    private sendNickCommand(nick: string): Promise<string> {
         if (!this.unsafeClient) {
-            return Promise.reject(new Error("You are not connected to the network."));
+            throw new Error("You are not connected to the network.");
```
jaller94, line 358: In all other cases this function is returning a Promise and not throwing. If this was from m, I wanted to make this an async function, which is tricky with the callbacks below.
```suggestion
            return Promise.reject(new Error("You are not connected to the network."));
```

Half-Shot: Well, if we should probably make the func be async if it returns a promise anyway. Let me do that.

spec/integ/admin-rooms.spec.js
Half-Shot: We shouldn't be sending a nickname request if we know the nick is in use, as per the test title?

spec/integ/admin-rooms.spec.js
Half-Shot: Good catch. I believe only nick

spec/integ/admin-rooms.spec.js
Half-Shot: oh...oh....ah

PR #1020, opened 2020-04-03
Test !nick when the user already has the nick

jaller94:
In preparation of #1018 I want the CI to ring an alarm, if changing the nick to your current nick causes a different notice to appear in Matrix.

### DIFF
--- a/spec/integ/admin-rooms.spec.js
+++ b/spec/integ/admin-rooms.spec.js
@@ -277,6 +277,7 @@ describe("Admin rooms", function() {
         sdk.sendEvent.and.callFake(function(roomId, type, content) {
             expect(roomId).toEqual(adminRoomId);
             expect(content.msgtype).toEqual("m.notice");
+            expect(content.body).toEqual(`Nick changed from '${userIdNick}' to '${newNick}'.`);
             sentAckNotice = true;
             return Promise.resolve();
         });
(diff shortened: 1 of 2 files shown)


spec/integ/admin-rooms.spec.js
```diff
@@ -308,6 +309,77 @@ describe("Admin rooms", function() {
...
+
+            // make sure that the nick command is sent (expected to NOT run)
+            let sentNickCommand = false;
+            env.ircMock._whenClient(roomMapping.server, userIdNick, "send",
+                function (client, command, arg) {
+                    expect(client.nick).toEqual(userIdNick, "use the old nick on /nick");
+                    expect(client.addr).toEqual(roomMapping.server);
+                    expect(command).toEqual("NICK");
+                    expect(arg).toEqual(newNick);
+                    client._changeNick(userIdNick, newNick);
```
Half-Shot, line 325: You can remove these statements if you are just ensuring it doesn't run, fyi.

spec/integ/admin-rooms.spec.js
```diff
@@ -308,6 +309,77 @@ describe("Admin rooms", function() {
         expect(sentSay).toBe(true, "sent say IRC command");
     }));
 
+    it("should be able to keep their name using !nick",
+        test.coroutine(function* () {
```
Half-Shot, line 313: We prefer async functions

```suggestion
        async () {
```

spec/integ/admin-rooms.spec.js
Half-Shot: and below you can exchange the yields for await

spec/integ/admin-rooms.spec.js
jaller94: `test.coroutine(function* ` stays consistent with all other tests in the file. May I find-replace it to async everywhere then?

jaller94: sounds like a separate short PR to me...

PR #1021, opened 2020-04-03
Add the ability to deactivate users to prevent them from ever being bridged

Half-Shot:
A few users have noted that they are re-bridged even after deactivating their account. The bridge will always reconnect users if they join an IRC bridged room. However, due to *state resets* (a bug where older v1-2 version rooms could revert their state to a previous point in time) these users may end up reappearing in the room. I can't seem to find an actual issue that describes the 'state reset bug', but there you go.

If the user reappears in the room, they may get reconnected even if they have since deleted their account. The IRC bridge should therefore keep track of users who have deactivated their accounts. In the future. [MSC2438](https://github.com/matrix-org/matrix-doc/pull/2438) will allow us to listen for these deactivation requests, but for now this offers us a manual way to do it.

### DIFF
--- a/src/irc/ClientPool.ts
+++ b/src/irc/ClientPool.ts
@@ -208,6 +208,10 @@ export class ClientPool {
             return bridgedClient;
         }
 
+        if (userId && await this.ircBridge.getStore().isUserDeactivated(userId)) {
+            throw Error("Cannot create bridged client - user has been deactivated");
+        }
+
         const mxUser = new MatrixUser(userId);
         if (displayName) {
             mxUser.setDisplayName(displayName);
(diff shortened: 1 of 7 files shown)


src/irc/ClientPool.ts
```diff
@@ -208,6 +208,10 @@ export class ClientPool {
             return bridgedClient;
         }
 
+        if (userId && await this.ircBridge.getStore().isUserDeactivated(userId)) {
```
jaller94, line 211: The check for truthiness looks a bit odd here. According to the type annotation the parameter must contain a string and the only falsey string is an emtpy string.

Half-Shot: good point

jaller94:
The changelog file is empty.

Half-Shot:
Good catch

PR #1022, opened 2020-04-07
Drop userId &&

Half-Shot:


Half-Shot:
Not sure why I made this a PR -_-. Dozy.

Issue #671, opened 2018-08-31
We should be kicking on "Closing Link: ... (Bad user info)"


Half-Shot:
https://github.com/freenode/ircd-seven/blob/02023be9f8cdd13937814eb83ed4445b7be5081a/doc/sgml/oper-guide/commands.sgml#L388

Mikaela:
It looks like ircd-seven documentation, here is possibly cleaner highlight:

* https://github.com/freenode/ircd-seven/blob/02023be9f8cdd13937814eb83ed4445b7be5081a/doc/sgml/oper-guide/commands.sgml#L378-#L400

397-398 seems to be the main point.

```
	  The reason is never sent to users. Instead, they will be exited
with <quote>Bad user info</quote>.
```

PR #1023, opened 2020-04-08
Kick on X:line

Half-Shot:
Fixes https://github.com/matrix-org/matrix-appservice-irc/issues/671


jaller94:
This PR caused the issue https://github.com/matrix-org/matrix-appservice-irc/issues/1029 which gets fixed by the PR https://github.com/matrix-org/matrix-appservice-irc/issues/1030.

PR #1025, opened 2020-04-09
npm audit fix 

Half-Shot:


Half-Shot:
(self approving this, as it's an audit)

PR #1030, opened 2020-04-14
Fix issue where failure to get connection with a reasonable failure would kick users

### LINKED ISSUE #1029: Failure to get connection with a reasonable failure kicks users
After a restart of the Freenode bridge on Tue, April 14, 2019 many Matrix users weren't reconnecting to IRC.

Introduced in: https://github.com/matrix-org/matrix-appservice-irc/pull/1023

Chat: https://matrix.to/#/!BAXLHOFjvDKUeLafmO:matrix.org/$uvpjB9FW0qDZmL/Cqg7CZFsfv9u2w6QC/qGImCLn+Ks?via=matrix.org&via=privacytools.io&via=feneas.org

Trace from the server:
```
ERROR:main TypeError: Cannot read property '0' of undefined
at Function.create (/opt/freenodeirc/matrix-appservice-irc-0.17.0-rc1/lib/irc/ConnectionInstance.js:360:35)
```

Half-Shot:
Fixes #1029


jaller94:
Why against `master`? This includes a lot of commits.

Half-Shot:
>Why against master? This includes a lot of commits.

Yeah I cocked it up :/

Half-Shot:
Flaky test is flaky.

src/irc/ConnectionInstance.ts
```diff
@@ -424,7 +424,7 @@ export class ConnectionInstance {
 
                 // Closing Link: gateway/shell/matrix.org/session (Bad user info)
                 // ircd-seven doc link: https://git.io/JvxEs
-                if ((err.args[0] as string|null)?.match(/Closing Link: .+\(Bad user info\)/)) {
+                if (err.args && err.args[0]?.match(/Closing Link: .+\(Bad user info\)/)) {
```
jaller94, line 427: Why the question mark on `err.args[0]?`? You can't run `.match` on `undefined`.
Should we also test if `err.args[0]` is actually defined?

Half-Shot: That's what `?` does :). It's a Typescript feature that returns null early if `args` is not truthy. Hence, err.args[0].match will not be evaluated if [0] is undefined.

src/bridge/MatrixHandler.ts
```diff
@@ -13,14 +13,15 @@ import { toIrcLowerCase } from "../irc/formatting";
...
+        const res = await promise;
         req.resolve(res);
-        return res;
-    }, function(err) {
+    }
+    catch (err) {
         req.reject(err);
         throw err;
-    });
+    }
```
jaller94, line 24: The promise `reqHandler` returns used to resolve in `res` but no longer does this.
The try block or the function block should end on `return res;`.

src/bridge/MatrixHandler.ts
```diff
@@ -856,13 +857,14 @@ export class MatrixHandler {
                         ircRoom.server, event.sender, displayName
                     );
                     await this.sendIrcAction(req, ircRoom, bridgedClient, ircAction, event);
+                    return null;
```
jaller94, line 860: Why is this now returning `Promise<null>` instead of `Promise<undefined>` into `messageSendPromiseSet`?

Half-Shot: leftover from me trying something out, removed.

src/bridge/MatrixHandler.ts
Half-Shot: fixed.

jaller94:
Do you have a test plan? Some steps I can run to verify some issue has been fixed.

Half-Shot:
>Do you have a test plan? Some steps I can run to verify some issue has been fixed.

Yep, turn off the ircd and start the bridge. See if you get kicked from rooms for failing to connect.

src/irc/ConnectionInstance.ts
jaller94: In that case, why is this not `err.args?[0]?.match(…`?

PR #1034, opened 2020-04-16
Use Maps for ClientPool

Half-Shot:
This is mainly to improve readability, as working with a set of nested objects is a little bit of an eyesore. There is also [rumors on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) that iterating/counting and adding/deleting items are quicker on Maps than Objects, so they may be more ideal for our use case.

### DIFF
--- a/src/irc/ClientPool.ts
+++ b/src/irc/ClientPool.ts
@@ -420,7 +420,7 @@ export class ClientPool {
 
         // find the oldest client to kill.
         let oldest: BridgedClient|null = null;
-        for (const client of Object.values(this.virtualClients[server.domain].nicks)) {
+        for (const client of this.virtualClients[server.domain].nicks.values()) {
             if (!client) {
                 // possible since undefined/null values can be present from culled entries
                 continue;
--- a/src/irc/ClientPool.ts
+++ b/src/irc/ClientPool.ts
@@ -80,28 +80,16 @@ export class ClientPool {
         return this.virtualClients[server.domain].pending.has(nick)
     }
 
-    public killAllClients(): Bluebird<void[]> {
-        const domainList = Object.keys(this.virtualClients);
-        let clients: (BridgedClient|undefined)[] = [];
-        domainList.forEach((domain) => {
-            clients = clients.concat(
-                [...this.virtualClients[domain].nicks.values()]
-            );
-
-            clients = clients.concat(
-                [...this.virtualClients[domain].userIds.values()]
-            );
-
-            clients.push(this.botClients.get(domain));
-        });
-
-        const safeClients = clients.filter((c) => Boolean(c)) as BridgedClient[];
-
-        return Bluebird.all(
-            safeClients.map(
-                (client) => client.kill()
-            )
-        );
+    public killAllClients() {
+        return Bluebird.all(Object.keys(this.virtualClients).map((domain) =>
+            [
+                ...this.virtualClients[domain].nicks.values(),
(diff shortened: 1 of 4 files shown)


src/irc/ClientPool.ts
```diff
@@ -84,27 +77,22 @@ export class ClientPool {
...
 
             clients = clients.concat(
-                Object.keys(this.virtualClients[domain].userIds).map(
-                    (userId: string) => this.virtualClients[domain].userIds[userId]
-                )
+                [...this.virtualClients[domain].userIds.values()]
             );
 
-            clients.push(this.botClients[domain]);
+            clients.push(this.botClients.get(domain));
```
jaller94, line 95: why concat? and why re-wrap the output of `.values()` which creates a new array anyway?
```javascript
clients = [
  ...clients,
  ...this.virtualClients[domain].nicks.values(),
  ...this.virtualClients[domain].userIds.values(),
  this.botClients.get(domain),
];
```

src/irc/ClientPool.ts
```diff
@@ -39,30 +39,23 @@ interface ReconnectionItem {
...
+        this.botClients = new Map();
 
         // list of virtual users on servers
         this.virtualClients = { };
 
-        // map of numbers of connected clients on each server
-        // Counting these is quite expensive because we have to
-        // ignore entries where the value is undefined. Instead,
-        // just keep track of how many we have.
-        this.virtualClientCounts = { };
```
jaller94, line 64: Does this no longer need to be initialised?

src/irc/ClientPool.ts
```diff
@@ -500,17 +487,11 @@ export class ClientPool {
 
     public getNickUserIdMappingForChannel(server: IrcServer, channel: string): {[nick: string]: string} {
         const nickUserIdMap: {[nick: string]: string} = {};
-        const cliSet = this.virtualClients[server.domain].userIds;
-        Object.keys(cliSet).filter((userId: string) => {
-            if (!userId) {
-                return false;
+        for (const [userId, client] of this.virtualClients[server.domain].userIds.entries()) {
```
jaller94, line 490: Why is it safe to assume that `this.virtualClients[server.domain]` is defined?

src/irc/ClientPool.ts
```diff
@@ -39,30 +39,23 @@ interface ReconnectionItem {
...
     private ircEventBroker: IrcEventBroker;
     constructor(private ircBridge: IrcBridge, private store: DataStore) {
         // The list of bot clients on servers (not specific users)
-        this.botClients = { };
+        this.botClients = new Map();
 
         // list of virtual users on servers
         this.virtualClients = { };
 
-        // map of numbers of connected clients on each server
```
Half-Shot, line 60: This is no longer expensive, Maps keep their own counters \o/

src/irc/ClientPool.ts
jaller94: Solved by https://github.com/matrix-org/matrix-appservice-irc/pull/1034/files#r409475606

src/irc/ClientPool.ts
Half-Shot: I think we assume that if `server` is defined, then it will have a client pool. I think in this case it's a harmless assumption.

src/irc/ClientPool.ts
Half-Shot: Good catch.

PR #1035, opened 2020-04-16
Authenticate requests to the /provision endpoint

Half-Shot:


src/provisioning/Provisioner.ts
```diff
@@ -132,12 +132,18 @@ export class Provisioner {
...
+                    error: "Bad token supplied,"
+                });
+            }
             if (this.isProvisionRequest(req)) {
                 res.header("Access-Control-Allow-Origin", "*");
                 res.header("Access-Control-Allow-Headers",
                     "Origin, X-Requested-With, Content-Type, Accept");
             }
-            next();
+            return next();
```
jaller94, line 146: What does returning `next()` do? I've never seen that as a pattern.

src/provisioning/Provisioner.ts
```diff
@@ -132,12 +132,18 @@ export class Provisioner {
 
         // Deal with CORS (temporarily for s-web)
         app.use((req, res, next) => {
+            if (!this.ircBridge.getAppServiceBridge().requestCheckToken(req)) {
+                return res.status(403).send({
+                    errcode: "M_FORBIDDEN",
+                    error: "Bad token supplied,"
```
jaller94, line 138:
```suggestion
                    error: "Bad token supplied"
```

src/provisioning/Provisioner.ts
```diff
@@ -132,12 +132,18 @@ export class Provisioner {
 
         // Deal with CORS (temporarily for s-web)
```
jaller94, line 133: This comment which once described the entire purpose of this handler has now become outdated.

Half-Shot:
>When there are many projects already using matrix-appservice-bridge, why don't we release a types/matrix-appservice-bridge package instead of keeping the types updated locally?

We should probably just bundle a definition with the library itself. I've not done it yet due to time

Cogitri:
Hey,

since this is marked as `SECURITY FIX` in the release notes, should this get a CVE? Does this also apply to older releases? Without a CVE (or similiar tracking information) distros don't really know how to act upon this.

Half-Shot:
We're not treating it as high enough severity to warrant a CVE (which is why it didn't get a minor release and has been tacked onto the next release). It only affects users who enabled provisioning and didn't firewall the application out from the rest of the internet. We've determined this to be of low severity because the provisioning system is useless unless you are also running an integration manager (and the feature is off by default), and you'd have to be exposing your appservice to the world.

Cogitri:
Hm, I personally I feel like even for low severity security issues a CVE should be issued, otherwise the chance for people who happen to run the appservice in that configuration (for whatever reason), won't know about that security issue.

PR #1040, opened 2020-04-29
Queue connection failure kicks on the membership queue

Half-Shot:


Half-Shot:
I'm hoping our integration tests are good nuff for this one :)

Issue #311, opened 2016-12-16
Make ConnectionInstance.js:CONNECT_TIMEOUT_MS configurable


Nebukadneza:
As seen in #292, some networks take longer to connect to than others. One example are many IRCNet servers, for example the uni-erlangen one.

I had to manually bump the timeout in:
https://github.com/matrix-org/matrix-appservice-irc/blob/b8c784ff20d31d76e8c3f25a9a1424aca4fa2835/lib/irc/ConnectionInstance.js#L10

It’d be great if this was configurable per-network in the config.

Thanks!
-Dario

4nd3r:
before big next release, can we get this fixed, so I don't have to patchi it every release?
life with IRCnet is hard enough :crying_cat_face: 
or at least raise the default to 60 as quick-and-easy win? pretty please :cowboy_hat_face:

PR #1042, opened 2020-05-01
increase CONNECT_TIMEOUT_MS to 60 seconds

4nd3r:
Until #311, can we please increase this to 60 seconds and make IRCnet users happy?


Half-Shot:
Rather than bumping the limit, how about we just add a config field. It's not a lot of work.

PR #1046, opened 2020-05-06
Ignore the bot user when connection reaping.

Half-Shot:
Fixes:
```
Connection reaping for chat.freenode.net
Found 25486 real users for chat.freenode.net
TypeError: Cannot read property 'split' of null
```


Half-Shot:
MATT!

PR #1048, opened 2020-05-09
Changed docker port numbers to be consistent

tfreedman:
The tutorial uses 9995 externally, but all other documentation uses 9999 externally. The internal port is set to 1234, but this is also wrong as it will assume 9999 internally by default.


Half-Shot:
Sounds good, can you follow https://github.com/matrix-org/matrix-appservice-irc/blob/master/CONTRIBUTING.md#making-pull-requests, which should give you a green tick

PR #1054, opened 2020-05-23
Implement mechanisms to fix powerlevels in rooms if messages fail to bridge

Half-Shot:
We have a number of rooms on matrix.org which have large enough `m.room.power_level` events, that they can no longer be appended to. This causes problems for users when they join bridged rooms, as it means they might be stuck unable to talk if they are not voiced, as well as messages from IRC failing in the same way.

To combat this, there are two mechanisms put in place here:
- If the failure code from setting a powerlevel is `M_TOO_LARGE`, we try to filter out left users from the event, to try to pull back some space. Users should be dropped from this event anyway, but due to disruptions from restarts, to netsplits and homeserver issues, this cannot be relied upon.
- If sending a message from IRC to Matrix fails often enough and it's due to a powerlevel issue, then the modes for the room are rechecked to ensure that the moderation bit hasn't been left on and is causing issues.

Hopefully between these two mechanisms, we can try to make more rooms usable. Unfortunately, IRC rooms with genuinely complicated powerlevels (i.e. having 1000s of members with voice) aren't possible to be bridged as it is difficult to refactor the powerlevel system to handle that. Maybe one day Matrix rooms will increase the size of power level events to handle this, or offer alternative authorisation methods like groups. For now though, this is the best solution.


src/bridge/IrcHandler.ts
```diff
@@ -623,16 +623,29 @@ export class IrcHandler {
...
-            promises.push(
-                this.ircBridge.sendMatrixAction(room, virtualMatrixUser, mxAction)
-            );
+            try {
+                await this.ircBridge.sendMatrixAction(room, virtualMatrixUser, mxAction);
+            }
+            catch (ex) {
+                // Check if it was a permission fail.
+                if (ex.data?.errcode === "M_FORBIDDEN" &&
+                    ex.data?.error?.startsWith("You don't have permission to post that to the room.")) {
```
jaller94, line 637: Is this string required in homeservers from the spec?
Comparing error messages seems like it's testing against a specific implementation. Is the error code not enough?

src/bridge/RoomAccessSyncer.ts
```diff
@@ -265,6 +291,26 @@ export class RoomAccessSyncer {
         await botClient.sendStateEvent(roomId, "m.room.power_levels", plContent, "");
     }
 
+    public async onFailedMessage(req: BridgeRequest, server: IrcServer, channel: string) {
+        const key = `${server.getNetworkId()}:${channel}`;
+        const currThreshold = this.accessRefreshCount.get(key) || 0;
```
jaller94, line 296: `currThreshold` should be named `currRefreshCount`.
It's not a threshold as far as I can tell.

jaller94, line 296:
```suggestion
        const currThreshold = (this.accessRefreshCount.get(key) || 0) + 1;
```
And then remove the +1s in the following lines for better readability.
I found it confusing why the current count (which I assumed included the current failure) needed to be one higher.

src/bridge/IrcHandler.ts
Half-Shot: The problem is they might also not be able to join the room because the join rules have changed, which we don't want to treat the same way. It's hard though, because there isn't an easy way in the spec to distinguish.

jaller94 (approved):
Looks good to me.

I made ESLint warnings not fail CI (even though we're currently at 0 warnings).

package.json
```diff
@@ -11,7 +11,7 @@
     "prepare": "npm run build",
     "build": "tsc --project ./tsconfig.json",
     "test": "BLUEBIRD_DEBUG=1 node --max_old_space_size=3072 node_modules/jasmine/bin/jasmine.js --stop-on-failure=true",
-    "lint": "eslint -c .eslintrc --max-warnings 0 src/**/*.ts",
```
Half-Shot, line 14: I totally misunderstood you in bridge crew. We want to keep this for now, as we're trying to keep the warnings at 0.

PR #1057, opened 2020-05-26
Add org.matrix.message-length to the bridge state

### LINKED ISSUE #701: "Long messages" are annoying; prefer to disallow entering them on the Matrix UI
When a Matrix user is connected to a channel which is bridged to IRC, limitations on message length should be communicated to the Matrix client, which should enforce them. I find the "long message, click here to read" links incredibly irritating and won't be reading them, and may ban Matrix users from my channels if they use them.

Half-Shot:
Fixes #701 in conjunction with https://github.com/matrix-org/matrix-react-sdk/pull/4642

This PR leverages the existing bridge state event by adding a warning about long messages for matrix users.

MSC pending


Half-Shot:
The web PR has long since rotted away, and I think if we were going to do this again there are better MSCs out there to follow.

PR #1060, opened 2020-06-01
Add configurable limit to the number of rooms that can be provisioned

Half-Shot:
This does some basic checks to see if the bridge has reached the configured maximum number of rooms when provisioning. It also adds a endpoint to check the current number and limit.

### DIFF
--- a/src/bridge/IrcBridge.ts
+++ b/src/bridge/IrcBridge.ts
@@ -1113,7 +1113,7 @@ export class IrcBridge {
     }
 
     public async atBridgedRoomLimit() {
-        const limit = this.config.ircService.provisioning.roomLimit;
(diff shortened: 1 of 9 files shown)


src/bridge/IrcBridge.ts
```diff
@@ -1111,4 +1111,13 @@ export class IrcBridge {
         }
         logCb(`Quit ${offlineCount}/${users.length}`);
     }
+
+    public async atBridgedRoomLimit() {
+        const limit = this.config.ircService.provisioning.roomLimit;
```
jaller94, line 1116: `provisioning` is not required.
```suggestion
        const limit = this.config.ircService.provisioning?.roomLimit;
```

PR #1062, opened 2020-06-04
Add external sasl with with key/cert (tested on freenode)

Francesco149:
NOTE: this requires https://github.com/matrix-org/node-irc/pull/48 to be merged to build correctly (it references that commit in the package-lock)

this allows authenticating through sasl external on services like freenode by specifying cert and key and setting saslMode to EXTERNAL. tested on freenode from my ec2 instance

see the node-irc pull for more info

this also gets around the issue of having a hardcoded bot name since the external auth on freenode doesn't care if your nickname matches


src/irc/ConnectionInstance.ts
```diff
@@ -382,8 +384,11 @@ export class ConnectionInstance {
             retryCount: 0,
             family: server.getIpv6Prefix() || server.getIpv6Only() ? 6 : null,
             bustRfc3484: true,
-            sasl: opts.password ? server.useSasl() : false,
```
Half-Shot, line 385: one concern that a blank password can now be incorrectly used with sasl. We should probably catch and throw this?

Francesco149:
i added the check and rebased to develop. not tested but only the 3 line check for the password and the node-irc commit hash have changed

Francesco149:
rebased again, tested on freenode and can confirm it works great

Half-Shot (changes requested):
Code looks fine, but CI needs fixing (check the tests and linting). You'll also need to sign off as per https://github.com/matrix-org/matrix-appservice-irc/blob/develop/CONTRIBUTING.md#making-pull-requests

PR #1065, opened 2020-06-12
Allow global membersync.initial to be false but still sync named roomIds

### LINKED ISSUE #1063: membershipLists inital for specific rooms is not working [M -> I]
We want to mirror all users that are in one specific matrix to IRC so that virtual IRC Users are created for everyone who is able to read messages. Our configuration looks like this:

```yaml
      membershipLists:
        enabled: true
        floodDelayMs: 10000
        global:
          ircToMatrix:
            initial: false
            incremental: false
          matrixToIrc:
            initial: false
            incremental: false
        rooms:
          - room: "!yMromrRUzExxxxxxxx:our.matrix.tld"
            matrixToIrc:
              initial: true
              incremental: true
```

### Expected behavior

We expect, upon restarting the appservice using the new configuration, to join all matrix users which are in the specified room (`!yMromrRUzExxxxxxxx:our.matrix.tld`) to be joined in to the IRC counterpart. We also expect users which join and or leave to be joined in to the IRC room.

### Actual behavior

The `incremental` option works (joins/and leaves are being synchronized) while the room-specific `inital` option doesn't seem to have any effect. If we turn set `initial: true` globally it works as expected.

This may be related to #371 

- Version: latest docker container (docker pull matrixdotorg/matrix-appservice-irc:latest)
- IRC Network: Hackint

Half-Shot:
Fixes #1063

### DIFF
--- a/src/bridge/MemberListSyncer.ts
+++ b/src/bridge/MemberListSyncer.ts
@@ -77,18 +77,14 @@ export class MemberListSyncer {
             log.info("%s does not have membership list syncing enabled.", server.domain);
             return;
         }
-        if (!server.shouldSyncMembershipToIrc("initial")) {
-            log.info("%s shouldn't sync initial memberships to irc.", server.domain);
-            return;
-        }
         log.info("Checking membership lists for syncing on %s", server.domain);
         let start = Date.now();
         const rooms = await this.getSyncableRooms();
         log.info("Found %s syncable rooms (%sms)", rooms.length, Date.now() - start);
-        this.leaveIrcUsersFromRooms(rooms, server);
+        this.leaveIrcUsersFromRooms(rooms);
         start = Date.now();
         log.info("Joining Matrix users to IRC channels...");
-        await this.joinMatrixUsersToChannels(rooms, server, this.injectJoinFn);
+        await this.joinMatrixUsersToChannels(rooms, this.injectJoinFn);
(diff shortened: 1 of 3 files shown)


src/bridge/MemberListSyncer.ts
```diff
@@ -251,13 +250,13 @@ export class MemberListSyncer {
...
 
-    private joinMatrixUsersToChannels(rooms: RoomInfo[], server: IrcServer, injectJoinFn: InjectJoinFn) {
+    private joinMatrixUsersToChannels(rooms: RoomInfo[], injectJoinFn: InjectJoinFn) {
         const d = promiseutil.defer();
 
         // filter out rooms listed in the rules
         const filteredRooms: RoomInfo[] = [];
         rooms.forEach((roomInfo) => {
-            if (!server.shouldSyncMembershipToIrc("initial", roomInfo.id)) {
+            if (!this.server.shouldSyncMembershipToIrc("initial", roomInfo.id)) {
```
jaller94, line 259: The constructor does not store `server` in `this.server`.

https://github.com/matrix-org/matrix-appservice-irc/blob/fb5cc2beac11c6758873fcbc8f534553b30a3cda/src/bridge/MemberListSyncer.ts#L53

Is this some language "feature" to silently add unused parameters of the constructor to the object?

Half-Shot: That's exactly what that does :). `private|public` before a parameter makes it a class property and stores it. Pretty handy

PR #1066, opened 2020-06-12
Update sanitize html package

### LINKED ISSUE #1003: sanitizeHtml is broken
I looked into this for a while, and it looks like typings are broken and haven't yet been fixed. Keeping this issue open as a placeholder so we can fix it, when they fix it.

https://github.com/apostrophecms/sanitize-html/issues/320

Half-Shot:
Fixes #1003

### DIFF
--- a/changelog.d/1066.misc
+++ b/changelog.d/1066.misc
@@ -1,0 +1,1 @@
+Update `sanitizeHtml` package.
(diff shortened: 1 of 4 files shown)


jaller94:
Something's wrong with the Docker step, but it only shows me HTTP 404.

Half-Shot:
>Something's wrong with the Docker step, but it only shows me HTTP 404.

I had to go into docker and look at the logs of the builds (you might need access for that, ops can help). Anyway, my package-lock is out of sync.

PR #1069, opened 2020-06-25
Add metrics worker

Half-Shot:
This workerizes the metrics worker *if* the configuration specifies a custom port for the metrics to be listened on. The usefulness of this is that the metrics endpoint will not timeout should the main event loop be chocked up by IRC requests, since worker threads run their own event loop.

In the future I'd like to split more work out to workers, but for now this should be a good start.

### DIFF
--- a/src/bridge/IrcBridge.ts
+++ b/src/bridge/IrcBridge.ts
@@ -36,6 +36,7 @@ import { BridgeConfig } from "../config/BridgeConfig";
 import { MembershipQueue } from "../util/MembershipQueue";
 import { BridgeStateSyncer } from "./BridgeStateSyncer";
 import { Registry } from "prom-client";
+import { spawnMetricsWorker } from "../workers/MetricsWorker";
 
 const log = getLogger("IrcBridge");
 const DEFAULT_PORT = 8090;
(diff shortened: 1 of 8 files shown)


src/bridge/IrcBridge.ts
```diff
@@ -183,8 +185,22 @@ export class IrcBridge {
...
+
+        const metrics = this.bridge.getPrometheusMetrics(!usingRemoteMetrics, registry);
+
+        if (this.config.ircService.metrics.port) {
+            log.info(`Spawned metrics worker on http://${this.config.ircService.metrics.host}:${this.config.ircService.metrics.port}`);
+            spawnMetricsWorker(
+                this.config.ircService.metrics.port,
+                this.config.ircService.metrics.host,
+                () => {
+                    return registry.metrics();
```
jaller94, line 202: Can the third parameter just be `registry.metrics`?
I don't see why it needs to be wrapped in a function. Please use `.bind()` if that's the intention behind this.

src/workers/MetricsWorker.ts
```diff
@@ -0,0 +1,73 @@
...
+        });
+
+        parentPort?.postMessage("metricsdump");
+    }).listen(workerData.port, workerData.hostname, 1);
+}
+
+export function spawnMetricsWorker(port: number, hostname = "127.0.0.1", onMetricsRequested: () => string) {
+    const worker = new Worker(__filename, { workerData: { port, hostname } });
+    const workerLogger = getLog("MetricsWorker");
+    worker.on("message", (msg) => {
```
jaller94, line 63:
```suggestion
    worker.on("message", (msg: string) => {
```
I don't see how TypeScript implies this type.

PR #1070, opened 2020-06-25
Bump minimum NodeJS version to 12.x

Half-Shot:


Half-Shot:
(Node 10 tests are expected to fail until the linked PR is merged)

dalcde:
What was the motivation for bumping to node 12? It seems to work on node 10 with the `--experimental-worker` flag, with all tests passing.

Node 10 is the version that comes with the latest LTS Ubuntu. Being able to run on node 10 would be very helpful.

PR #1071, opened 2020-06-26
Report bridge version in metrics

Half-Shot:
A simple label hack to show the bridge version, so we can show it in grafana.

### DIFF
--- a/src/bridge/IrcBridge.ts
+++ b/src/bridge/IrcBridge.ts
@@ -37,6 +37,7 @@ import { MembershipQueue } from "../util/MembershipQueue";
 import { BridgeStateSyncer } from "./BridgeStateSyncer";
 import { Registry } from "prom-client";
 import { spawnMetricsWorker } from "../workers/MetricsWorker";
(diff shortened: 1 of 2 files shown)


src/bridge/IrcBridge.ts
```diff
@@ -274,6 +275,12 @@ export class IrcBridge {
             labels: ["server"]
         });
 
+        metrics.addCounter({
+            name: "app_version",
+            help: "Version number of the bridge",
+            labels: ["version"],
+        }).inc({ version: getBridgeVersion()}, 1);
```
jaller94, line 282: Why use a Counter for the bridge version?
Will this show how often a bridge has been started with a specific version?

Half-Shot: No, it's a dummy value of 1. Prometheus doesn't support strings as values, so we're using a constant counter of 1 with an attached label.

Half-Shot: The logic is similar to https://github.com/siimon/prom-client/blob/master/lib/metrics/version.js

PR #1072, opened 2020-06-26
Limit number of kicked users, and order by inactive time for reapUsers

Half-Shot:


src/DebugApi.ts
```diff
@@ -199,13 +199,15 @@ export class DebugApi {
...
+        const limit = query["limit"] !== undefined ? parseInt(query["limit"] as string) : undefined;
         const reason = query["reason"] as string;
         const dry = query["dryrun"] !== undefined && query["dryrun"] !== "false";
         const defaultOnline = (query["defaultOnline"] ?? "true") === "true";
         const excludeRegex = query["excludeRegex"] as string;
         this.ircBridge.connectionReap(
-            msgCb, server, since, reason, dry, defaultOnline, excludeRegex
+            msgCb, server, since, reason, dry, defaultOnline, excludeRegex, limit,
         ).catch((err: Error) => {
+            // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
jaller94, line 210: Why is this all of a sudden required without a change to eslint?

src/DebugApi.ts
```diff
@@ -199,13 +199,15 @@ export class DebugApi {
         };
         const server = query["server"] as string;
         const since = parseInt(query["since"] as string);
+        const limit = query["limit"] !== undefined ? parseInt(query["limit"] as string) : undefined;
```
jaller94, line 202: `limit` as in maxKickedUsers?
The term limit could apply to many different restrictions and I wish the naming was more specific.

src/DebugApi.ts
Half-Shot: My IDE complained about it, and I submitted to it's demands.

src/DebugApi.ts
Half-Shot: hmm, how about targetUserCount ?

Half-Shot: I chose limit specifically because it's easier to type into curl, tbh :p

Half-Shot: Have settled for `targetCount` which is shorter but more descriptive than `limit`.

PR #1074, opened 2020-06-29
Refactor room creation to use 

Half-Shot:
(Not sure what's going on with the git history here)

This PR refactors the tracking of channels and subsequent creation of bridge rooms into a single function, whereas we had 3 separate bits of code doing near identical things.


src/bridge/RoomCreation.ts
```diff
@@ -0,0 +1,96 @@
...
+                history_visibility: "joined"
+            }
+        }
+    ];
+    if (server.areGroupsEnabled()) {
+        initialState.push({
+            type: "m.room.related_groups",
+            state_key: "",
+            content: {
+                groups: server.getGroupId(),
```
jaller94, line 47: This changed from an array to a string. Is that intended?

Half-Shot: No, oops. Good find.

PR #1076, opened 2020-06-29
Code improvements: Simplify use of Map and RegEx methods

jaller94:
* Use `map.forEach()` instead of iterating a map into an unused array.
* Use `regex.test()` when we don't care about the match.


jaller94:
@Half-Shot Why is CI not crying about the changelog missing? Is it not a requirement yet?

Half-Shot:
It should be :/

PR #1078, opened 2020-07-01
Make ESLint stricter

jaller94:
* Inherit rules from "eslint-recommended".
* Add the env "es6" for globals like "Promise" (that's not included in setting `ecmaVersion` to 9)
* Prefer `array.includes()` over `array.indexOf() !== 1`.

### DIFF
--- a/src/bridge/AdminRoomHandler.ts
+++ b/src/bridge/AdminRoomHandler.ts
@@ -158,11 +158,11 @@ export class AdminRoomHandler {
             case "!help":
                 await this.showHelp(adminRoom);
                 break;
-            default:
+            default: {
                 const notice = new MatrixAction("notice",
                 "The command was not recognised. Available commands are listed by !help");
                 await this.ircBridge.sendMatrixAction(adminRoom, this.botUser, notice);
-                break;
+            }
         }
     }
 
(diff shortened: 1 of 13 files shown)


src/bridge/AdminRoomHandler.ts
```diff
@@ -158,11 +158,11 @@ export class AdminRoomHandler {
             case "!help":
                 await this.showHelp(adminRoom);
                 break;
-            default:
+            default: {
```
Half-Shot, line 161: What's up with this rule, the others don't need brackets?

src/irc/ConnectionInstance.ts
```diff
@@ -188,7 +188,6 @@ export class ConnectionInstance {
                 return;
             }
             // do the callback
-            // eslint is usually confused about IArguments
```
Half-Shot, line 191: We've removed the comment but not the disable line?

src/bridge/AdminRoomHandler.ts
Half-Shot: Oh this is for multi-line? Fair enough.

jaller94: It's for declaring variables inside a `case`, because case is not a block so it will be defined for the entire `switch`.
https://eslint.org/docs/rules/no-case-declarations

src/irc/ConnectionInstance.ts
jaller94: Nothing changed. The comment was outdated.
The JS variable `arguments` has the type `IArguments` but when collecting parameters as `...args` it's just an array.

Half-Shot: But below we have `fn.apply(fn, args as any);` which isn't doing ...args? Should we also fix that?

jaller94:
```
function foo(...args) {
  //args is an Array<any>
}

function bar() {
  //arguments (defined in every function as part of the JS spec) is an IArguments
}
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments

PR #1087, opened 2020-07-23
Media filenames

### LINKED ISSUE #1085: URLs of uploaded images should have a file extension
Currently the IRC equivalent of Matrix images looks like this:

> Matrixuser uploaded an image: image.png (26KB)
> < https://matrix.org/_matrix/media/r0/download/example.org/qXNYnYjcyfyxXwEQkkQPlsxi >

Modern IRC Clients like Textual parse text to highlight the URLs and if the URLs seems to be an image download and display the image inline. One of the criteria for displaying URL content inline is a file extension, e.g. https://example.org/image.png​. Matrix URLs are missing that extension, making the experience of reading Matrix users in an IRC channel a little more annoying.

Half-Shot:
Fixes #1085 

This PR includes the filename (if given) onto a media link. It will also call all long message uploads "message.txt".

### DIFF
--- a/src/models/MatrixAction.ts
+++ b/src/models/MatrixAction.ts
@@ -155,7 +155,7 @@ export class MatrixAction {
         }
     }
 
-    public static fromEvent(event: MatrixMessageEvent, mediaUrl: string) {
+    public static fromEvent(event: MatrixMessageEvent, mediaUrl: string, forceFilename?: string) {
         event.content = event.content || {};
         let type = EVENT_TO_TYPE[event.type] || "message"; // mx event type to action type
         let text = event.content.body;
@@ -179,7 +179,14 @@ export class MatrixAction {
                     fileSize = " (" + Math.round(event.content.info.size / 1024) + "KB)";
                 }
 
-                const url = ContentRepo.getHttpUriForMxc(mediaUrl, event.content.url);
+                let url = ContentRepo.getHttpUriForMxc(mediaUrl, event.content.url);
+                if (forceFilename) {
+                    url += `/${forceFilename}`;
+                }
+                else if (event.content.body?.match(/\.[\w\d]{2,4}$/)) {
+                    // Add filename to url if body is a filename.
+                    url += `/${event.content.body}`;
+                }
                 text = `${event.content.body}${fileSize} < ${url} >`;
             }
         }
--- a/src/models/MatrixAction.ts
+++ b/src/models/MatrixAction.ts
@@ -155,7 +155,7 @@ export class MatrixAction {
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/models/MatrixAction.ts
```diff
@@ -179,7 +179,14 @@ export class MatrixAction {
                     fileSize = " (" + Math.round(event.content.info.size / 1024) + "KB)";
                 }
 
-                const url = ContentRepo.getHttpUriForMxc(mediaUrl, event.content.url);
+                let url = ContentRepo.getHttpUriForMxc(mediaUrl, event.content.url);
+                if (forceFilename) {
+                    url += `/${forceFilename}`;
+                }
+                else if (event.content.body?.match(/\.[\w\d]{2,4}$/)) {
```
jaller94, line 186: That's a pretty generous RegExp for filenames,

`I am not a file name / but I match 3.1415`
🙃

How about /^\S\.[\w\d]{2,4}$` to rule out whitespaces (yes, some of which are legal in filenames)?
Well, honestly... whatever. This is trying to simplify very complicated, OS-dependent rules. 🙈

jaller94, line 186: `RegExp.test()` is slightly faster and less confusing, if you're not interested in the match.
```suggestion
                else if (/\.[\w\d]{2,4}$/.test(event.content.body)) {
```

src/models/MatrixAction.ts
```diff
@@ -179,7 +179,14 @@ export class MatrixAction {
                     fileSize = " (" + Math.round(event.content.info.size / 1024) + "KB)";
```
jaller94, line 179:
```suggestion
                    fileSize = " (" + Math.round(event.content.info.size / 1024) + "KiB)";
```
Divided by 1024, so it's [KiB](https://en.wikipedia.org/wiki/Kibibyte).

You didn't change this.. so feel free to opt for "Christian, please file this as a separate issue".

src/models/MatrixAction.ts
Half-Shot: Yep, I understand. This is just a very simple heuristic to get a filename for the file if we can. I am not terribly happy that the spec doesn't have an explicit field for filename and caption, as some clients use `body` for descriptive text.

src/models/MatrixAction.ts
Half-Shot: Nah, this is good.

src/models/MatrixAction.ts
Half-Shot: On the other hand, having a filename with spaces in it will *only* mean we don't bother with a file name, which isn't that common anyway. Let's use yours.

PR #1097, opened 2020-08-05
Replace .indexOf with more specific methods

jaller94:
I would like to replace `.indexOf()` with `.startsWith()` and `.includes()` where that's the intended logic.


Half-Shot:
@jaller94 could you detangle this from it's parent PR?

PR #1098, opened 2020-08-05
Make sure a killed BridgedClient is dead, even if connect was never called (continuation)

jaller94:
Copy of #976 with additional commits.


jaller94:
I reviewed the #976 several times and think it's safe to merge.
I'm not 100% sure it fixes the issue #944, but it's a safer approach to handling connections.

spec/integ/provisioning.spec.js
```diff
@@ -96,7 +96,7 @@ describe("Provisioning API", function() {
                     if (content.status === "failure") {
                         env.isFailed.resolve();
                     }
-                    else if (content.status == "success") {
+                    else if (content.status === "success") {
```
Half-Shot, line 99: in the future these should definitely be in a separate PR. These tests tend to be flaky and if they regress, it'd be hard to track down which PR did it.

src/irc/BridgedClient.ts
```diff
@@ -58,16 +58,34 @@ export interface BridgedClientLogger {
...
 
+export enum BridgedClientStatus {
+    CREATED,
+    CONNECTING,
+    CONNECTED,
+    DEAD,
+    KILLED,
+}
+
+interface NotConnected {
```
Half-Shot, line 69: I'm not really convinced by these Connected/NotConnected interfaces. I'd rather status,client,inst were all top level. If we must wrap it, does ConnectionInstance suit that purpose as a wrapper for the other two values?

src/DebugApi.ts
```diff
@@ -277,8 +277,8 @@ export class DebugApi {
         // wait 3s to pool responses
         return Bluebird.delay(3000).then(function() {
             // unhook listener to avoid leaking
-            if (client.unsafeClient) {
-                client.unsafeClient.removeListener("raw", listener);
+            if (client.state.status === BridgedClientStatus.CONNECTED) {
```
Half-Shot, line 280: `client.state.status` feels verbose. We should probably at least expose `get status()` on the client.

spec/integ/provisioning.spec.js
jaller94: The ESLint `eqeqeq` rule became an error in a separate PR, which got merged after this one got opened.
https://github.com/matrix-org/matrix-appservice-irc/pull/1078/files#diff-1dc6ee56b778cd91e0327b52aaeaa1b9R87

src/DebugApi.ts
tilosp: i am not sure that typescript will allow access to `client.state.client` then

src/irc/BridgedClient.ts
tilosp: this allows the the removal of the ugly `null` checks that were there before on `inst` and `unsafeClient`.

It's a pattern to get rust like enums, that contain values in some variants.

src/DebugApi.ts
Half-Shot: To be honest I've never been happy with bits of the bridge accessing unsafeClient to begin with. I've switched it around so that BridgedClient exposes methods for each of these calls.

src/irc/BridgedClient.ts
tilosp: I am not sure how you would want to use ConnectionInstance to achieve the same level of type safety (Ensuring that a ConnectionInstance is guaranteed to be there if the status is `CONNECTED`).

Half-Shot: In the end I evaluated it again and I'm going to go with the approach, but I'm not going to expose client/inst outside of `BridgedClient`. Code that makes use of those will use methods on BridgedClient and we can do some safety checks there.

src/irc/BridgedClient.ts
```diff
@@ -793,14 +795,7 @@ export class BridgedClient extends EventEmitter {
             if (reason === "banned") {
                 // If we've been banned, this is intentional.
                 this._explicitDisconnect = true;
-            }
-
-            // make sure a killed connection stays killed
-            if (this.state.status !== BridgedClientStatus.KILLED) {
-                this._state = {
-                    status: BridgedClientStatus.DEAD,
```
tilosp: I am not sure why you removed this, i think now the status will stay on `CONNECTED` if the bridge get disconnected from irc because of some network error.

Half-Shot: Because I read `this.state.status === BridgedClientStatus.KILLED` :facepalm:. Now it makes sense

src/bridge/MatrixHandler.ts
```diff
@@ -905,14 +905,11 @@ export class MatrixHandler {
             this.eventCache.delete(delKey);
         }
 
-        // Check for the existance of the getSplitMessages method.
-        if (!(ircClient.unsafeClient && ircClient.unsafeClient.getSplitMessages)) {
-            await this.ircBridge.sendIrcAction(ircRoom, ircClient, ircAction);
-            return;
```
jaller94, line 911: Instead of waiting for a connection, we formerly send an action and returned here.
Is this change expected?

(The also makes me think if this code is called a ton during a netsplit and queuing many thousands of actions.)

Half-Shot: Yeah, so I believe this in itself was a bug because getSplitMessages would always be defined on unsafeClient if unsafeClient was defined. I think what likely happens here is that it fails the request because the client isn't ready yet.

That being said, I made this change because of a test failure. That test wasn't failing before.

Half-Shot: Oh hang on I see `!(ircClient.unsafeClient && ircClient.unsafeClient.getSplitMessages)` would evaluate to true if `unsafeClient` is undefined. Which presumably is not what should be happening here. That'd explain why the test used to pass.

Half-Shot: >(The also makes me think if this code is called a ton during a netsplit and queuing many thousands of actions.)

Shouldn't be, it's M->I direction so it's only called on new events

jaller94: Isn't a netsplit causing many many leaves and joins on the Matrix side (caused by the bridge)?

Half-Shot: No, netsplits are many leaves and joins on the irc side. They don't reflect back across the bridge though, so the M->I path is not affected.

src/irc/BridgedClient.ts
```diff
@@ -465,7 +465,7 @@ export class BridgedClient extends EventEmitter {
         if (this.state.status !== BridgedClientStatus.CONNECTED) {
             return Promise.resolve(); // we were never connected to the network.
         }
-        if (Object.keys(this.state.client.chans).indexOf(channel) === -1) {
+        if (Object.keys(this.state.client.chans).includes(channel)) {
```
tilosp, line 468: i think this should be inverted

jaller94: yes, my bad

jaller94:
I'm happy with the PR and think it looks a lot cleaner. :) :+1:

PR #1105, opened 2020-08-12
Add user activity metrics

Half-Shot:
This should be equivalent to https://github.com/matrix-org/matrix-appservice-slack/pull/380 (sans rooms)

We're basically just cribbing off our postgres data and inserting it into prometheus

### DIFF
--- a/src/bridge/IrcBridge.ts
+++ b/src/bridge/IrcBridge.ts
@@ -45,6 +45,7 @@ const DELAY_TIME_MS = 10 * 1000;
 const DELAY_FETCH_ROOM_LIST_MS = 3 * 1000;
 const DEAD_TIME_MS = 5 * 60 * 1000;
 const TXN_SIZE_DEFAULT = 10000000 // 10MB
+export const METRIC_ACTIVE_USERS = "active_users";
 
 export class IrcBridge {
     public static readonly DEFAULT_LOCALPART = "appservice-irc";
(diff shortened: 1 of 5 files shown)


src/bridge/IrcBridge.ts
```diff
@@ -296,6 +303,31 @@ export class IrcBridge {
...
+                        else if (appserviceBot.isRemoteUser(user.user_id)) {
+                            remote++;
+                        }
+                        else {
+                            matrix++;
+                        }
+                    }
+                    activeUsers.set({remote: "true"}, remote);
+                    activeUsers.set({remote: "false"}, matrix);
+                });
```
jaller94, line 328: Needs a `.catch()`, otherwise we may end up with an unhandled promise.

src/bridge/IrcBridge.ts
```diff
@@ -685,9 +717,12 @@ export class IrcBridge {
     private async _onEvent (baseRequest: Request): Promise<BridgeRequestErr|undefined> {
         const event = baseRequest.getData();
         let updatePromise: Promise<void>|null = null;
-        if (event.sender && this.activityTracker) {
+        if (event.sender && (this.activityTracker ||
+            this.config.ircService.metrics.userActivityThresholdHours !== undefined)) {
```
jaller94, line 721: * `ircService` may not be defined according to the schema.
* `metrics` may not be defined according to the schema.

src/bridge/IrcBridge.ts
```diff
@@ -266,6 +267,12 @@ export class IrcBridge {
             labels: ["server"]
         });
 
+        const activeUsers = metrics.addGauge({
+            name: METRIC_ACTIVE_USERS,
+            help: "Numer of users actively using the bridge.",
```
jaller94, line 272:
```suggestion
            help: "Number of users actively using the bridge.",
```

src/bridge/IrcBridge.ts
Half-Shot: >ircService may not be defined according to the schema.

That's bad, it definitely should be defined.

> metrics may not be defined according to the schema.

That sound possible, our types are wrong.

PR #1107, opened 2020-08-12
Fix metrics

Half-Shot:


src/bridge/IrcBridge.ts
```diff
@@ -198,11 +198,9 @@ export class IrcBridge {
         const usingRemoteMetrics = !!this.config.ircService.metrics.port;
 
         const metrics = this.bridge.getPrometheusMetrics(!usingRemoteMetrics, registry);
-
+        let metricsUrl = `${this.config.homeserver.bindHostname || "0.0.0.0"}:${this.config.homeserver.bindPort}`;
         if (this.config.ircService.metrics.port) {
-            log.info(
-            `Started metrics on http://${this.config.ircService.metrics.host}:${this.config.ircService.metrics.port}`
-            );
+            metricsUrl = `${this.config.ircService.metrics.host}:${this.config.ircService.metrics.port}`;
```
jaller94, line 203: I think `metrics`, `host` and `port` are all optional, aren't they?

src/bridge/IrcBridge.ts
```diff
@@ -198,11 +198,9 @@ export class IrcBridge {
         const usingRemoteMetrics = !!this.config.ircService.metrics.port;
 
         const metrics = this.bridge.getPrometheusMetrics(!usingRemoteMetrics, registry);
-
+        let metricsUrl = `${this.config.homeserver.bindHostname || "0.0.0.0"}:${this.config.homeserver.bindPort}`;
```
jaller94, line 201: Is bindPort required?

src/bridge/IrcBridge.ts
Half-Shot: We check metrics further up, port is checked but the hostname could be missing

Half-Shot:
Tried to unmangle the url stuff

PR #1109, opened 2020-08-12
Add metrics worker safety

### LINKED ISSUE #1089: Spinning master thread takes out metrics thread
Despite them being hopefully seperated, we've noticed that metrics will stop being reported if the master thread goes down. Since the http listener and metrics are all hosted on the metrics worker, with appropriate timeouts on scraping the master thread - it's not clear why it's not working.

Half-Shot:
Fixes #1089 

A few things to make writes safer, as the metrics worker died while the master was heavily saturated.

### DIFF
--- a/src/workers/MetricsWorker.ts
+++ b/src/workers/MetricsWorker.ts
@@ -2,7 +2,7 @@ import { Worker, isMainThread, parentPort, workerData } from "worker_threads";
 import { createServer, ServerResponse } from "http";
 import { Registry, Gauge } from "prom-client";
 import getLog from "../logging";
-const METRICS_DUMP_TIMEOUT_MS = 5000;
+const METRICS_DUMP_TIMEOUT_MS = 20000;
 
 function writeLog(level: string, msg: string) {
     return parentPort?.postMessage(`log:${level}:${msg}`);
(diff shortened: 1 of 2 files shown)


src/workers/MetricsWorker.ts
```diff
@@ -22,6 +22,15 @@ function workerThread() {
         throw Error("Missing parentPort");
     }
 
+    const writeAndEnd = (res: ServerResponse, data: string) => {
+        if (res.writable) {
+            res.write(data);
+        }
+        if (!res.finished) {
```
jaller94, line 29: Why .end when res is not finished?
Because we have to end our response?

src/workers/MetricsWorker.ts
```diff
@@ -47,9 +56,12 @@ function workerThread() {
             intervalCounter.set(time - lastDumpTs);
             lastDumpTs = time;
             const dump = msg.substr('metricsdump:'.length);
-            res.statusCode = 200;
-            res.write(`${dump}\n${registry.metrics()}`);
-            res.end();
+            if (res.finished) {
+                // Sometimes a message will come in far too late because we've already
+                // sent an empty response. Drop it here.
```
jaller94, line 61: Drop it here? This isn't doing anything.

src/workers/MetricsWorker.ts
Half-Shot: Yup. `write` doesn't end it, but also calling `end` on a finished response will err and crash the process.

src/workers/MetricsWorker.ts
Half-Shot: Apparently I was half asleep and missed a return

Issue #1103, opened 2020-08-11
Metrics for BridgedClient state


Half-Shot:
So we can see how many connections are broken, if any.

PR #1110, opened 2020-08-13
Connection state metrics

Half-Shot:
Fixed by #1103 

Often we find people are unable to get connected and it's hard to know exactly how big the problem is, let's solve that with moar data.

### DIFF
--- a/src/bridge/IrcBridge.ts
+++ b/src/bridge/IrcBridge.ts
@@ -356,12 +356,7 @@ export class IrcBridge {
         });
 
         metrics.addCollector(async () => {
-            clientStates.reset();
-            for (const [server, states] of Object.entries(this.clientPool.getConnectionStatesForAllServers())) {
-                for (const [state, value] of Object.entries(states)) {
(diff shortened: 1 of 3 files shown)


src/bridge/IrcBridge.ts
```diff
@@ -348,6 +354,15 @@ export class IrcBridge {
                 ircHandlerCalls.inc({method: kv[0]}, kv[1]);
             });
         });
+
+        metrics.addCollector(async () => {
+            clientStates.reset();
+            for (const [server, states] of Object.entries(this.clientPool.getConnectionStatesForAllServers())) {
+                for (const [state, value] of Object.entries(states)) {
```
benparsons, line 361: this is a lot of looping, four levels deep at this point to get the data in the right shape. anything that can be done here?

Half-Shot: Decided to save two loops and expose the gauge in the called function. Better?

Half-Shot:
I think this is better.

PR #1111, opened 2020-08-14
Publicity sync speedup

Half-Shot:
Currently we do a O(N) operation across all bridged channels if any of the modes change, which while safe was becoming computationally far too expensive to use.

This change makes it so that we only compute the publicity changes for the channel being changed, which should be safe enough.

### DIFF
--- a/src/bridge/IrcBridge.ts
+++ b/src/bridge/IrcBridge.ts
@@ -410,8 +410,6 @@ export class IrcBridge {
 
         await this.dataStore.removeConfigMappings();
 
-        this.clientPool = new ClientPool(this, this.dataStore);
-
         if (this.config.ircService.debugApi.enabled) {
             this.debugApi = new DebugApi(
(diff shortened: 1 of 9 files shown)


src/bridge/IrcBridge.ts
```diff
@@ -454,6 +452,8 @@ export class IrcBridge {
             this.ircServers.push(server);
         }
 
+        this.clientPool = new ClientPool(this, this.dataStore);
```
benparsons, line 455: I don't see why move this?

Half-Shot: it has a (somewhat hard to see) dependency on the configured server list (provided by `IrcBridge.getServers()`).

Half-Shot: Oh except I failed to mention that this PR has a dependency on #1091

PR #1113, opened 2020-08-15
Further improvements to public visibility syncing

Half-Shot:
This solves a few problems I encountered:
- rooms were being set to public
- onModeIs was not reliably arriving on startup, solved by a queue

### DIFF
--- a/src/bridge/PublicitySyncer.ts
+++ b/src/bridge/PublicitySyncer.ts
@@ -1,6 +1,7 @@
 import logger from "../logging";
 import { IrcBridge } from "./IrcBridge";
 import { IrcServer } from "../irc/IrcServer";
+import { Queue } from "../util/Queue";
 
 const log = logger("PublicitySyncer");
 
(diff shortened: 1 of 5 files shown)


src/bridge/PublicitySyncer.ts
```diff
@@ -23,47 +24,44 @@ export class PublicitySyncer {
...
     };
-    constructor (private ircBridge: IrcBridge) { }
-
+    private initModeQueue: Queue<{server: IrcServer; channel: string}>;
+    constructor (private ircBridge: IrcBridge) {
+        this.initModeQueue = new Queue(this.initModeForChannel.bind(this));
+    }
 
-    public async initModeForChannel(server: IrcServer, chan: string) {
+    public async initModeForChannel(opts: {server: IrcServer; channel: string}) {
```
jaller94, line 44: A semi-colon is valid here? Is it equivalent to a comma?
```suggestion
    public async initModeForChannel(opts: {server: IrcServer, channel: string}) {
```

Half-Shot: Yup. Apparently our linter favours it.

PR #1121, opened 2020-08-19
IndentGenerator fixes

Half-Shot:
This PR attempts to address a long standing performance issue with the username generation portion of the bridge. But first, some backstory.

Due to IRC limitations, we decided to limit usernames to be 10 characters long. The usernames need to be unique, so we use the user_id as a base. This works fine, but if you end up with two @matthews, it simply truncates to `matthew` and `matthew_1`, or in the case of `thehalfshot`: `thehalfs_1`. This works fine in most cases, but what happens if you try to bridge @_discord_12345 users into IRC. It works fine up until a point, where you may end up accumulating thousands of discord clients over a long enough period. The alogrithm is smart enoguh to start doing `thehalf_10` if you end up with enough halfshots.

The current code is dumb, and will simply generate a username, check if it's registered by querying the DB, and moving on. This means if you are discord user #7001, you will need to perform 7001 queries to the database. This is only a problem on first join, but Discord users often join the matrix.org bridge for the first time which then causes the DB to hammer.

### Solution Time

I've gone for the most obvious solution which is to query the database to determine if a username *prefix* is "full". That is, if `matthew_1` through til `matthew_9` is registered, then don't bother checking all those user_ids. And the same goes for 10-99, 100-999 and so on. This ends up saving us a lot of time, as we can quickly arrive at `disco_7001` with 4 queries to the DB.

### DIFF
--- a/src/irc/IdentGenerator.ts
+++ b/src/irc/IdentGenerator.ts
@@ -27,6 +27,10 @@ export class IdentGenerator {
     private static readonly MAX_REAL_NAME_LENGTH = 48;
     // The max length of <username> in USER commands
     private static readonly MAX_USER_NAME_LENGTH = 10;
+    // The delimiter ofthe username.
+    private static readonly USER_NAME_DELIMITER = "_";
+    // The delimiter ofthe username.
+    private static readonly MAX_USER_NAME_SUFFIX = 10000;
 
     private queue: Queue<{ matrixUser: MatrixUser; ircClientConfig: IrcClientConfig}>;
     constructor (private readonly dataStore: DataStore) {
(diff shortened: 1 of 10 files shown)


src/irc/IdentGenerator.ts
```diff
@@ -27,6 +27,10 @@ export class IdentGenerator {
     private static readonly MAX_REAL_NAME_LENGTH = 48;
     // The max length of <username> in USER commands
     private static readonly MAX_USER_NAME_LENGTH = 10;
+    // The delimiter ofthe username.
+    private static readonly USER_NAME_DELIMITER = "_";
+    // The delimiter ofthe username.
+    private static readonly MAX_USER_NAME_SUFFIX = 10000;
```
jaller94, line 33:
```suggestion
    // The delimiter of the username.
    private static readonly USER_NAME_DELIMITER = "_";
    // The delimiter of the username.
    private static readonly MAX_USER_NAME_SUFFIX = 10000;
```

src/irc/IdentGenerator.ts
```diff
@@ -208,6 +177,31 @@ export class IdentGenerator {
...
+     * Get the next available suffix for a given complete username.
+     * This function will find the correct suffix for a user in as fewest
+     * DB calls as possible.
+     * @param uname The IRC username
+     * @param domain The IRC domain.
+     */
+    private async getSuffixForUsername(username: string, domain: string) {
+        let suffixLength = 1;
+        while (suffixLength < IdentGenerator.MAX_USER_NAME_SUFFIX.toString().length) {
+            const prefix = username.substr(
```
jaller94, line 190: substr is deprecated and for your parameters they behave the same.
```suggestion
            const prefix = username.substring(
```

src/irc/IdentGenerator.ts
```diff
@@ -208,6 +177,31 @@ export class IdentGenerator {
...
+    /**
+     * Get the next available suffix for a given complete username.
+     * This function will find the correct suffix for a user in as fewest
+     * DB calls as possible.
+     * @param uname The IRC username
+     * @param domain The IRC domain.
+     */
+    private async getSuffixForUsername(username: string, domain: string) {
+        let suffixLength = 1;
+        while (suffixLength < IdentGenerator.MAX_USER_NAME_SUFFIX.toString().length) {
```
jaller94, line 189: Log10 tells you how many zeros there are after the 1. 😏
```suggestion
        while (suffixLength < Math.log10(IdentGenerator.MAX_USER_NAME_SUFFIX) + 1) {
```
giphub

Fell free to ignore this as yours is more readable.

src/irc/IdentGenerator.ts
jaller94, line 33: We talked a bit about this being weird to be set to 10000 instead of 9999.
If this was changed, the condition checking the length would need to become `<=` instead of `<`.

jaller94 (approved):

PR #1125, opened 2020-08-22
Fix matrix name change not propagating to irc

dalcde:
When we receive a matrix join event, we compare `bridgedClient.displayName` with the new display name to see if we need to update the irc client's nick. However, we didn't set `bridgedClient.displayName` to the new nick afterwards. This means if the matrix user changes their display name back to the original, the change is not propagated to irc.


Half-Shot:
I think this ended up being another PR I happened to see first. Sorry :(

PR #1128, opened 2020-08-26
WIP Support certfp

### LINKED ISSUE #747: Support client certificate authentication (certfp)
I thought there was an issue about this already, but I cannot find it. Client certificates would allow passwordless login to IRC server and generally IRC services allow multiple certificate fingerprints to be configured so users could upload (or have the bridge generate) unique SSL certificate which the users then `cert add` and if the private key got compromised, users would just remove that fingerprint from their IRC accounts and generate a new one.

Client certificates are also the only way to identify to OFTC automatically as far as I am aware.

I couldn't find technical details on how to implement it so I imagine it's the same certificate authentication that web browsers and everything else support and in some wider specification.

Instructions by IRC networks:

* [OFTC: Automatically Identifying Using SSL + CertFP](https://www.oftc.net/NickServ/CertFP/)
* [freenode: CertFP](https://freenode.net/kb/answer/certfp)

Half-Shot:
Fixes #747 

This PR adds support for authenticating via certificates, by storing certificates inside the br for each matrix user and then supplying them to the irc network on connect.

It's still a bit WIP in places. The private key isn't being encrypted yet, and I've not managed to get a successful authentication out of freenode.net yet, though I expect it's only a matter of time.

### DIFF
--- a/src/bridge/AdminRoomHandler.ts
+++ b/src/bridge/AdminRoomHandler.ts
@@ -47,6 +47,10 @@ const COMMANDS = {
         example: `!storepass [irc.example.net] passw0rd`,
         summary: `Store a NickServ password (server password)`,
     },
+    "!storecert": {
+        example: `!storecert [irc.example.net]`,
+        summary: `Store a certificate for authentication`,
+    },
     "!removepass": {
         example: `!removepass [irc.example.net]`,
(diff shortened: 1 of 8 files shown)


Half-Shot:
This has rotted a bit, we're instead going with #1331 first.

PR #1131, opened 2020-09-04
Update matrix-lastactive to fix an issue where sometimes reaps would fail

Half-Shot:
See the latest few changes on https://github.com/Half-Shot/matrix-lastactive for info.


jaller94 (approved):
TIL You maintain a lib for determining when a user was last active.

PR #1143, opened 2020-10-09
Do not leave after joining

Half-Shot:
We were allowing the .join call to fall through to the next statement. Ideally this should have been caught by tests.

### DIFF
--- a/changelog.d/1143.bugfix
+++ b/changelog.d/1143.bugfix
@@ -1,0 +1,1 @@
+Fix a bug where the bridge would leave a user after joining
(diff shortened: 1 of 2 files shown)


jaller94 (approved):
:D That's a good one.
Lets join... and now leave again.

Good one to have fixed.

jaller94:
https://github.com/matrix-org/matrix-appservice-irc/pull/1143/files#diff-b111e56781e7df3f23689602c86f556dR117-R118

These debug lines may not be accurate.
Maybe the leaving or kicking failed.

Half-Shot:
Nice spot! Have fixed.

PR #1146, opened 2020-10-12
Refactor nick => irc client mapping

Half-Shot:
In order to fix cases of double bridging.

Rather than storing two mappings of nickname => Client and user_id => Client, we will store one map for the user_id => Client and a LRU cache for the nickname. The intention being that if for whatever reason we have a hole in the nick mapping, the bridge can fall back to searching through the other mapping, at a speed penalty.

We also now consider dead clients ours, to avoid the case where we kill a client and then subsequently bridge across resulting parting information.

### DIFF
--- a/changelog.d/1146.bugfix
+++ b/changelog.d/1146.bugfix
@@ -1,0 +1,1 @@
+Fix more cases of double bridged users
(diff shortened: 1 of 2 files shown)


benparsons (approved):
lgtm

PR #1147, opened 2020-10-12
Don't wait for users to leave when quitting

### LINKED ISSUE #1141: User reap script should not await leaves
Because it stops us from processing other users and the script will timeout. Instead, we should quit all the users first and handle the leaves in the background.

Half-Shot:
Fixes #1141


benparsons (approved):
lgtm

PR #1150, opened 2020-10-13
Handle m.audio files

### LINKED ISSUE #1149: Bridge doesn't handle m.audio
`src/models/MatrixAction.ts` is able to handle `m.audio` just like other files, ie sending the MXC url to IRC, however `src/models/IrcAction.ts` `fromMatrixAction` does not have a way to handle these and thus drops it with an "Unknown action: audio"

looking at #429 this seems to be a regression

Half-Shot:
Fixes #1149

### DIFF
--- a/changelog.d/1150.bugfix
+++ b/changelog.d/1150.bugfix
@@ -1,0 +1,1 @@
+Fix bug where m.audio files would not be forwarded to IRC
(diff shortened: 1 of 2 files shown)


benparsons (approved):
might be neater to collapse image, video, audio and file, but this works

Half-Shot:
Annoyingly they all have slightly different action text, so it's not possible :(

PR #1152, opened 2020-10-13
Use EDU activity to determine user idleness

Half-Shot:
Whereas previously we've relied on users sending messages to prove if they are a lurker, this change would check to see if the users have sent any federated activity like presence, read receipts or typing. This allows the idle kicker to be much more fair about who it kicks.


src/bridge/IrcBridge.ts
```diff
@@ -803,6 +808,34 @@ export class IrcBridge {
...
+        const event = request.getData();
+        let userIds: string[]|undefined = undefined;
+        if (!this.activityTracker) {
+            return;
+        }
+        if (event.type === "m.presence" && event.content.presence === "online") {
+            userIds = [event.sender];
+        }
+        else if (event.type === "m.receipt") {
+            userIds = Object.keys(Object.values(event.content).map((v) => v["m.read"]));
```
benparsons, line 823: Is this taking ALL the `m.read`s? Surely we still only want the sender in this point? Or you refresh the whole list whenever there's a new `m.receipt`?

benparsons (commented):
question inline

src/bridge/IrcBridge.ts
Half-Shot: Synapse sends new receipts in the format https://matrix.org/docs/spec/client_server/r0.6.0#m-receipt, which makes it a bit tricky to determine what's changed. We *could* do some quick filtering to exclude *stale* receipts though.

Half-Shot: I've tried something else.

benparsons (approved):
lgtm

PR #1156, opened 2020-10-19
Ignore idle users on startup rather than connecting them

### LINKED ISSUE #1142: Remove idle users on startup before connecting them
The freenode bridge usually blows it's user cap on startup and we have to manually run the reap script in response to a restart. Ideally the bridge would remove idlers on startup

Half-Shot:
Fixes #1142 

matrix.org runs a kick script every day to remove idle users, but due to permission issues not every user will be kicked (the bridge lacks permissions to kick in some rooms). In order ensure we don't reconnect users who are idle, this config change will cause the bridge to ignore them on startup.

### DIFF
--- a/src/bridge/MemberListSyncer.ts
+++ b/src/bridge/MemberListSyncer.ts
@@ -250,8 +250,7 @@ export class MemberListSyncer {
         return this.syncableRoomsPromise;
     }
 
-    private joinMatrixUsersToChannels(rooms: RoomInfo[], injectJoinFn: InjectJoinFn) {
-        const d = promiseutil.defer();
+    private async joinMatrixUsersToChannels(rooms: RoomInfo[], injectJoinFn: InjectJoinFn) {
 
         // filter out rooms listed in the rules
         const filteredRooms: RoomInfo[] = [];
@@ -278,8 +277,17 @@ export class MemberListSyncer {
         // map the filtered rooms to a list of users to join
         // [Room:{reals:[uid,uid]}, ...] => [{uid,roomid}, ...]
         const entries: { roomId: string; displayName: string; userId: string; frontier: boolean}[] = [];
-        filteredRooms.forEach((roomInfo) => {
-            roomInfo.realJoinedUsers.forEach((uid, index) => {
+        const idleRegex = this.server.ignoreIdleUsersOnStartupExcludeRegex;
+        for (const roomInfo of filteredRooms) {
+            for (const uid of roomInfo.realJoinedUsers) {
+                if (this.server.ignoreIdleUsersOnStartup) {
+                    const idle = await this.ircBridge.activityTracker?.isUserOnline(
+                        uid, this.server.ignoreIdleUsersOnStartupAfterMs, false
+                    );
+                    if (!(idle?.online) && !idleRegex?.exec(uid)) {
+                        continue;
+                    }
+                }
                 entries.push({
                     roomId: roomInfo.id,
                     displayName: roomInfo.displayNames[uid],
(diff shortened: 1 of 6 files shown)


src/bridge/MemberListSyncer.ts
```diff
@@ -278,19 +277,28 @@ export class MemberListSyncer {
         // map the filtered rooms to a list of users to join
         // [Room:{reals:[uid,uid]}, ...] => [{uid,roomid}, ...]
         const entries: { roomId: string; displayName: string; userId: string; frontier: boolean}[] = [];
-        filteredRooms.forEach((roomInfo) => {
-            roomInfo.realJoinedUsers.forEach((uid, index) => {
+        const idleRegex = this.server.ignoreIdleUsersOnStartupExcludeRegex;
+        for (const roomInfo of filteredRooms) {
+            for (const uid of roomInfo.realJoinedUsers) {
+                if (this.server.ignoreIdleUsersOnStartup) {
+                    const idle = await this.ircBridge.activityTracker?.isUserOnline(
```
jaywink, line 284: What kind of effect will this have on the freenode bridge startup time?

src/bridge/MemberListSyncer.ts
```diff
@@ -278,19 +277,28 @@ export class MemberListSyncer {
...
+                        uid, this.server.ignoreIdleUsersOnStartupAfterMs, false
+                    );
+                    if (!(idle?.online) && !idleRegex?.exec(uid)) {
+                        continue;
```
jaywink, line 288: The effect will be some extra users on the matrix side which are not shown on irc side? If the user on matrix side becomes active (ie writes a message, for example), do things just work and the user is joined to the irc side?

src/bridge/MemberListSyncer.ts
```diff
@@ -278,19 +277,28 @@ export class MemberListSyncer {
...
+                }
                 entries.push({
                     roomId: roomInfo.id,
                     displayName: roomInfo.displayNames[uid],
                     userId: uid,
                     // Mark the first real matrix user f.e room so we can inject
                     // them first to get back up and running more quickly when there
                     // is no bot.
-                    frontier: (index === 0)
+                    frontier: !entries.find((ent) => ent.roomId === roomInfo.id),
```
jaywink, line 298: I don't understand this change in relation to the PR description or commits. Is this supposed to be here and it's related?

src/bridge/MemberListSyncer.ts
```diff
@@ -250,8 +250,7 @@ export class MemberListSyncer {
         return this.syncableRoomsPromise;
     }
 
-    private joinMatrixUsersToChannels(rooms: RoomInfo[], injectJoinFn: InjectJoinFn) {
-        const d = promiseutil.defer();
+    private async joinMatrixUsersToChannels(rooms: RoomInfo[], injectJoinFn: InjectJoinFn) {
```
jaywink, line 253: Just want to ensure that the addition of `async` here is taken into account as I see no changes to any calling code.

jaywink (approved):
Approved with comments - can be merged if those are all ok :+1:

src/bridge/MemberListSyncer.ts
Half-Shot: I'd imagine a ms or two per user, this function will short circuit if we have a last active time for users already set, which we will have for 99.9% of users. For users we have never seen before, it will do a lookup for whois/presence, or finally give up and say they are offline. For matrix.org, neither of those lookup options work so it will short circuit those and immediately return false.

Basically, not a lot of time spent per user.

src/bridge/MemberListSyncer.ts
Half-Shot: If the user sends a message then yeah they will be joined/connected. A tiny proportion of users who did an action while the bridge is down may be missed.

src/bridge/MemberListSyncer.ts
Half-Shot: Yep, it's important.

We previously assumed all joined members of a room would be joined to IRC so the first user would be considered the "frontier" user. However, since we *may* not bridge the first user, we need to do a slightly different check.

src/bridge/MemberListSyncer.ts
Half-Shot: This function used to return a deferred (a Promise), and the caller was already awaiting it. We're now using await inside the function, but otherwise the return value remains the same.

Issue #757, opened 2019-05-29
Matrix account randomly unregister from Freenode without nosiy warnings


erciccione:
I have a matrix account grouped with a registered freenode account.

After i identify as the Freenode user on Riot, everything is fine for some time, but randomly, the matrix user silently disconnects from Freenode and on reconnection NickServ notify me:

> This nickname is registered. Please choose a different nickname, or identify via /msg NickServ identify <password>.

The problem is that i often don't notice the message from NickServ and this cause the disconnection from my Freenode user and consequently the change of my nickname there. The change is not notified in any way, and i end up posting on Freenode convinced that my nickname is showing (since appear correct on matrix), but it actually shows as `Guest[random_number]` on Freenode.

I used the `!storepass` command in 'appservice-irc:matrix.org', but doesn't seem to solve to work at all. Right now i can only manually fix re-identifying to freenode when i realize the issue is happening.

I couldn't reproduce. The issue appear to happen randomly and in various timeframes (sometimes it disconnects after a month, sometimes twice in a week), doesn't matter if the matrix client is running or not.

Mikaela:
I think `!storepass` / `PASS` could be called as a hack and this issue would be properly resolved by https://github.com/matrix-org/matrix-appservice-irc/issues/747 (which has the added benefit of identifying you in the corner case that the bridge connects to a splitted server without services) and/or https://github.com/matrix-org/matrix-appservice-irc/issues/470 (which needs to be SASL 3.2 and supported by the IRC server to avoid the corner case).

I prefer #747, because it would also allow having a different certificate for the bridge and revoking it in case of server compromise instead of requiring changing passwords everywhere.

paul90:
Certainly seems very random, got kicked from one of a group of freenode channels when I started the Riot desktop client this morning. The other freenode channels seem to be unaffected, updating with the few overnight messages.

Could there be a timing issue?

erciccione:
Just a heads up: this stll "regularly" happens.

erciccione:
The problem is still happening but now got worse. I still get disconnected, but NickServ and @appservice-irc:matrix.org don't even let me know that i got disconnected anymore. My nick silently changes to `Guestxyz` and the only way to find it out if it's somebody from the chat let me know.

[...]

PR #1157, opened 2020-10-19
Check the displayname of a Matrix user before updating their nick

### LINKED ISSUE #1155: Check /profile before changing nick on m.room.member update
The bridge will change your nick if if sees a new m.room.member event for a user with a different displayname. Unfortunately lots of people use per-room displaynames for things that only make sense in that room. We should check that the displayname change is global before committing a new nick.

Half-Shot:
Fixes #1155

### DIFF
--- a/src/bridge/MatrixHandler.ts
+++ b/src/bridge/MatrixHandler.ts
@@ -547,14 +547,20 @@ export class MatrixHandler {
                     if (config && room.server.allowsNickChanges() &&
                         !config.getDesiredNick()
                     ) {
-                        try {
-                            const newNick = room.server.getNick(
-                                bridgedClient.userId, event.content.displayname
-                            );
-                            bridgedClient.changeNick(newNick, false);
-                        }
-                        catch (e) {
-                            req.log.warn(`Didn't change nick on the IRC side: ${e}`);
+                        const intent = this.ircBridge.getAppServiceBridge().getIntent();
+                        // Check that the /profile matches the displayname.
+                        const userProfile = await intent.getProfileInfo(event.state_key, "displayname");
+                        // We only want to update the nickname if the profile contains the displayname
+                        if (userProfile.displayname === event.content.displayname) {
+                            try {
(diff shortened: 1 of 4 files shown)


src/bridge/MatrixHandler.ts
```diff
@@ -547,14 +547,20 @@ export class MatrixHandler {
...
-                            const newNick = room.server.getNick(
-                                bridgedClient.userId, event.content.displayname
-                            );
-                            bridgedClient.changeNick(newNick, false);
-                        }
-                        catch (e) {
-                            req.log.warn(`Didn't change nick on the IRC side: ${e}`);
+                        const intent = this.ircBridge.getAppServiceBridge().getIntent();
+                        // Check that the /profile matches the displayname.
+                        const userProfile = await intent.getProfileInfo(event.state_key, "displayname");
```
jaywink, line 552: My understanding of the bridges is still poor, but looks like this code is only handled "on join"? So wondering shouldn't it be whenever the matrix side changes display name? And also if this is code that runs for everyone during a startup, will the profile details gets fetched for everyone or just changed names since previous restart?

jaywink (approved):
One comment due to poor understanding of bridge code.

src/bridge/MatrixHandler.ts
Half-Shot: Ah, changing displaynames manifests as a resend of the m.room.member[content.membership="join"] event, which is handled by this function every time. onJoin is a slight misnomer

Half-Shot: >And also if this is code that runs for everyone during a startup, will the profile details gets fetched for everyone or just changed names since previous restart?

We shouldn't see a join for everyone during startup, so I don't think this is a huge concern.

erciccione:
Curious to see how this will relate to #757, since the IRC nick is changed to GuestXYZ when the user gets unidentified (for reasons unknown for now)

PR #1159, opened 2020-10-20
Propagate a reason all the way through killing the bridge

xnrand:
This makes quit messages less confusing


Half-Shot (approved):
Looking good! A easy change that makes a big difference :)

PR #1160, opened 2020-10-21
Ping the bridge via the homeserver on startup to ensure the AS transport works

Half-Shot:
This PR will cause the bridge to ping the homeserver on startup, by sending an event to a room, and then checking it can read it back across the transaction stream.


benparsons (approved):
lgtm

PR #1161, opened 2020-10-21
Drop the old membership queue implementation and use the matrix-appservice-bridge one

Half-Shot:
Depends on #1160 

Depends on the next matrix-appservice-bridge release

This change also sneakily adds support for part reasons from IRC.


changelog.d/1161.misc
```diff
@@ -0,0 +1 @@
+Bridge IRC `reason`s when users QUIT or PART
```
Half-Shot, line 1: Due to accidental dependency on #1160. Diffs should look more sensible once that one's been merged.

src/bridge/IrcHandler.ts
```diff
@@ -793,20 +793,28 @@ export class IrcHandler {
...
-        const promise = await Promise.all(matrixRooms.map(async (room) => {
-            await this.membershipQueue.leave(
-                room.getId(), userId, req, true, "Client PARTed from channel",
-                leavingUser.isVirtual ? this.ircBridge.appServiceUserId : undefined);
-            try {
-                await this.roomAccessSyncer.removePowerLevels(room.getId(), [userId]);
+        req.log.info("Mapped nick %s to %s (leaving %s room(s))", nick, userId, matrixRooms.length);
+        await Promise.all(matrixRooms.map(async (room) => {
+            if (leavingUser.isVirtual) {
+                return this.membershipQueue.leave(
```
benparsons, line 799: A quick read of https://github.com/matrix-org/matrix-appservice-bridge/blob/develop/src/components/membership-queue.ts didn't enlighten me - how are errors handled here? This `Promise.all` doesn't seem to be stored anywhere, we just await and ignore the return value?

src/bridge/MemberListSyncer.ts
```diff
@@ -358,28 +348,17 @@ export class MemberListSyncer {
...
             log.debug(`Leaving ${userId} from ${item.roomId}`);
-            // Do this here, we might not manage to leave but we won't retry.
             this.usersToLeave--;
-            await this.ircBridge.getAppServiceBridge().getIntent(userId).leave(item.roomId);
-        });
-
-        await Promise.all(item.userIds.map((userId) =>
-            q.enqueue(userId, userId)
-        ));
+            return this.memberQueue.leave(item.roomId, userId, staticRequest, false);
```
benparsons, line 357: Same q as above

benparsons (commented):
Some questions

src/bridge/IrcHandler.ts
Half-Shot: membershipQueue.(join|leave) will return a promise that will eventually resolve or reject depending on if we successfully send presence (after a set number of retries). If any of these fail, the promise.all will fail, and that will be caught in the `complete` function in `IrcEventBroker`, which in turn logs an error and that's it.

benparsons: great summary thanks

benparsons (approved):
lgtm

PR #1162, opened 2020-10-21
Remove hacks around reconnect interval

Half-Shot:
We added some hacks to the bridge a while ago, and this is simply me cleaning them up with some slightly nicer code.


benparsons (approved):
lgtm

PR #1165, opened 2020-10-23
Check if a user is in the room before doing a kick on QUIT

### LINKED ISSUE #1163: Do not kick users who aren't in the room
Doing a .kick() will NOT check if the user is actually in the room first, which can lead to lots of spammy leave -> leave operations. It's not terribly expensive to do a state lookup first, so let's do that.

Half-Shot:
Fixes #1163 

This change does a state lookup before kicking a user to determine if they should be kicked, as synapse will accept a (null) -> leave.

### DIFF
--- a/changelog.d/1165.bugfix
+++ b/changelog.d/1165.bugfix
@@ -1,0 +1,1 @@
+Fixed an issue where the bridge would kick users from rooms they never joined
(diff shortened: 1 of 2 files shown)


jaywink (approved):
Looks good :+1:

PR #1169, opened 2020-10-27
Rename /killPortal to /killRoom and allow it to delete any mapping

### LINKED ISSUE #1041: killPortal doesn't work when origin=join
Hi,

I was using killPortal on a number of rooms, and had one that was particularly stubborn.  It kept saying "room not found".  Finally, I edited my rooms.db while the server was down and changed `"origin":"join"` to `"origin":"alias"` and then it worked.

The problem appears to be this in src/DebugApi.ts:

```
        // Find room
        const room = await store.getRoom(
            roomId,
            domain,
            channel,
            "alias"
        );
        if (room === null) {
            result.error.push("Room not found");
            this.wrapJsonResponse(result, false, response);
            return;
        }
```

where `"alias"` is hard-coded.

Half-Shot:
Which would fix #1041

### DIFF
--- a/src/DebugApi.ts
+++ b/src/DebugApi.ts
@@ -80,8 +80,8 @@ export class DebugApi {
             this.onReapUsers(query, response);
             return;
         }
-        else if (req.method === "POST" && path === "/killPortal") {
-            this.killPortal(req, response);
+        else if (req.method === "POST" && path === "/killRoom") {
+            this.killRoom(req, response);
             return;
         }
         else if (req.method === "GET" && path === "/inspectUsers") {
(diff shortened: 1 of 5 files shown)


src/DebugApi.ts
```diff
@@ -281,7 +281,7 @@ export class DebugApi {
         });
     }
 
-    private async killPortal (req: IncomingMessage, response: ServerResponse) {
+    private async killRoom (req: IncomingMessage, response: ServerResponse) {
```
jaywink, line 284: This looks very backwards incompatible. Was the `killPortal` undocumented and thus can just be replaced? Or are we fine because it's in the debug API?

jaywink (approved):
One note to consider

src/DebugApi.ts
Half-Shot: The latter. It's called out in the changelog, and we don't make any guarantees that the debugAPI won't change.

PR #1179, opened 2020-11-03
Nicer format for Matrix quotes in IRC

Midar:
This addresses one of the most common complaints about using Matrix
quotes in IRC channels: That IRC users consider the format of quotes
"weird".

This commit changes the current format from

    <user "Original message"> Reply

to

    "<user> Original message" <- Reply

which is much more aligned with how IRC users manually quote a message.

### DIFF
--- a/changelog.d/1179.misc
+++ b/changelog.d/1179.misc
@@ -1,0 +1,1 @@
+Improve the format for Matrix quotes on IRC
--- a/spec/integ/matrix-to-irc.spec.js
+++ b/spec/integ/matrix-to-irc.spec.js
@@ -263,7 +263,7 @@ describe("Matrix-to-IRC message bridging", function() {
                 expect(client.nick).toEqual(testUser.nick);
                 expect(client.addr).toEqual(roomMapping.server);
                 expect(channel).toEqual(roomMapping.channel);
-                expect(text).toEqual(`<${repliesUser.nick} "This is the real message"> Reply Text`);
+                expect(text).toEqual(`"<${repliesUser.nick}> This is the real message" <- Reply Text`);
             }
         );
         const formatted_body = constructHTMLReply(
@@ -308,7 +308,7 @@ describe("Matrix-to-IRC message bridging", function() {
                 expect(client.addr).toEqual(roomMapping.server);
                 expect(channel).toEqual(roomMapping.channel);
                 // We use the nick over the displayname
-                expect(text).toEqual(`<M-friend "This is the real message"> Reply Text`);
+                expect(text).toEqual(`"<M-friend> This is the real message" <- Reply Text`);
             }
         );
         const formatted_body = constructHTMLReply(
@@ -352,7 +352,7 @@ describe("Matrix-to-IRC message bridging", function() {
                 expect(client.nick).toEqual(testUser.nick);
                 expect(client.addr).toEqual(roomMapping.server);
                 expect(channel).toEqual(roomMapping.channel);
-                expect(text).toEqual(`<${repliesUser.nick} "This"> Reply Text`);
+                expect(text).toEqual(`"<${repliesUser.nick}> This" <- Reply Text`);
             }
         );
         const formatted_body = constructHTMLReply(
@@ -460,7 +460,7 @@ describe("Matrix-to-IRC message bridging", function() {
                 expect(client.nick).toEqual(testUser.nick);
                 expect(client.addr).toEqual(roomMapping.server);
                 expect(channel).toEqual(roomMapping.channel);
-                expect(text).toEqual('<M-friend "Message #2"> Message #3');
+                expect(text).toEqual('"<M-friend> Message #2" <- Message #3');
             }
         );
 
@@ -507,7 +507,7 @@ describe("Matrix-to-IRC message bridging", function() {
                 expect(client.nick).toEqual(testUser.nick);
                 expect(client.addr).toEqual(roomMapping.server);
                 expect(channel).toEqual(roomMapping.channel);
-                expect(text).toEqual('<WibbleWob "This is the real message"> Reply Text');
+                expect(text).toEqual('"<WibbleWob> This is the real message" <- Reply Text');
             }
         );
         const formatted_body = constructHTMLReply(
--- a/src/bridge/MatrixHandler.ts
+++ b/src/bridge/MatrixHandler.ts
@@ -1156,8 +1156,6 @@ export class MatrixHandler {
             if (lines[0].length > REPLY_SOURCE_MAX_LENGTH) {
                 rplSource = rplSource + "...";
             }
-            // Wrap in formatting
-            rplSource = ` "${rplSource}"`;
         }
         else {
             // Don't show a source because we couldn't format one.
@@ -1179,7 +1177,7 @@ export class MatrixHandler {
         }
 
         return {
-            formatted: `<${rplName}${rplSource}> ${rplText}`,
+            formatted: `"<${rplName}> ${rplSource}" <- ${rplText}`,
             reply: rplText,
         };
     }


ptman:
While this might be an improvement, I've mostly been hearing complaints about *how long* quotes are with replies. Most irc users seem to be of the opionion that you should reply only with the nick of the person you're replying to, instead of quoting the message.

Half-Shot:
>While this might be an improvement, I've mostly been hearing complaints about how long quotes are with replies. Most irc users seem to be of the opionion that you should reply only with the nick of the person you're replying to, instead of quoting the message.

I'm in agreement with this one tbh. Could we try it with just the nick: in the form `Nick: the reply text`?

zouppen:
I'd go for `Nick: the reply text` as @Half-Shot suggested. It's the way IRC users have done it for decades. It loses the context, but in most cases the replies are done quickly after the original message so it's easy to infer it. So in my opinion IRC users prefer compactness over of maximal clarity.

If you want to keep context and do it time zone independently would be to a format such as `Nick: (46 mins later) the reply text`. Doing it perfectly might include a condition if the message is being replied before the original poster has written any more messages (the context is clear anyway).

But I'd prefer implementing Half-shot's suggestion first and make it nicer later if people complain about the loss of context.

Midar:
> I'm in agreement with this one tbh. Could we try it with just the nick: in the form Nick: the reply text?

This won't work.

I've talked about this with a group of die-hard IRC users (no way to switch to something else, everybody else should just use IRC anyway, and if you have the audacity to not use IRC at least behave like an IRC user), and the consensus was:

* For far away quotes, the format I proposed is fine.
* For near quotes, mentioning the user is preferred.

This makes sense, but is nothing we can enforce via software. So, what we can do from a technical perspective is offer the far quote format that is acceptable. Not using quotes at all when the message is close is "user education", nothing we can do technically. And I agree that it's even annoying on Matrix if someone quotes the message JUST above their message. Quotes should be used to add context when it has been lost, not to reply to each and every message.

Therefore, I think this PR is doing as much as can be done: Fix the format for far quotes to one that is less confusing. The main complaint about the current format is that it was not even immediately clear that this is a quote. The new format mimics how IRC users quote manually: By copying the line from their IRC client.

Half-Shot:
I think you might be giving up too easy. We *can* do near quotes by either by checking if the message being replied to has follow up messages from the same user. This can be done trivially, as the bridge is already bridging messages n the room, so could easily and cheaply cache this information. And as the others suggest, this would make the experience so much better.

thibaultamartin:
Hello! Chiming in and adding the workarounds we’re discussing with IRC users from the GNOME community:

There seem to be several ways to make the experience better

## The minimalist way

When somebody on Matrix makes a reply, completely discard the "quote" part of the reply. That is, if "alex" replies "oh me too!" to "fred" who said "I love pizza!", from IRC the message would look like:

`<alex>`: `fred: oh me too!`

## The URL way

When somebody on Matrix makes a reply, URLize the "quote" part of the reply. That is, if "alex" replies "oh me too!" to "fred" who said "I love pizza!", from IRC the message would look like:

`<alex>`: `fred: oh me too! (Re: https://pastebin.tld/aoifhqs1d345)`

## The URL + beginning of the message

When somebody on Matrix makes a reply, URLize the "quote" part of the reply but display the first chars. That is, if "alex" replies "oh me too!" to "fred" who said "I love pizza but I don’t like it when they put too much cheese on it", from IRC the message would look like:

`<alex>`: `fred: oh me too! (Re: I love pizza but... https://pastebin.tld/aoifhqs1d345)`

toonn:
As an IRC user, I'd like to request dropping the quote entirely. The "context reestablishing" doesn't work. People would've had to have seen the message being replied to and remember a very short truncated prefix of it to recall it. Most people will only recall the message if it was so recent the quote is entirely redundant.

The added burden of parsing these quotes as being quotes and therefore irrelevant is high, many IRC clients don't differentiate any more between messages than they have to so it's basically a solid wall of text, a terminal IRC client in a tmux session on a server is a really popular way of circumventing IRC's lack of channel history for example. Anything between the user's nick and their message makes the association less clear and anything trailing after their message makes it less clear where the next message starts.

For those rare occasions when the reference is not clear, simply asking for clarification can often bring everybody up to speed quickly, while many may not be motivated enough to go digging through the backlog.

A simple mention is all we need/want (no @'s cause those often trigger bots : ) )

Midar:
I have to *srtrongly* disagree with that sentiment: The current Discord bridge doesn't do replies at all and entirely drops the context, and the confusion to which it has come already is so, *so* bad that it's almost better turning down any bridged room. As long as someone can quote a message and expects others to see the same quote, it doesn't work - you can't just silently drop context. It very quickly then comes to the point that there's so much confusion that no briddging is better. This has been proven in the real world to not work already now.

toonn:
Discord is not IRC.  I think you're making a mistake considering the two even similar. The current situation means bridged matrix users get asked not to use the feature at all, which is probably annoying both in that they need to remember precisely which rooms to refrain from using the feature in *and* because every time they slip up they get another couple requests to please stop.

Bridges should respect the implicit customs of the networks they join. You keep clinging to this idea that the truncated quote provides "context" but that's simply not true. To provide enough context to actually be useful it would have to pretty much repeat the message in its entirety (*please* don't do this, it would be *even* worse), of course making it even less readable in the process.

Midar:
It doesn't matter if it's IRC, Discord or Matrix - as soon as there is a bridge and there is a reply feature, it *will* be used. And we've already seen the confusion it causes if context is just silently dropped. Of all the options, silently dropping context is by, far, *far, **far** the worst of all possible options - to the point I would kill the IRC bridge from my rooms if it were actually going that way, because the amount of confusion will just be so much that it kills all discussion.

I disagree that the context is useless - the quote is oftentimes complete, and it is only truncated if it does not fit into a single line. This limit should probably be increased a litlle. It also allows scrolling up to the message it was about and reading up the surrounding messages.

All of this is gone with just silently dropping quotes. I'd be ok with a way to say "No quotes in this room", but silently dropping it is so bad that it would make the entire IRC bridge useless.

toonn:
It does matter whether it's IRC, custom is that the mention is all the context you need. I doubt there would be much confusion past "Oh, you didn't get the message this was replying to?" Certainly not a complete breakdown of communication. I'm fine with the dropping of context not being silent but I imagine a notification every time you use the reply-to feature in a bridged room gets old really quick.

Midar:
Ok, let me give you an example why this is a *terrible* idea, by showing you the same conversation in 4 diferent formats (current, my proposal, your two proposals):

Version 1 (current version):
```
<a> https://github.com/matrix-org/matrix-appservice-irc/pull/1179
<a> Someone proposed a better quoting format for IRC
<a> Someone else proposed to drop quotes entirely
<b> Hmm, that sounds like it might create a lot of confusion.
<a> Yeah, toonn's idea is pretty terrible
<c> <a "Someone else proposed to drop quotes entirely"> I agree
```

Version 2 (this PR):
```
<a> https://github.com/matrix-org/matrix-appservice-irc/pull/1179
<a> Someone proposed a better quoting format for IRC
<a> Someone else proposed to drop quotes entirely
<b> Hmm, that sounds like it might create a lot of confusion.
<a> Yeah, toonn's idea is pretty terrible
<c> > <a> Someone else proposed to drop quotes entirely <- I agree
```

Version 3 (drop entirely):
```
<a> https://github.com/matrix-org/matrix-appservice-irc/pull/1179
<a> Someone proposed a better quoting format for IRC
<a> Someone else proposed to drop quotes entirely
<b> Hmm, that sounds like it might create a lot of confusion.
<a> Yeah, toonn's idea is pretty terrible
<c> I agree
```

Version 4 (only mention):
```
<a> https://github.com/matrix-org/matrix-appservice-irc/pull/1179
<a> Someone proposed a better quoting format for IRC
<a> Someone else proposed to drop quotes entirely
<b> Hmm, that sounds like it might create a lot of confusion.
<a> Yeah, toonn's idea is pretty terrible
<c> I agree
```

You notice why your idea is terrible? I turned around c's statement by 180°, making c say the opposite on IRC than on Matrix. You are now at the point where the bridge is twisting the meaning to say the opposite. This is no worse than having no bridge whatsoever.

toonn:
Version 4 is the only one I suggested. And I still prefer it. Afaihs (on IRC channels) that's not how people actually use the reply-to feature. They usually simply always use it. This is the case the UX needs to be optimised for, not the rare exception where it'll cause slight confusion until "c" clarifies their position.

Midar:
Except the clarification will never happen, because there is nothing giving away that the bridge just changed the meaning. You cannot fix a social probem (people overusing quotes) with technical means: It just doesn't work, you'll only make it worse. You will not fix it by intentionally breaking it.

If I'm a regular user, I would not want to have the bridge silently twist my words to mean the opposite. If I were a user who would know nothing about this and the bridge would have just relayed the opposite of what I said and I would only have found out by pure chance, I would be very upset - and rightfully so. And would probably stop using the IRC bridge or Matrix altogether. Something like that would immediately disqualify any chat product for me and I would never ever touch it again with a ten foot pole.

How would you feel if the bridge silently rewrites it when you write "yes" on IRC to be "no" on Matrix? Would you like that? Because you are essentially arguing that this wouldn't be so bad, because you could clarify if it someday gets noticed by pure luck.

Midar:
I'm closing my own pull request and deleting it, since this went into a very wrong direction very quickly. I want to avoid making the bridge worse rather than better, and it would be sad if I have to stop using the IRC bridge just because this PR was giving people the idea to entirely break the bridge. Lesson learned: Never create a PR trying to improve things, you'll only give people the idea to make it much worse.

ptman:
“Meanwhile, the poor Babel fish, by effectively removing all barriers to communication between different races and cultures, has caused more and bloodier wars than anything else in the history of creation.”

toonn:
I don't understand why a couple of my comments necessitate completely abandoning the work. Any improvement would be welcome. As an IRC user I've just observed numerous complaints about this and when it came up yet again yesterday someone pointed out nothing would be improved unless people voiced their opinion. I had remained quiet until now because I figured the bridge maintainers must already be away of what a painpoint this is for IRC users and I wanted to avoid drama.

Re misunderstandings, it's hardly as if most interactions via chat are binary along the lines of yes/no, it really seems like an exaggeration to say that reducing reply-to to a mention would cause more than a limited amount of confusion, once this has happened once the matrix user would be somewhat aware of avoiding it in the future. Which, IMO, would still be better than having to retrain behavior so you don't use reply-to in rooms bridged to IRC.

If the maintainers prefer OP's suggestions, I'm sure my comments wouldn't keep them from merging the PR as-is. Though @thibaultamartin's suggestion to make the quote a suffix rather than a prefix reads a lot clearer to me. It was really not my intent to have this PR closed without review.

I'm sorry if my opinions aren't welcome here, I'll refrain from sharing them.

Midar:
I'd rather maintain the status quo than making it even more confusing, hence I closed the PR since this was moving into the opposite direction of what I wanted to achieve.

PR #1186, opened 2020-11-12
Cache MODE -> power level changes before applying

### LINKED ISSUE #1180: 'Voice only' Freenode channels are spammed by a crazy amount of "changed the power level to" messages
A channel on Freenode was recently object of some spam. To temporarely fix, the OPs made the channel voice only. Beside breaking the bridge and being not allowed to write anything, the channels are full of spammy messages. See the gif:

spamfest

This happens to all three channels currently restricted. As you can see, this problem makes the chat almost unreadable.

Half-Shot:
Fixes #1180 

If we see a lot of mode changes at once, the bridge will attempt to send one power level event for each. This PR will instead cache the power level changes for a short while in memory, and will apply them once the requests stop coming in. This is configurable in the config file, but defaults to 1 second.


jaller94 (approved):
Code-wise I don't see anything wrong with this.

Issue #1176, opened 2020-11-03
Cannot route PM to $$* (null@chat.freenode.net)


Half-Shot:
Many many log lines for many users seeing PMs for $$*, to the point where the process becomes saturated.

PR #1189, opened 2020-11-13
Do an early check to see if the to nickname is valid before trying to handle a message

Half-Shot:
Partial fix to #1176 

Ideally we would work out why we are getting these messages too, but at the moment we can at least reject invalid nicks early on to avoid doing more processing than needed.

### DIFF
--- a/src/irc/IrcEventBroker.ts
+++ b/src/irc/IrcEventBroker.ts
@@ -284,7 +284,6 @@ export class IrcEventBroker {
             const req = createRequest();
             // Check and drop here, because we want to avoid the performance impact.
             if (!IrcEventBroker.isValidNick(to)) {
-                req.log.warn(`Dropped message due to invalid nick. to: ${to}`);
                 req.resolve(BridgeRequestErr.ERR_DROPPED);
                 return;
             }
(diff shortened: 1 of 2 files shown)


src/irc/IrcEventBroker.ts
```diff
@@ -564,4 +582,9 @@ export class IrcEventBroker {
             await req();
         })();
     }
+
+    static isValidNick(nick: string) {
+        // The first character must be one of these.
+        return /^[A-Za-z\[\]\\`_^\{\|\}]/.test(nick[0])
```
jaller94, line 588: ((we don't enforce lines to end on a semicolon?))

src/irc/IrcEventBroker.ts
```diff
@@ -282,6 +282,12 @@ export class IrcEventBroker {
         connInst.addListener("message", (from: string, to: string, text: string) => {
             if (to.startsWith("#")) { return; }
             const req = createRequest();
+            // Check and drop here, because we want to avoid the performance impact.
+            if (!IrcEventBroker.isValidNick(to)) {
+                req.log.warn(`Dropped message due to invalid nick. to: ${to}`);
```
jaller94, line 287: This makes me wonder if it is safe to output user-input in the logs.

jaller94 (approved):
Loos good to me.

src/irc/IrcEventBroker.ts
Half-Shot: strange. will fix.

Issue #1191, opened 2020-11-19
Slow function hitlist


Half-Shot:
- getNickUserIdMappingForChannel (around `2s` of time spent in the function over a 49s period!)
- ~getUserFeatures (`144ms`, could easily be cached)~
- formatMentions(`354ms`. Regex's were cheapish, but the function itself is not)
- ~`BridgedClient.chanList` does a Set -> Array conversion each time it's called, which is fairly often.~
- ~The console.logs from the js-sdk take up around 2s of time in total. We don't really need those.~

Also 4s spent in total handling unhandledPromises within sentry. We could save a lot of time here by clearing those up.

Half-Shot:
We're not actively trying to fix these atm as we have ample bandwidth for now. Going to close.

PR #1192, opened 2020-11-20
Small performance fixes

Half-Shot:
Partial fixes for #1191 

- Stop converting the channel list from a Set to an Array which costs some time
- Cache requests to get "user features" from the DB.


src/irc/ClientPool.ts
```diff
@@ -496,7 +496,7 @@ export class ClientPool {
     public getNickUserIdMappingForChannel(server: IrcServer, channel: string): {[nick: string]: string} {
         const nickUserIdMap: {[nick: string]: string} = {};
         for (const [userId, client] of this.virtualClients[server.domain].userIds.entries()) {
-            if (client.chanList.includes(channel)) {
+            if (client.inChannel(channel)) {
```
jaller94, line 499: I'm a bit surprised that this became `.inChannel` instead of `client.chanList.has(channel)`

jaller94 (approved):
Looks good to me.

src/irc/ClientPool.ts
Half-Shot: We should have been using inChannel from the start really :/

PR #1196, opened 2020-11-24
Batch together PL changes on leave

Half-Shot:
Also ensure we update the ircHandler config when we cause a reload.

### DIFF
--- a/src/bridge/RoomAccessSyncer.ts
+++ b/src/bridge/RoomAccessSyncer.ts
@@ -88,7 +88,7 @@ export class RoomAccessSyncer {
     }
 
     public async removePowerLevels(roomId: string, users: string[], req: BridgeRequest) {
-        return Promise.all(users.map(userId =>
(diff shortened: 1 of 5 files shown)


src/bridge/RoomAccessSyncer.ts
```diff
@@ -88,9 +88,9 @@ export class RoomAccessSyncer {
     }
 
     public async removePowerLevels(roomId: string, users: string[], req: BridgeRequest) {
-        const map: {[userId: string]: null} = {};
-        users.forEach((u) => {map[u] = null});
-        return this.changePowerLevels(roomId, map, req);
+        return Promise.all(users.map(userId =>
```
jaller94, line 91: Avoids unexpected return types (array full of `void`/`undefined`).
```suggestion
        await Promise.all(users.map(userId =>
```

Also easier to read, because I won't be looking for a possible return type.

jaller94 (approved):
Looks good.

PR #1204, opened 2020-12-08
Warn when the matrix user is unable to talk in a channel

### LINKED ISSUE #1203: We should warn on 404 Cannot send to nick/channel
Rather than silently failing. Perhaps send a message to the users' admin room as we do for other errors.

Half-Shot:
This will send an error inside the user's admin room, like we do for other failures.

Fixes #1203

### DIFF
--- a/changelog.d/1204.feature
+++ b/changelog.d/1204.feature
@@ -1,0 +1,1 @@
+Warn Matrix users if they are unable to speak in a channel.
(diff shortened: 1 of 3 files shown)


jaller94 (approved):
Looks good.

Emmagodnuel:
hs/warn-on-muted

PR #1206, opened 2020-12-11
Create DebugAPI after its dependency ClientPool

### LINKED ISSUE #1199: Error in killUser debug endpoint: Cannot read property 'getBridgedClientsForUserId' of undefined
**Describe the bug**
The `killUser` debug endpoint gives me the following response:
```
TypeError: Cannot read property 'getBridgedClientsForUserId' of undefined
```

The logs show:

```
2020-12-02 12:05:41 ERROR:DebugApi TypeError: Cannot read property 'getBridgedClientsForUserId' of undefined
    at DebugApi.killUser (/usr/local/share/.config/yarn/global/node_modules/matrix-appservice-irc/lib/DebugApi.js:220:35)
    at IncomingMessage.<anonymous> (/usr/local/share/.config/yarn/global/node_modules/matrix-appservice-irc/lib/DebugApi.js:157:36)
    at IncomingMessage.emit (events.js:314:20)
    at IncomingMessage.EventEmitter.emit (domain.js:483:12)
    at endReadableNT (_stream_readable.js:1241:12)
    at processTicksAndRejections (internal/process/task_queues.js:84:21)
```

**To Reproduce**
Steps to reproduce the behavior (with HTTPie):

```
$ http :11101/killUser access_token==$token user_id=<redacted> reason=deprecation deactivate:=false
HTTP/1.1 500 Internal Server Error
Connection: keep-alive
Content-Type: text/plain
Date: Wed, 02 Dec 2020 11:05:33 GMT
Keep-Alive: timeout=5
Transfer-Encoding: chunked

TypeError: Cannot read property 'getBridgedClientsForUserId' of undefined
```

**Expected behavior**
A successfull response.

**Server:**
 - OS: Debian 10.6
- NodeJS version: v12.20.0
 - Bridge Version: 0.23.0

jaller94:
Fixes #1199.

The issue is that DebugApi gets created before clientPool gets created. The parameter `this.clientPool` is undefined at this point.
`this.clientPool` is updated later but that does not get passed to DebugApi.

I tested this for a while but I'm not 100% confident that delaying the creation of DebugApi doesn't have some drawbacks.
Steps that are now run before its creation:
1. Restoring last active times from DB
2. maintain a list of IRC servers in-use (**this modifies config.ircServers which is another parameter of DebugApi**)
2.1. store the config mappings in the DB to keep everything in one place.
3. clientPool gets created

### DIFF
--- a/changelog.d/1206.bugfix
+++ b/changelog.d/1206.bugfix
@@ -1,0 +1,1 @@
+Create DebugAPI after its dependency ClientPool
--- a/changelog.d/1206.bugfix
+++ b/changelog.d/1206.bugfix
@@ -1,1 +1,1 @@
-Create DebugAPI after its dependency ClientPool
+Fix potential error when using killUser debug endpoind
--- a/changelog.d/1206.bugfix
+++ b/changelog.d/1206.bugfix
@@ -1,1 +1,1 @@
-Fix potential error when using killUser debug endpoind
+Fix potential error when using killUser debug endpoint
--- a/src/bridge/IrcBridge.ts
+++ b/src/bridge/IrcBridge.ts
@@ -530,17 +530,6 @@ export class IrcBridge {
 
         await this.dataStore.removeConfigMappings();
 
-        if (this.config.ircService.debugApi.enabled) {
-            this.debugApi = new DebugApi(
-                this,
-                this.config.ircService.debugApi.port,
-                this.ircServers,
-                this.clientPool,
-                this.registration.getAppServiceToken() as string
-            );
-            this.debugApi.run();
-        }
-
         if (this.activityTracker) {
             log.info("Restoring last active times from DB");
             const users = await this.dataStore.getLastSeenTimeForUsers();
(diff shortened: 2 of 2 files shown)


changelog.d/1206.bugfix
```diff
@@ -0,0 +1 @@
+Create DebugAPI after its dependency ClientPool
```
Half-Shot, line 1: We should mention what bug this fixes, rather than the solution as it's a user facing changelog.

Half-Shot (approved):
Otherwise good :+1:

changelog.d/1206.bugfix
jaller94: Does my new message better fit the purpose of the changelog?

Half-Shot (approved):
Seems reasonable. It's a shame Typescript didn't catch this.

changelog.d/1206.bugfix
```diff
@@ -1 +1 @@
-Create DebugAPI after its dependency ClientPool
\ No newline at end of file
+Fix potential error when using killUser debug endpoind
```
Half-Shot, line 1: typo

jaller94:
@Half-Shot This is not the fault of TypeScript. The compiler is instructed to ignore `clientPool` (which gets defined in `.run()` and invalidated in `.kill()`.

https://github.com/matrix-org/matrix-appservice-irc/blob/1c8ec4cc39d758b7fb5e34f8b9b2c6e6f18d8961/src/bridge/IrcBridge.ts#L65

This partly also applies to this property:
https://github.com/matrix-org/matrix-appservice-irc/blob/1c8ec4cc39d758b7fb5e34f8b9b2c6e6f18d8961/src/bridge/IrcBridge.ts#L81

PR #1212, opened 2021-01-06
Copy package.json into Docker

### LINKED ISSUE #1205: bridgeversion command reports "unknown" in docker container
**Describe the bug**
When running the matrixdotorg/matrix-appservice-irc docker container, the `!bridgeversion` command reports `BridgeVersion: unknown`.

I see that it should be getting the version from `package.json`.  I'm not sure exactly where that file is supposed to exist in the directory hierarchy, but it doesn't appear to be anywhere under `/app`:

```
# docker exec -ti <container id> /bin/bash
root@6965c452e95f:/app# find . -name package.json -print0 | xargs -0 grep matrix-appservice-irc
root@6965c452e95f:/app# 
```

**To Reproduce**
1. Set up docker container as described in https://github.com/matrix-org/matrix-appservice-irc/blob/develop/Dockerfile.README.md
2. Give the `!bridgeversion` command to the bridge bot.

**Expected behavior**
The bridge bot should respond with something like `BridgeVersion: 0.23.0`, but instead if responds with `BridgeVersion: unknown`.

Half-Shot:
This should fix #1205

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,6 +33,7 @@ RUN mkdir ./data
 
 COPY --from=freebind /freebindfree/libfreebindfree.so /app/libfreebindfree.so
 COPY --from=builder /build/node_modules /app/node_modules
+COPY --from=builder /build/package.json /app/package.json
 COPY --from=builder /build/lib /app/lib
 
 COPY app.js config.schema.yml /app/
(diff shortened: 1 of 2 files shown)


uhoreg (approved):
That looks like it will do what it claims to do.

PR #1215, opened 2021-01-08
Use replies for admin command responses

### LINKED ISSUE #1214: !plumb and other admin commands should use replies for automated interactions
I tried to do this but the MatrixAction class looks nontrivial to modify.

Half-Shot:
Fixes #1214 

This basically refactors the admin room handler class to return the `MatrixAction` rather than send it inside each command function, and then attach the parent event_id to the `MatrixAction` before sending it. It's a pretty nice refactor :)

image

### DIFF
--- a/changelog.d/1215.feature
+++ b/changelog.d/1215.feature
@@ -1,0 +1,1 @@
+Use replies when responding to admin commands.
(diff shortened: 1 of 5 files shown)


turt2live (approved):
looks like js

unfortunate we don't have fallback support

Issue #1216, opened 2021-01-08
Self-help commands for third party bridged users


vranki:
This feature came up when a Telegram user bridged to Matrix and to IRC got a stupid nickname by default.
Changing nick in Telegram does not change the nick automatically on IRC side. Currently there's no way for
user to change nick as bridged users can't talk to the integration bot.

Suggestion:

Add "self-help" commands that can be given in any IRC bridged room. For example if user says "!irc nick newnick", bridge
would change that user's nick. There could be also other commands as needed but I suppose changing nick is
the most needed feature.

There's no need to limit these commands to bridged users - anyone could use them. Of course private chat with 
integration bot would be preferred when possible.

vranki:
#1217 implements this.

PR #1217, opened 2021-01-10
Allow changing nickname in any room

vranki:
Related to #1216 - the most important command is to change nick. This allows doing so from any room.


src/bridge/IrcBridge.ts
```diff
@@ -1202,6 +1202,18 @@ export class IrcBridge {
             "Sending IRC message in %s as %s (connected=%s)",
```
Half-Shot, line 1202: I worry that this is no longer quite true.

src/bridge/IrcBridge.ts
```diff
@@ -1202,6 +1202,18 @@ export class IrcBridge {
...
             ircRoom.channel, bridgedClient.nick, Boolean(bridgedClient.status === BridgedClientStatus.CONNECTED)
         );
+        if (action.text.startsWith("!irc nick")) {
```
Half-Shot, line 1205:
```suggestion
        if (action.text.startsWith("!irc nick ")) {
```

to check formatting.

src/bridge/IrcBridge.ts
```diff
@@ -1202,6 +1202,18 @@ export class IrcBridge {
...
+            const newNick = action.text.substr(10)
```
Half-Shot, line 1206: This is easier to read

```suggestion
            const newNick = action.text.substr("!irc nick ".length)
```

src/bridge/IrcBridge.ts
```diff
@@ -1202,6 +1202,18 @@ export class IrcBridge {
...
+            log.info("Matrix user wants to change nick from %s to %s", bridgedClient.nick, newNick)
+            try {
+               return bridgedClient.changeNick(newNick, true)
```
Half-Shot, line 1209: I wonder if we should await this one, and give the user some feedback if they couldn't change their nick?

Half-Shot (commented):
Some thoughts, looks handy though.

src/bridge/IrcBridge.ts
vranki: I suck at JavaScript and failed to do this. Added a todo comment.

vranki:
I did the requested changes. Someone more skilled could add the result message, but at least the nick change functionality works now.

vranki (commented):
Nice, this is better architecture allowing more commands to be implemented in future cleanly.

Half-Shot:
(Merging as we've both reviewed it)

PR #1220, opened 2021-01-19
Correct the comment above bindPort

xiaq:
The comment seems to get it the other way around: the port specified in the command line takes precedence over the one in the config file.

### DIFF
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -32,7 +32,7 @@ homeserver:
   # Default: true
   enablePresence: true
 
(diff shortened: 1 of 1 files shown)


Half-Shot (approved):
That's correct. Thanks!

Half-Shot:
Can you sign off as per https://github.com/matrix-org/matrix-appservice-irc/blob/develop/CONTRIBUTING.md#making-pull-requests

xiaq:
I can just post it as a comment, right? Here it is:

Signed-off-by: Qi Xiao <xiaqqaix@gmail.com>

PR #1228, opened 2021-01-25
Debounce performance improvements

Half-Shot:
Currently the debouncer does too much work when a netsplit is ongoing. It will immediately try to send QUITs for a channel as soon as it sees anyone join it from the IRC side, and it will also repeatedly send those QUITs because the list of nicks is not removed afterwards. This PR aims to fix those things.

This PR changes the debounce code so that:
- We now log when a debounce has begun, for easy tracing in the logs.
- We *only* try to send QUITs after a random interval of time (between min,max in the config), rather than as soon as we see the first join.
- :exclamation:  We do not **reprocess** the QUIT queue for the same channel

### DIFF
--- a/changelog.d/1228.bugfix
+++ b/changelog.d/1228.bugfix
@@ -1,0 +1,1 @@
+Fix an issue where the QuitDebouncer would reprocess old QUITs, and process QUITs too early during the debouncing process.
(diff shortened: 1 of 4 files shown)


benparsons (approved):
lgtm

PR #1230, opened 2021-01-27
Fix thundering herd problem with debouncer

Half-Shot:
There is a bug in the debouncer code that would often delay the debounce for the min period rather than a random period. This change ensures that two debounce efforts are hopefully not processed at the same time.

### DIFF
--- a/src/bridge/QuitDebouncer.ts
+++ b/src/bridge/QuitDebouncer.ts
@@ -61,6 +61,7 @@ export class QuitDebouncer {
             // Nobody to debounce, yay.
             return;
         }
+        const delay = server.getQuitDebounceDelay();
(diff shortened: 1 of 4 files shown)


jaller94 (approved):
Instead of having private functions (which are only used once), we could also access the variables directly.

src/bridge/QuitDebouncer.ts
```diff
@@ -61,15 +61,14 @@ export class QuitDebouncer {
             // Nobody to debounce, yay.
             return;
         }
-        const delay = Math.max(server.getQuitDebounceDelayMinMs(), server.getQuitDebounceDelayMaxMs() * Math.random());
         log.info(`Will attempt to reconnect users for ${channel} after ${delay}ms`)
```
jaller94, line 64: This line uses `delay` which is no longer defined.

PR #1231, opened 2021-01-27
Debounce joins to improve performance

Half-Shot:
This was a quick win that struck me as I read the code.

(Note to reviewer: We debounce quits when they come in great numbers and defer them for later so that the bridge doesn't melt)

When we handle QUITs for users while debouncing, we perform no action and store the quit in memory instead. However when we see the user join the channel again, we remove them from memory BUT process the join. In reality we don't need to take any action if we have seen a QUIT>JOIN as the user will still be joined to the room.

With this in mind, we do a fairly simple check to ensure that if a user is in the quitters pool, we will take no action to join them which will save a few CPU cycles.

### DIFF
--- a/src/irc/IrcEventBroker.ts
+++ b/src/irc/IrcEventBroker.ts
@@ -361,6 +361,12 @@ export class IrcEventBroker {
                 req, server, createUser(nick), chan, "part", reason
             ));
         });
+        this.hookIfClaimed(client, connInst, "kick", (chan: string, nick: string, by: string, reason: string) => {
+            const req = createRequest();
+            complete(req, ircHandler.onKick(
+                req, server, createUser(by), createUser(nick), chan, reason
+            ));
+        });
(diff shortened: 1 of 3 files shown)


src/irc/IrcEventBroker.ts
```diff
@@ -361,6 +361,12 @@ export class IrcEventBroker {
                 req, server, createUser(nick), chan, "part", reason
             ));
         });
+        this.hookIfClaimed(client, connInst, "kick", (chan: string, nick: string, by: string, reason: string) => {
+            const req = createRequest();
+            complete(req, ircHandler.onKick(
+                req, server, createUser(by), createUser(nick), chan, reason
+            ));
+        });
         this.hookIfClaimed(client, connInst, "quit", (nick: string, reason: string, chans: string[]) => {
```
Half-Shot, line 370: I moved the quit function to be near the join function so that it would appear

benparsons: clever

benparsons (approved):
lgtm

PR #1232, opened 2021-01-29
Allow for configurable ping time

Half-Shot:
This PR allows us to configure how long to wait for pinging out, which helps for large bridges for freenode who have given us a slightly increased ping time to allow the node process some breathing room.


jaller94 (approved):
lgtm

PR #1237, opened 2021-02-12
Properly urlencode media urls

### LINKED ISSUE #1219: The bridge will send url with spaces on it for pasted images
**Describe the bug**
When pasting an image with spaces in it, IRC will see something like this:
> \* nick uploaded an image: Screenshot from 2021-01-11 19-03-08.png (332KiB) < https://matrix.org/_matrix/media/r0/download/gnugen.ch/eIgbEuqYFfYzZuWeKqllgKlJ/Screenshot from 2021-01-11 19-03-08.png >

**To Reproduce**
* join an IRC channel (both from Matrix and IRC)
* send a picture with spaces in the name
* admire how the link isn't clickable on IRC

**Expected behavior**
The URL is URL-encoded.

Half-Shot:
Fixes #1219 

This PR url encodes the filename parameter of a media file when sending to IRC and removes the filaname from the beginning of the message, if the body is detected to be a filename.

### DIFF
--- a/src/models/MatrixAction.ts
+++ b/src/models/MatrixAction.ts
@@ -192,7 +192,7 @@ export class MatrixAction {
                 }
                 else {
                     // If not a filename, print the body
-                    text = `${event.content.body} ${fileSize} < ${url} >`;
+                    text = `${event.content.body}${fileSize} < ${url} >`;
                 }
             }
         }
--- a/src/models/MatrixAction.ts
+++ b/src/models/MatrixAction.ts
@@ -187,9 +187,13 @@ export class MatrixAction {
                 }
 
                 if (filename) {
-                    url += `/${filename}`;
+                    url += `/${encodeURIComponent(filename)}`;
+                    text = `${fileSize} < ${url} >`;
                 }
-                text = `${event.content.body}${fileSize} < ${url} >`;
+                else {
+                    // If not a filename, print the body
+                    text = `${event.content.body} ${fileSize} < ${url} >`;
+                }
             }
         }
         return new MatrixAction(type, text, htmlText, event.origin_server_ts);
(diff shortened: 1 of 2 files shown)


src/models/MatrixAction.ts
```diff
@@ -187,9 +187,13 @@ export class MatrixAction {
                 }
 
                 if (filename) {
-                    url += `/${filename}`;
+                    url += `/${encodeURIComponent(filename)}`;
+                    text = `${fileSize} < ${url} >`;
+                }
+                else {
+                    // If not a filename, print the body
+                    text = `${event.content.body} ${fileSize} < ${url} >`;
```
jaller94, line 195: What do we expect to be in the body? An accompanying message from the Matrix user?
Do we not post that if there's a filename?

src/models/MatrixAction.ts
```diff
@@ -187,9 +187,13 @@ export class MatrixAction {
                 }
 
                 if (filename) {
-                    url += `/${filename}`;
+                    url += `/${encodeURIComponent(filename)}`;
+                    text = `${fileSize} < ${url} >`;
```
jaller94, line 191: If it's not an empty string, `fileSize` starts with a space.
Both lines that use `fileSize` in your change seem to expect `fileSize` to not have an additional space.

jaller94 (changes requested):
I don't know why body is important or not important based on the filename.

src/models/MatrixAction.ts
Half-Shot: The Matrix spec is a bit weird regarding this https://matrix.org/docs/spec/client_server/r0.6.0#m-file

>Required. A human-readable description of the file. This is recommended to be the filename of the original upload.

Which means it could be either a filename or a human-readable description. We have to do a quick regex check above to try to understand which one it is.

src/models/MatrixAction.ts
jaller94: This will still start with a space.
I don't think that's what you expect.

jaller94 (approved):
Looks good. Though the `text` still starts with a space if `filename` is truthy.
Would be good to fix that.

PR #1245, opened 2021-02-19
Add a configuration option to drop joins entirely on debounce

Half-Shot:
We typically join new IRC usrs to a room during a debounce if the user was not previously joined, which is usually fine but if the bridge is of a certain size and the netsplit is serious enough, even that can cause issues. This PR adds a config option to drop membership if a channel is debouncing.


config.sample.yaml
```diff
@@ -113,6 +113,10 @@ ircService:
         delayMinMs: 3600000 # 1h
         # Default: 7200000, = 2h
         delayMaxMs: 7200000 # 2h
+        # Should the bridge drop any joins that happen during the debounce period. If your
+        # bridge suffers from extensive downtime when debouncing, this might help smooth
+        # things over.
```
jaller94, line 118: What does smooth things over mean?
It would be good to highlight the advantages and downfalls of this, similar to what you wrote in the PR description.
```suggestion
        # bridge suffers from extensive downtime when debouncing, this might keep the bridge
        # operable by not adding new IRC users as Matrix room members. A new users will
        # get added when they post a message.
```

config.sample.yaml
```diff
@@ -113,6 +113,10 @@ ircService:
         delayMinMs: 3600000 # 1h
         # Default: 7200000, = 2h
         delayMaxMs: 7200000 # 2h
+        # Should the bridge drop any joins that happen during the debounce period. If your
```
jaller94, line 116:
```suggestion
        # Should the bridge drop any IRC->Matrix joins that happen during the debounce period. If your
```

jaller94 (approved):
This looks ok to me.
As mentioned, I think the option's description in config.sample.yaml could get improved.

src/bridge/QuitDebouncer.ts
```diff
@@ -54,13 +54,9 @@ export class QuitDebouncer {
...
             return true;
         }
-        if (set.size === 0) {
-            // Nobody to debounce, yay.
-            return true;
-        }
         if (!set.delete(nick)) {
             // This user did NOT quit the channel, so we should treat them as a new joiner and handle immediately.
-            return true;
+            return server.shouldDropJoinsDuringDebounce();
```
Half-Shot, line 59: This probably needs inverting?

jaller94, line 59: Good point. Given the JSDocs of both funtions this should get inverted.
```suggestion
            return !server.shouldDropJoinsDuringDebounce();
```

Half-Shot:
This hasn't made it in, and I'm not 100% convinced we still need it.

PR #1249, opened 2021-02-25
Upgrade Dockerfile nodejs 12 => 14

3nprob:
14 is the current LTS as per 2020-10

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Freebind build
-# node:12-slim uses debian:stretch-slim as a base, so it's safe to build on here.
+# node:14-slim uses debian:stretch-slim as a base, so it's safe to build on here.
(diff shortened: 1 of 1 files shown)


Half-Shot (commented):
Seems like a reasonable change, can we have a changelog and signoff as per https://github.com/matrix-org/matrix-appservice-irc/blob/develop/CONTRIBUTING.md

Half-Shot:
@3nprob ping?

Half-Shot:
Sorry, we're going with #1299 because that has a changelog.

3nprob:
Ah sorrry missed the notification on this one @Half-Shot 

All good, progress is what matters, not whose pseudonym shows up in the commit log ;)

Thanks for maintaining!

PR #1255, opened 2021-03-01
Catch channel join failures during reconnect

Half-Shot:
Previously if we failed to rejoin a channel, we would exit the loop prematurely and not join any other channels either. This PR ensures that we only log a failure to rejoin


benparsons (approved):
lgtm

Issue #1250, opened 2021-02-25
Some IRC users are not bridged


piegamesde:
**Describe the bug**
For some reason, not all users in an IRC room are bridged to the Matrix side when joining a channel. This results in them "joining" the Matrix room the first time they speak which is confusing.

**To Reproduce**
I don't have any explicit reproduction steps. Joining small IRC rooms makes the problem more obvious though. I mainly noticed this on freshly joined IRC rooms.

I have enabled all options on `membershipLists` in the configuration.

**Expected behavior**
All users that are on the IRC side are on the Matrix side as well.

**Additional context**
This is is really similar to #371, but in the opposite bridging direction.

Half-Shot:
Logs would be useful here I think, if you've enabled the membership options then I'd expect it to work. Be sure to enable debug logging.

Half-Shot:
The user is running the bridge with the botEnabled config, which has subtly different logic, see the linked PR.

PR #1256, opened 2021-03-01
Fix race around irc->matrix membership bridging for bot mode bridges

Half-Shot:
Fixes https://github.com/matrix-org/matrix-appservice-irc/issues/1250

This comes down to some confusing logic in the bridge when users attempt to bridge a new IRC channel, what would typically happen is:

- User tries a `!join #channel` or a `/join #irc_#channel:example.com`
- The bridge will try to join the channel **if the irc bridge bot is configured**
- The bridge will join the channel and immediately get NAMES information on the IRC side, and try to bridge I->M
- It will fail because the room isn't stored yet
- The bridge joins and the ircChannel,server,roomId mapping gets stored
- The room is now stored, but it's too late because the irc side has already sent the membership info and we missed it.

However if the bridge bot isn't configured then the bridge will only join the IRC channel **after** the mapping is stored, which means the bridge will be able to handle the NAMES information. This is typically the mode that most bridges run under.

Because it's very hard to make the bridge join the channel first without either:
1) Lots of deferred storage of membership until the room is mapped, creating a lot of fragile code and further increasing our memory footprint
2) Re-requesting the membership, causing additional load.

I'm going to make it so that we never join the IRC channel in either mode, as we seem to have been happy to do without it on our major bridges anyway. This also means our logic for bot and botless bridges are identical which should avoid surprises.

### DIFF
--- a/changelog.d/1256.bugfix
+++ b/changelog.d/1256.bugfix
@@ -1,0 +1,1 @@
+Fix an issue where IRC membership would not be bridged to new rooms when `botConfig.enabled` is `true`. 
--- a/src/bridge/IrcBridge.ts
+++ b/src/bridge/IrcBridge.ts
@@ -1145,21 +1145,6 @@ export class IrcBridge {
         return new IrcUser(ircInfo.server, ircInfo.nick, true);
     }
 
-    public async trackChannel(server: IrcServer, channel: string, key?: string): Promise<IrcRoom> {
-        if (!server.isBotEnabled()) {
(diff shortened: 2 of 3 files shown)


benparsons (approved):

benparsons:
Can the text here be captured somehow? Might want to include a version of it in the source somehow?

Half-Shot:
I'll link back to this PR in RoomCreation.ts, because I think the logic largely speaks for itself.

PR #1257, opened 2021-03-01
Immediately remove PLs from users when unbridging

Half-Shot:
Related to #1248 

A while ago we altered the bridge code to asynchronously alter the PLs of users so that we would send fewer PL events into the room so that if 5 users all needed changing, we would debounce and send one change. However, this causes problems for unbridging where we will:

1. Tell the bridge to remove PLs for the irc ghosts in the room
2. Leave the IRC bridge bot from the room
3. Re-join the IRC bridge bot as 1. will execute asynchronously.
4. This leaves the IRC bridge in the room forever.

### DIFF
--- a/changelog.d/1257.bugfix
+++ b/changelog.d/1257.bugfix
@@ -1,0 +1,1 @@
+Update powerlevels immediately when unbridging to avoid rejoining the bridge bot to the room.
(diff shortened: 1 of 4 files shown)


benparsons (approved):
lgtm

PR #1258, opened 2021-03-01
Remove admin/pm rooms when the matrix-side user leaves to allow server to GC

### LINKED ISSUE #1207: Private/direct message history retention
As a Matrix user, I expect the history of a direct message room to be destroyed when both users leave said room.
IRC users however do not have the notion of persistent rooms and cannot leave the direct message rooms created by the bridge.

1. Is the Matrix user leaving a two-party conversation involving some IRC user enough for the room to be deleted?
   Is the virtual user of the IRC participant automatically leaving the room as well?
   A ghost user haunting a room alone would probably prevent the room from being garbage-collected.
2. Is the content of direct message rooms with no participants actually deleted from the involved homeserver(s)?
   Or if it is just left unreachable, when is the deletion happening?

Half-Shot:
Fixes #1207 

The idea here is that the rooms serve no purpose once the Matrix user has left them, so we might as well leave the bridge users from the room too and then Synapse can garbage collect the room (as it is unjoinable). Of course, https://github.com/matrix-org/synapse/issues/4720 needs to happen first.

### DIFF
--- a/changelog.d/1258.misc
+++ b/changelog.d/1258.misc
@@ -1,0 +1,1 @@
+Leave DM rooms and admin rooms if the Matrix user leaves so that a homeserver may clear them up later.
--- a/src/bridge/MatrixHandler.ts
+++ b/src/bridge/MatrixHandler.ts
@@ -719,9 +719,34 @@ export class MatrixHandler {
             serverLookup[ircClient.server.domain] = ircClient;
         });
 
+        const store = this.ircBridge.getStore();
 
         // which channels should the connected client leave?
-        const ircRooms = await this.ircBridge.getStore().getIrcChannelsForRoomId(event.room_id);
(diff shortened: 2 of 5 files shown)


benparsons (approved):
lgtm

pacien:
If I understand this correctly, this makes ghost users leave rooms only when real Matrix users are leaving them.

Is this also handling the case in which the real Matrix user already has left the room in the past (before this patch), leaving some rooms with only ghost users?

Half-Shot:
>Is this also handling the case in which the real Matrix user already has left the room in the past (before this patch), leaving some rooms with only ghost users?

This would probably be better handled by a script that iterates the database, parts the ghost and deletes the entry. I can probably write one up in a separate PR.

PR #1261, opened 2021-03-04
config.schema.yml: fix permissions schema

mweinelt:
The key reflects a key, value mapping from mxid to "admin", and
additionalProperties requires this to be either a boolean or an object,
but it is currently defined as a list. This breaks validation for us.

> The additionalProperties keyword may be either a boolean or an object

https://json-schema.org/understanding-json-schema/reference/object.html#id2

And objects to me sound like they describe an associative array, or
dictionary:

> Objects are the mapping type in JSON. They map “keys” to “values”.

https://json-schema.org/understanding-json-schema/reference/object.html#object

So let's fix this by replacing the spurious list definition with an
actual object.

### DIFF
--- a/config.schema.yml
+++ b/config.schema.yml
@@ -113,7 +113,7 @@ properties:
                     enableReload:
                         type: "boolean"
                     roomLimit:
-                        type: "number" 
+                        type: "number"
             passwordEncryptionKeyPath:
                 type: "string"
             matrixHandler:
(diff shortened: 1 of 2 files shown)


config.schema.yml
```diff
@@ -113,7 +113,7 @@ properties:
                     enableReload:
                         type: "boolean"
                     roomLimit:
-                        type: "number" 
+                        type: "number"
```
mweinelt, line 116: drive-by trailing whitespace fix

mweinelt:
I can write a newsblock when we agree this is the correct fix. I am by no means a JSON schema expert.

Munksgaard:
This may be blocking https://github.com/NixOS/nixpkgs/pull/113988

Half-Shot (approved):
This looks fine to me. Thanks! (Please sign off and newsfile it :+1: )

mweinelt:
This is ready.

PR #1262, opened 2021-03-04
Use BridgeInfoStateSyncer (and update to matrix-appservice-bridge 2.6.0-rc1)

Half-Shot:
This change does the following:
- Updates us to use matrix-appservice-bridge 2.6.0-rc1
- Use Typescript 4.2 (from 3.7.*)
- Fix a config bug where we specified https:// urls rather than mxc:// urls.


benparsons (approved):
lgtm

Issue #1272, opened 2021-03-25
Docker image overrides config port


abdulocracy:
**Describe the bug**
The Docker image overrides the configured binding port, so I had issues when upgrading to the latest Docker image (0.25.0), as the port was changed to 9999. My config was still on 9995, but the bridge wasn't using it, therefore the homeserver couldn't reach the bridge, and therefore I had major issues that went unnoticed for a while.

Once I fixed it by adjusting the registration file to use port 9999, all the messages I had sent since last Friday were spammed into channels.

**To Reproduce**
1. Use Docker image
2. Configure port other than the default 9999 in the config file
3. Watch as it doesn't change

**Expected behavior**
It should change based on the configured port

Half-Shot:
We did highlight the port change in the changelog, but I suspect we could have made it more obvious that this would break configuration. However, the bridge *does* warn in the logs if there is a connectivity issue so I would caution to check your logs after upgrading to make sure that things are working after upgrade (especially as we are pre-1.0).

PR #1274, opened 2021-03-26
Doc changes: Unify use of port 9999, the registration file name, and other minor changes

jaller94:
- The `DEFAULT_PORT` is 8090 but the config sample suggested it may be 9999 or 9995.
- The registration file changed names during the guide.


Dockerfile.README.md
```diff
@@ -39,4 +39,4 @@ If you are storing passwords for users, you should also run:
 openssl genpkey -out ./dockerdata/passkey.pem -outform PEM -algorithm RSA -pkeyopt rsa_keygen_bits:2048
 ```
 
-You can now run your shiny new image using `docker run -p 9999:9999 -v $PWD/dockerdata:/app/data`.
+You can now run your shiny new image using `docker run --publish 8090 --volume $PWD/dockerdata:/app/data`.
```
Half-Shot, line 42: As per https://github.com/matrix-org/matrix-appservice-irc/issues/1272 we've already hit issues when we changed the port to 9999. I would sooner change our default to 9999 than encounter this mess again :(

Half-Shot (changes requested):

Dockerfile.README.md
jaller94: I don't think we can ever change `DEFAULT_PORT` as that would change it for everyone who's not set it up in their config or CLI options.
https://github.com/matrix-org/matrix-appservice-irc/blob/73545600b848e816694ec2ebe0c33b65f7adc323/src/bridge/IrcBridge.ts#L47

That's why I am proposing to change the Docs.

Half-Shot: You're also changing the scripts / default config and lots of non doc things, in particular changing the docker start script will break existing setups (again), so we need to tread carefully here.

jaller94: Ok, I reverted most of that.
In docs/bridge_setup.md the port changed midway through the guide from 9999 to 9995, so I changed that to 9999.

Half-Shot (approved):
Thanks for bearing with, this looks good to me

PR #1275, opened 2021-03-26
Trim Markdown code block syntax

### LINKED ISSUE #1271: ``` should be stripped instead of triggering the "long message" mechanism
**Is your feature request related to a problem? Please describe.**

When sending this on Matrix:

````
```
a
b
```
````

Matrix sends a huge pastebin link, such as:

```
* pinkie[m] < https://matrix.org/_matrix/media/r0/download/matrix.org/ZBiUWxhBCkRMCFIIXsfEYyTP/message.txt >
```

**Describe the solution you'd like**

Strip the backticks and send this instead:

```
<pinkie> a
<pinkie> b
```

And when a Matrix user sends

````
```
a
```
````

the bridge should not send this:

````
15:28 <pinkie[m]> ```
15:28 <pinkie[m]> a
15:28 <pinkie[m]> ```
````

but this:

```
15:28 <pinkie[m]> a
```

Because lines with the three backticks are just noise for IRC users, no client interprets Markdown (especially across lines) as far as I know. EDIT: actually IRCCloud does parse the backticks

However, the bridge could prefix such messages with [the `0x11` byte](https://modern.ircdocs.horse/formatting.html#monospace), which indicates clients should format it with a monospace font.

**Additional context**
This happens to me frequently with people sharing short snippets.

### PR DESCRIPTION
**Pull request by jaller94**
Fixes #1271

Removes the code block syntax and [adds the monospace character `\x11` to the beginning of each line](https://modern.ircdocs.horse/formatting.html#monospace).

If the trimmed message is longer than the `lineLimit`, it still gets uploaded instead of posted directly.

Screenshot_2021-03-26_13-30-13

### DIFF
--- a/src/bridge/MatrixHandler.ts
+++ b/src/bridge/MatrixHandler.ts
@@ -975,7 +975,7 @@ export class MatrixHandler {
             return;
         }
 
-        const potentialCodeText = trimCodeMessages(text)
+        const potentialCodeText = trimCodeMessages(text);
         if (potentialCodeText) {
             ircAction.text = text = potentialCodeText;
         }
--- a/src/bridge/MatrixHandler.ts
+++ b/src/bridge/MatrixHandler.ts
@@ -955,7 +955,7 @@ export class MatrixHandler {
             return;
         }
 
-        let cacheBody = event.content.body;
+        let cacheBody = ircAction.text;
         if (event.content["m.relates_to"] && event.content["m.relates_to"]["m.in_reply_to"]) {
             const eventId = event.content["m.relates_to"]["m.in_reply_to"].event_id;
             const reply = await this.textForReplyEvent(event, eventId, ircRoom);
@@ -964,8 +964,13 @@ export class MatrixHandler {
                 cacheBody = reply.reply;
             }
         }
+        let body = cacheBody.trim().substr(0, REPLY_SOURCE_MAX_LENGTH);
+        const nextNewLine = body.indexOf('\n');
+        if (nextNewLine !== -1) {
+            body = cacheBody.substr(0, nextNewLine);
+        }
         this.eventCache.set(event.event_id, {
-            body: cacheBody.substr(0, REPLY_SOURCE_MAX_LENGTH),
+            body,
             sender: event.sender
         });
 
@@ -1005,7 +1010,7 @@ export class MatrixHandler {
         try {
             // Try to upload as a file and get URI
             //  (this could fail, see the catch statement)
-            contentUri = await this.ircBridge.uploadTextFile(fileName, event.content.body);
+            contentUri = await this.ircBridge.uploadTextFile(fileName, ircAction.text);
         }
         catch (err) {
             // Uploading the file to HS could fail
@@ -1207,8 +1212,9 @@ export class MatrixHandler {
             );
         }
 
+        const RESET_CODE = '\u000f';
         return {
-            formatted: `<${rplName}${rplSource}> ${rplText}`,
+            formatted: `<${rplName}${rplSource}${RESET_CODE}> ${rplText}`,
             reply: rplText,
         };
     }
--- a/src/bridge/MatrixHandler.ts
+++ b/src/bridge/MatrixHandler.ts
@@ -950,18 +950,17 @@ export class MatrixHandler {
     private async sendIrcAction(req: BridgeRequest, ircRoom: IrcRoom, ircClient: BridgedClient, ircAction: IrcAction,
         event: MatrixMessageEvent) {
         // Send the action as is if it is not a text message
-        let text = event.content.body;
-        if (event.content.msgtype !== "m.text" || !text) {
+        if (event.content.msgtype !== "m.text" || !event.content.body) {
             await this.ircBridge.sendIrcAction(ircRoom, ircClient, ircAction);
             return;
         }
 
-        let cacheBody = text;
+        let cacheBody = event.content.body;
         if (event.content["m.relates_to"] && event.content["m.relates_to"]["m.in_reply_to"]) {
             const eventId = event.content["m.relates_to"]["m.in_reply_to"].event_id;
             const reply = await this.textForReplyEvent(event, eventId, ircRoom);
             if (reply !== null) {
-                ircAction.text = text = reply.formatted;
+                ircAction.text = reply.formatted;
                 cacheBody = reply.reply;
             }
         }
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: src/bridge/MatrixHandler.ts [hunk 1]
```diff
@@ -956,6 +975,11 @@ export class MatrixHandler {
             return;
         }
 
+        const potentialCodeText = trimCodeMessages(text);
```
Half-Shot, line 978: Agreed that this isn't the best function to be doing this on. My wish is that we refactor this function to act as a pipeline of statements that modify a given event into an IRC message in a stateless context. Perhaps a better place would be `IrcAction.fromMatrixAction`, which is intended to be the point where we format a Matrix Event into an IRC action.

src/bridge/MatrixHandler.ts
```diff
@@ -23,6 +23,25 @@ async function reqHandler(req: BridgeRequest, promise: PromiseLike<unknown>) {
...
+    let trimmedText = text.trim();
+    // If this post isn't all code, ignore it.
+    if (!/^```.*\n.*```$/s.test(trimmedText)) {
+        return undefined;
+    }
+    // Remove the first line (e.g. ```js) and the ``` at the end
+    trimmedText = trimmedText.substring(trimmedText.indexOf('\n'), trimmedText.length - 3).trim();
+    // Add monospace character at the start of each line
+    // https://modern.ircdocs.horse/formatting.html#monospace
+    trimmedText = trimmedText.replace(/\n/, '\n\x11');
```
Half-Shot, line 41: oooh fancy

src/bridge/MatrixHandler.ts
```diff
@@ -23,6 +23,25 @@ async function reqHandler(req: BridgeRequest, promise: PromiseLike<unknown>) {
     }
 }
 
+/**
+ * Returns a trimmed string if the input text is all Markdown-style code.
+ * This is used to allow small code snippets to look nice in IRC and not be
+ * transformed into an upload.
+ */
+function trimCodeMessages(text: string): string|undefined {
```
Half-Shot, line 31: If you export this, it should be easy to write a simple unit test for it.

Half-Shot (commented):
Heading in the right direction :+1: . The formatter looks good. Overall I wish our Matrix->IRC formatting code was less clunky (I would favour a pipeline of sorts) but I don't think we should go for a grand refactor here.

src/bridge/MatrixHandler.ts
jaller94: I don't know how to test this, so I removed it for now.

jaller94:
Known issues that remain:
- If a Matrix user replies to a code block, the message gets uploaded. (with the default line limit of 3, because in the reply the code delimiters are not trimmed.

Half-Shot:
> If a Matrix user replies to a code block, the message gets uploaded. (with the default line limit of 3, because in the reply the code delimiters are not trimmed.

This is surprising, because I believe we have our own reply format which only includes a truncated form of the original message. Would it be possible to fix that?

src/bridge/MatrixHandler.ts
```diff
@@ -969,23 +950,17 @@ export class MatrixHandler {
...
             return;
         }
 
-        const potentialCodeText = trimCodeMessages(text);
-        if (potentialCodeText) {
-            ircAction.text = text = potentialCodeText;
-        }
-
-        let cacheBody = text;
+        let cacheBody = event.content.body;
```
Half-Shot, line 958: This changes the code so that we now cache the body, rather than `trimCodeMessages(body)`, which seems like a potentially dangerous change?

src/bridge/MatrixHandler.ts
```diff
@@ -1030,7 +1005,7 @@ export class MatrixHandler {
         try {
             // Try to upload as a file and get URI
             //  (this could fail, see the catch statement)
-            contentUri = await this.ircBridge.uploadTextFile(fileName, text);
+            contentUri = await this.ircBridge.uploadTextFile(fileName, event.content.body);
```
Half-Shot, line 1008: Not `ircAction.text`?

src/models/IrcAction.ts
```diff
@@ -42,8 +42,9 @@ export class IrcAction {
                     break;
                 }
                 if (matrixAction.htmlText) {
-                    const text = ircFormatting.htmlToIrc(matrixAction.htmlText);
-                    const ircText = text ?? matrixAction.text; // fallback if needed.
+                    const ircText = ircFormatting.htmlToIrc(matrixAction.htmlText)
```
Half-Shot, line 45: Does `htmlToIrc` handle codeblocks?

Half-Shot (commented):
Some more thoughts

src/models/IrcAction.ts
jaller94: No, it will reject `<pre>` and `<code>`

src/bridge/MatrixHandler.ts
jaller94: `ircAction.text` is a modified version with non-printable characters for IRC formatting.
If we decide not to send it via IRC, we may as well upload the original Matrix body.

Half-Shot: Fair enough

src/models/IrcAction.ts
Half-Shot: As in the contents of those elements? Sounds like we aught to include that. In fact, if you are sending code blocks over element, it will always be HTML?

Half-Shot: Oh I see `htmlToIrc` will return null if the HTML contained tags it doesn't recognise, so this is fine.

src/bridge/MatrixHandler.ts
```diff
@@ -964,8 +964,13 @@ export class MatrixHandler {
                 cacheBody = reply.reply;
             }
         }
+        let body = cacheBody.trim().substr(0, REPLY_SOURCE_MAX_LENGTH);
```
Half-Shot, line 967: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr is technically deprecated in favour of substring.

src/bridge/MatrixHandler.ts
```diff
@@ -964,8 +964,13 @@ export class MatrixHandler {
...
+        const nextNewLine = body.indexOf('\n');
```
Half-Shot, line 968: We tend to use " quotes here.

Half-Shot (approved):
Couple of nits, otherwise okay. CI looks unhappy btw

ProgVal:
Thanks!

PR #1282, opened 2021-04-08
Add retry mechanism to the creation of PM rooms

jaller94:
Fixes #1254

Issue #1254, opened 2021-03-01
Retry creating a DM room on failure


Half-Shot:
Sometimes the request can fail with a 500 and it's not clear why, but dropping DMs is v.bad

Half-Shot:
We should also create the room, and then invite the user. Sending an invite via createRoom can cause the request to fail if the remote server is unreachable, and retaining history should be our priority. We can do a backoff loop for inviting the users separately.

PR #1290, opened 2021-04-13
Decouple invite from the creation of a PM room

jaller94:
Issue: https://github.com/matrix-org/matrix-appservice-irc/issues/1254#issuecomment-788156440

Split of from: https://github.com/matrix-org/matrix-appservice-irc/pull/1282#discussion_r609654092


src/bridge/IrcHandler.ts
```diff
@@ -167,7 +169,6 @@ export class IrcHandler {
                         visibility: "private",
                         // We deliberately set our own power levels below.
                         // preset: "trusted_private_chat",
-                        invite: [toUserId],
```
Half-Shot, line 170: Hold on a sec, are we actually inviting the user at all now?

Half-Shot (commented):
I think this is okay, but are we inviting the user?

src/bridge/IrcHandler.ts
jaller94: `ensureMatrixUserJoined` should join the user right after the creation.

https://github.com/matrix-org/matrix-appservice-irc/blob/7dc74b92cd6906619e53c262ea14682fd901fb8f/src/bridge/IrcHandler.ts#L308-L311

jaller94: Works for me.
## Test plan
1. Rebuild & Restart bridge with this branch
2. As an IRC user, message a Matrix user.
3. Use `/nick new-name` to a new username.
4. Message the Matrix user again.

https://user-images.githubusercontent.com/10872136/116164333-e260e300-a6f9-11eb-968f-4f6beea07970.mp4

Half-Shot (approved):
I think this looks great!

Issue #1284, opened 2021-04-08
Add a config option to prevent kicking Matrix users who can't join a channel


Half-Shot:
This has been requested in the case of personal bridges and/or bridges where the Matrix side is considered authoritive. It would also be good for the IRC operator to allow this on a per-channel basis.

PR #1294, opened 2021-04-15
Add config option to disable kicking Matrix users from rooms

Half-Shot:
This setting allows a bridge to selectively opt out of kicking Matrix users depending on the reason for the kick. This solves #1284 for self hosters, while matrix.org users will need to wait for us to release a per-channel config option.


jaller94 (approved):
Can't see  anything wrong with this. :+1:

Half-Shot:
Christian says looks good :partying_face:

PR #1298, opened 2021-04-22
Add !unlink command

### LINKED ISSUE #1297: Add !unbridge admin command
To allow bridge admins to unbridge rooms from their admin rooms in a similar vein to #1211 

As an extra feature, it would be nice for people with permission (but not bridge admin rights) to unbridge channels from plumbed rooms.

jaller94:
The Matrix user needs to be in the Matrix room and have Moderator permissions for this to work.

This command is the inverse of #1211.

### DIFF
--- a/changelog.d/1298.feature
+++ b/changelog.d/1298.feature
@@ -1,1 +1,1 @@
-Add !unlink command
+Added an !unlink command for Matrix users to unbridge two rooms
--- a/changelog.d/1298.feature
+++ b/changelog.d/1298.feature
@@ -1,0 +1,1 @@
+Add !unlink command
--- a/changelog.d/1298.feature
+++ b/changelog.d/1298.feature
@@ -1,1 +1,1 @@
-Added an !unlink command for Matrix users to unbridge two rooms
+Added an !unlink command for Matrix users to unbridge a channel from Matrix
--- a/docs/admin_room.md
+++ b/docs/admin_room.md
@@ -99,7 +99,7 @@ commands are reserved for bridge administrators, which can be configured in the 
 
 ### `!plumb`
 
-`!feature !room:example.com irc.network.net #channel`
+`!plumb !room:example.com irc.network.net #channel`
 
 *This command only works for bridge administrators*
 
@@ -108,6 +108,17 @@ validate that you have permission to do this on the IRC channel so please take c
 aware of your actions.
 
 You must invite the bridge bot into the Matrix room for this to work.
+
+
+### `!unlink`
+
+`!unlink !room:example.com irc.network.net #channel`
+
+*This command only works for moderators of a bridged Matrix room and bridge administrators*
+
+This command allows you to unlink a IRC channel from a room without using the HTTP provisioning API. For bridge
+administrators this command does NOT validate that you have permission to do this on the IRC channel so please
+take care to ensure that the IRC channel is aware of your actions.
 
 
 ### `!help`
--- a/docs/admin_room.md
+++ b/docs/admin_room.md
@@ -116,9 +116,7 @@ You must invite the bridge bot into the Matrix room for this to work.
 
 *This command only works for moderators of a bridged Matrix room and bridge administrators*
 
-This command allows you to unlink a IRC channel from a room without using the HTTP provisioning API. For bridge
-administrators this command does NOT validate that you have permission to do this on the IRC channel so please
-take care to ensure that the IRC channel is aware of your actions.
+This command allows you to unlink a IRC channel from a room. Users are only able to remove links for rooms they are a moderator in (power level of 50 or greater). Administrators of the bridge are able to remove links from any room.
 
 
 ### `!help`
--- a/src/bridge/AdminRoomHandler.ts
+++ b/src/bridge/AdminRoomHandler.ts
@@ -82,8 +82,7 @@ const COMMANDS: {[command: string]: {example: string; summary: string; requiresP
     },
     '!unlink': {
         example: `!unlink !room:example.com irc.example.net #foobar`,
-        summary: "Unlink an IRC channel from a Matrix room.",
-        requiresPermission: 'admin'
+        summary: "Unlink an IRC channel from a Matrix room. You need to be a Moderator of the Matrix room.",
     }
 };
 
--- a/src/bridge/AdminRoomHandler.ts
+++ b/src/bridge/AdminRoomHandler.ts
@@ -230,15 +230,6 @@ export class AdminRoomHandler {
         if (!ircChannel || !ircChannel.startsWith("#")) {
             return new MatrixAction("notice", "The channel name must start with a #");
         }
-        // Check if the room exists and the user is invited.
-        const intent = this.ircBridge.getAppServiceBridge().getIntent();
-        try {
-            await intent.getStateEvent(matrixRoomId, 'm.room.create');
-        }
-        catch (ex) {
-            log.error(`Could not join the target room of an !unlink command`, ex);
-            return new MatrixAction("notice", "Could not join the target room, you may need to invite the bot");
-        }
         try {
             await this.ircBridge.getProvisioner().unlink(
                 ProvisionRequest.createFake("adminCommand", log,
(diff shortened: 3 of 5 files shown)


src/bridge/AdminRoomHandler.ts
```diff
@@ -213,6 +221,47 @@ export class AdminRoomHandler {
         return new MatrixAction("notice", "Room plumbed.");
     }
 
+    private async handleUnlink(args: string[], sender: string, userPermission: string | undefined) {
+        if (userPermission !== 'admin') {
```
Half-Shot, line 225: This would block anyone who isn't a bridge admin, but we'd like the logic to be `(isbridgeAdmin || isPriviledgedInRoom)`.

src/bridge/AdminRoomHandler.ts
```diff
@@ -213,6 +221,47 @@ export class AdminRoomHandler {
...
+            return new MatrixAction("notice", "You must be an admin to use this command");
+        }
+        const [matrixRoomId, serverDomain, ircChannel] = args;
+        const server = serverDomain && this.ircBridge.getServer(serverDomain);
+        if (!server) {
+            return new MatrixAction("notice", "The server provided is not configured on this bridge");
+        }
+        if (!ircChannel || !ircChannel.startsWith("#")) {
```
Half-Shot, line 233: Kinda surprised we don't have a static function to validate ircChannel names yet, that'd be handy rather than using the startsWith snippets over and over.

src/provisioning/Provisioner.ts
```diff
@@ -843,8 +843,15 @@ export class Provisioner {
         }
     }
 
-    // Unlink an IRC channel from a matrix room ID
-    public async unlink(req: ProvisionRequest) {
+    /**
+     * Unlink an IRC channel from a matrix room ID
+     * @param req An Express Request which triggered the action. Its body should contain
```
Half-Shot, line 848: Ahhh, now technically it isn't an express object

Half-Shot (commented):
Looking good but there are some things we could improve here

src/bridge/AdminRoomHandler.ts
jaller94: :+1: I removed that requirement.

src/provisioning/Provisioner.ts
jaller94:
```suggestion
     * @param req An ExpressJS-Request-like object which triggered the action. Its body should contain
```

src/bridge/AdminRoomHandler.ts
```diff
@@ -213,6 +221,47 @@ export class AdminRoomHandler {
...
+            return new MatrixAction("notice", "The channel name must start with a #");
+        }
+        // Check if the room exists and the user is invited.
+        const intent = this.ircBridge.getAppServiceBridge().getIntent();
+        try {
+            await intent.getStateEvent(matrixRoomId, 'm.room.create');
+        }
+        catch (ex) {
+            log.error(`Could not join the target room of an !unlink command`, ex);
+            return new MatrixAction("notice", "Could not join the target room, you may need to invite the bot");
```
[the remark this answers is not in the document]
Half-Shot: I would detatch the requirement of being joined to a room to deleting it.

src/bridge/AdminRoomHandler.ts
```diff
@@ -213,6 +221,47 @@ export class AdminRoomHandler {
...
+        if (!ircChannel || !ircChannel.startsWith("#")) {
+            return new MatrixAction("notice", "The channel name must start with a #");
+        }
+        // Check if the room exists and the user is invited.
+        const intent = this.ircBridge.getAppServiceBridge().getIntent();
+        try {
+            await intent.getStateEvent(matrixRoomId, 'm.room.create');
+        }
+        catch (ex) {
+            log.error(`Could not join the target room of an !unlink command`, ex);
```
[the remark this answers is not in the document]
Half-Shot: I would drop this sanity check

docs/admin_room.md
```diff
@@ -110,6 +110,17 @@ aware of your actions.
 You must invite the bridge bot into the Matrix room for this to work.
 
 
+### `!unlink`
```
Half-Shot, line 113: :star2:  for docs!

changelog.d/1298.feature
```diff
@@ -0,0 +1 @@
+Added an !unlink command for Matrix users to unbridge two rooms
```
Half-Shot, line 1:
```suggestion
Added an !unlink command for Matrix users to unbridge a channel from Matrix```

Possibly?

File: docs/admin_room.md [hunk 8]
```diff
@@ -110,6 +110,17 @@ aware of your actions.
...
+
+`!unlink !room:example.com irc.network.net #channel`
+
+*This command only works for moderators of a bridged Matrix room and bridge administrators*
+
+This command allows you to unlink a IRC channel from a room without using the HTTP provisioning API. For bridge
+administrators this command does NOT validate that you have permission to do this on the IRC channel so please
+take care to ensure that the IRC channel is aware of your actions.
```
Half-Shot, line 121: I'd just describe what it does rather than mentioning the provis API at all.

```suggestion
This command allows you to unlink a IRC channel from a room. Users are only able to remove links for rooms they are a moderator in (power level of 50 or greater). Administrators of the bridge are able to remove links from any room.
```

src/bridge/AdminRoomHandler.ts
```diff
@@ -79,6 +79,11 @@ const COMMANDS: {[command: string]: {example: string; summary: string; requiresP
         example: `!plumb !room:example.com irc.example.net #foobar`,
         summary: "Plumb an IRC channel into a Matrix room.",
         requiresPermission: 'admin'
+    },
+    '!unlink': {
```
Half-Shot, line 83: Might be worth mentioning !unplumb, saying that it's just an alias for !unlink?

src/provisioning/Provisioner.ts
```diff
@@ -843,8 +843,15 @@ export class Provisioner {
...
-    // Unlink an IRC channel from a matrix room ID
-    public async unlink(req: ProvisionRequest) {
+    /**
+     * Unlink an IRC channel from a matrix room ID
+     * @param req An ExpressJS-Request-like object which triggered the action. Its body should contain
+     * the parameters for this unlink action.
+     * @param ignorePermissions If true, permissions are ignored (e.g. for bridge admins).
+     * Otherwise, the user needs to be a Moderator in the Matrix room.
+     */
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
Half-Shot, line 853: No any here?

src/provisioning/Provisioner.ts
```diff
@@ -877,40 +884,42 @@ export class Provisioner {
...
-            }
-            else if (e.type === "m.room.power_levels" && e.state_key === "") {
-                // https://matrix.org/docs/spec/client_server/r0.6.0#m-room-power-levels
-                let powerRequired = e.content.state_default || 50; // Can be empty. Assume 50 as per spec.
-                if (e.content.events && e.content.events["m.room.power_levels"]) {
-                    powerRequired = e.content.events["m.room.power_levels"];
+        if (!ignorePermissions) {
+            // Make sure the requester is a mod in the room
+            const botCli = this.ircBridge.getAppServiceBridge().getBot().getClient();
+            const stateEvents = await botCli.roomState(roomId);
```
Half-Shot, line 890: Not for now, but this is pants and horribly bad for performance. We should really just fetch the two state events we care about. Up to you if you fancy fixing this here or if we should make an issue.

Half-Shot: I also wonder if this should be part of Intent.

Half-Shot (approved):
A few nits here and there but the code is of good quality. Thank you!

src/provisioning/Provisioner.ts
jaller94:
```suggestion
```

src/provisioning/Provisioner.ts
jaller94: What two events?
If we were to fetch a state event where the person became a moderator, wouldn't we also need to check all ways they may have lost that permission?

PR #1299, opened 2021-04-26
Docker image: Upgrade to NodeJS 14

jaller94:
It's the current LTS.
https://nodejs.org/en/about/releases/

* Build and run the Docker images with NodeJS 14.
* Recommend running the bridge with NodeJS 14.
* Fix typo and lower case "docker"

Replaces #1249


Half-Shot (approved):
Seems reasonable

PR #1306, opened 2021-05-06
Build documentation when pushing to master

Half-Shot:
Fixes https://github.com/matrix-org/matrix-appservice-irc/issues/1292

~This annoyingly still requires us to update the `latest` symlink though on `gh-pages`.~ We deploy `latest` and `v$version` directories at the same time now, there is some file duplication here but it's easier to work with.


jaller94 (approved):
This is the first GitHub action config and run I have seen..
That being said, I think it looks ok and if something is wrong, it's not a big issue.

PR #1319, opened 2021-05-18
Update node-irc.js to make use of it's Typescript types

Half-Shot:
Depends upon https://github.com/matrix-org/node-irc/pull/59


benparsons (approved):
lgtm

PR #1329, opened 2021-05-24
Ensure invites are sent with is_direct

### LINKED ISSUE #1302: The bridge no longer marks invites to DM rooms as is_direct
Since #1290 we no longer invite users in the createRoom call but this means that the seperate invite call doesn't set the is_direct state either. I assume we might be able to manually send the state into the room to fix this, maybe.

https://github.com/matrix-org/matrix-appservice-irc/blob/9222ff565a12a7d3054cfa0184106a01df40ac0a/src/bridge/IrcHandler.ts#L135

Half-Shot:
Fixes #1302

### DIFF
--- a/changelog.d/1329.bugfix
+++ b/changelog.d/1329.bugfix
@@ -1,0 +1,1 @@
+Fix an issue where invites to DM rooms are not marked as direct message invites.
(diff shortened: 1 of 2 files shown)


jaller94 (approved):
Nice!

PR #1330, opened 2021-05-24
Fix nickname validation bug

### LINKED ISSUE #1313: !irc nick fails when no argument is provided
Instead of saying that the argument is missing, it shows this error
> Unable to change nick: Cannot read property 'replace' of undefined

Half-Shot:
Fixes #1313

### DIFF
--- a/src/bridge/MatrixHandler.ts
+++ b/src/bridge/MatrixHandler.ts
@@ -811,7 +811,7 @@ export class MatrixHandler {
         const [targetRoom] = await this.ircBridge.getStore().getIrcChannelsForRoomId(event.room_id);
         if (command === "nick") {
             const newNick = args[0];
-            if (newNick?.length > 0) {
+            if (!(newNick?.length > 0)) {
                 await intent.sendMessage(event.room_id, {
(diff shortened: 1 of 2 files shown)


src/bridge/MatrixHandler.ts
```diff
@@ -811,12 +811,20 @@ export class MatrixHandler {
         const [targetRoom] = await this.ircBridge.getStore().getIrcChannelsForRoomId(event.room_id);
         if (command === "nick") {
             const newNick = args[0];
+            if (newNick?.length > 0) {
```
jaller94, line 814: I think this does the opposite of what it should do, rejecting all newNicks with a length greater 0.
```suggestion
            if (!(newNick?.length > 0)) {
```

jaller94 (changes requested):

src/bridge/MatrixHandler.ts
Half-Shot: Doh!

jaller94 (approved):
Looks good to me.

Issue #1253, opened 2021-03-01
Deadlock when connecting to inspircd servers with SASL


thomwiggers:
https://github.com/matrix-org/node-irc/pull/52

As evidenced in the transcript below, a server doesn't necessarily send back `*` if you send `CAP REQ` before `NICK` (observed with inspircd / Atheme).

This seems to deadlock the connection.

```
:irc.server NOTICE Auth :*** Looking up your hostname...                                                                                                  
:irc.server NOTICE Auth :*** Could not resolve your hostname: Domain name not found; using your IP address 2a02:X) instead.                                                                                                                                                           
CAP REQ sasl                                                                                                                                                  
:irc.server CAP 663AAB8QI ACK :sasl                                                                                                                       
```

(sorry, an earlier version of this message was a copy/paste fail)

PR #1331, opened 2021-05-24
Add support for setting a username for users

Half-Shot:
Additionally, this also:

- Adds metadata handlers for various sasl loggedin/loggedout/error events (requires https://github.com/matrix-org/node-irc/pull/62)
- Adds a `!reconnect` command to be used after setting a username / password.
- Stops `!storepass` from automatically reconnecting you, instead prompting you to reconnect.

This should hopefully improve the process wrt authenticating with services over the bridge, esepcially freenode and libera.chat which would prefer us to use SASL PLAIN over the PASS command.

### DIFF
--- a/src/bridge/AdminRoomHandler.ts
+++ b/src/bridge/AdminRoomHandler.ts
@@ -493,8 +493,7 @@ export class AdminRoomHandler {
                 config.setUsername(username);
                 await this.ircBridge.getStore().storeIrcClientConfig(config);
                 notice = new MatrixAction(
-                    "notice", `Successfully stored username for ${domain} for future connections. `
-                    + "Use !reconnect to use this username now."
+                    "notice", `Successfully stored username for ${domain}. Use !reconnect to use this username now.`
                 );
             }
         }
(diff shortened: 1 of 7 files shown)


src/bridge/AdminRoomHandler.ts
```diff
@@ -420,8 +435,81 @@ export class AdminRoomHandler {
...
+
+    private async handleUsername(req: BridgeRequest, args: string[], userId: string, server: IrcServer) {
+        const domain = server.domain;
+        const store = this.ircBridge.getStore();
+        let notice;
+
+        try {
+            // Allow passwords with spaces
+            const username = args[0]?.trim();
+            if (!username || username.length === 0) {
```
jaller94, line 467: Empty strings are falsy.
```suggestion
            if (!username) {
```

src/bridge/AdminRoomHandler.ts
```diff
@@ -420,8 +435,81 @@ export class AdminRoomHandler {
...
+                let config = await store.getIrcClientConfig(userId, server.domain);
+                if (!config) {
+                    config = IrcClientConfig.newConfig(
+                        new MatrixUser(userId), server.domain
+                    );
+                }
+                config.setUsername(username);
+                await this.ircBridge.getStore().storeIrcClientConfig(config);
+                notice = new MatrixAction(
+                    "notice", `Successfully stored username for ${domain} for future connections.`
```
jaller94, line 496: On lines similar to this, you added a separating space.
```suggestion
                    "notice", `Successfully stored username for ${domain} for future connections. `
```

jaller94 (approved):
Looks good to me.

thomwiggers:
The update to node-irc should fix #1253 as well

PR #1337, opened 2021-05-26
Block IRC messages appearing in Matrix rooms if Matrix user(s) cannot be joined.

### LINKED ISSUE #1332: Safety net feature to block bridging new messages to Matrix until all users are joined to the channel
Typically channels should have this implicitly as Matrix users are usually kicked if they cannot join a channel however we would like to add an optional safety net to preserve privacy. A Matrix room will not get messages from IRC while the membership for a room contains users who are not connected to IRC.

Basically, while we sync membership to IRC we will not bridge messages from IRC until all the IRC clients are joined to the network and channel. If a user fails to join either the network or a channel, they are kicked from the room. If the kick is not possible (after retries) then the bridge will warn in the room that the privacy could not be preserved and those users will be kicked.

Finally, this would be a feature that the IRC opers could opt out of if they wish by signalling to the bridge somehow that they do not wish to block messages based upon this (e.g. they've chosen to not bridge new matrix users on join for whatever reason).

Half-Shot:
Fixes #1332 

This feature will block IRC messages from appearing across the bridge if a Matrix user isn't connected to the IRC side. It's considered a tool of last resort by the bridge to ensure no history is leaked, as in theory the bridge should be vigilantly kicking users who cannot get connected.

This setting can be enforced on or off at a global level, and at a per-channel level. Optionally, bridge admins can also allow moderators of Matrix rooms to disable this setting at their own risk.

### DIFF
--- a/src/bridge/IrcHandler.ts
+++ b/src/bridge/IrcHandler.ts
@@ -449,6 +449,56 @@ export class IrcHandler {
     }
 
     /**
+     * If configured, check to see if the all Matrix users in a given room are
+     * joined to a channel. If they are not, drop the message.
+     * @param req The IRC request
+     * @param server The IRC server.
+     */
+    private async requiresAllMatrixUsersJoined(server: IrcServer, channel: string, roomId: string): Promise<boolean> {
+        // The room state takes priority.
+        const stateRequires =
+            await this.ircBridge.roomConfigs.allowUnconnectedMatrixUsers(roomId, new IrcRoom(server, channel));
+        if (stateRequires !== null) {
+            console.log("fooy", stateRequires);
+            return stateRequires;
+        }
+        return server.shouldRequireMatrixUserJoined(channel);
+    }
+
+    /**
+     * See if every joined Matrix user is also joined to the IRC channel. If they are not,
+     * this returns false. A seperate mechanism should be use to join the user if this fails.
+     * @param req The IRC request
+     * @param server The IRC server.
+     * @param channel The IRC channel.
+     * @param roomId The Matrix room.
+     * @returns True if all users are connected and joined, or false otherwise.
+     */
+    private async areAllMatrixUsersJoined(
+        req: BridgeRequest, server: IrcServer, channel: string, roomId: string): Promise<boolean> {
+        // Look for all the Matrix users in the room.
+        const members = await this.ircBridge.getMatrixUsersForRoom(roomId);
+        const pool = this.ircBridge.getClientPool();
+        for (const userId of members) {
+            if (userId === this.ircBridge.appServiceUserId) {
+                continue;
+            }
+            const client = pool.getBridgedClientByUserId(server, userId);
+            if (!client) {
+                req.log.warn(`${userId} has not connected to IRC yet, not bridging message`);
+                return false;
+            }
+            if (!client.inChannel(channel)) {
+                // TODO: Should we poke them into joining?
+                req.log.warn(`${userId} has not joined the channel yet, not bridging message`);
+                return false;
+            }
+        }
+        return true;
+    }
+
+
+    /**
      * Called when the AS receives an IRC topic event.
      * @param {IrcServer} server The sending IRC server.
      * @param {IrcUser} fromUser The sender.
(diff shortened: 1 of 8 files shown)


src/irc/IrcServer.ts
```diff
@@ -502,6 +504,21 @@ export class IrcServer {
         return shouldSync;
     }
 
+    /**
+     * Does the server/channel require all Matrix users to be joined.
+     * @param channel The IRC channel.
+     * @returns If the server requires all Matrix users to be joined.
```
jaller94, line 510:
```suggestion
     * Does the server/channel require all Matrix users to be joined?
     * @param channel The IRC channel.
     * @returns True if the server requires all Matrix users to be joined.
```

src/bridge/IrcHandler.ts
```diff
@@ -448,6 +448,55 @@ export class IrcHandler {
...
+            if (userId === this.ircBridge.appServiceUserId) {
+                continue;
+            }
+            const client = pool.getBridgedClientByUserId(server, userId);
+            if (!client) {
+                req.log.warn(`${userId} has not connected to IRC yet, not bridging message`);
+                return false;
+            }
+            if (!client.inChannel(channel)) {
+                // TODO: Should we poke them into joining?
```
jaller94, line 491: Poke them? As in "sending them a message" that they need to join?

I guess the Matrix users in the room might be interested as to who is the problematic user that needs to be joined or kicked.

src/bridge/IrcHandler.ts
```diff
@@ -448,6 +448,55 @@ export class IrcHandler {
         }
     }
 
+    /**
+     * If configured, check to see if the all Matrix users in a given room are
+     * joined to a channel. If they are not, drop the message.
+     * @param req The IRC request
+     * @param server The IRC server.
+     */
+    private async requiresAllMatrixUsersJoined(server: IrcServer, channel: string, roomId: string): Promise<boolean> {
```
jaller94, line 457: Why sometimes `requiresAllMatrixUsersJoined` and other times `shouldRequireMatrixUserJoined`?
If they are on different classes but behave the same I'd call them the same.

jaller94 (approved):
This adds so much complexity right in the heart of the bridge handlers 😩
Nice PR though and it looks to be all correctly implemented. :+1:

src/bridge/IrcHandler.ts
```diff
@@ -448,6 +448,55 @@ export class IrcHandler {
         }
```
Half-Shot, line 448: @ara4n mentioned that we should add some state event when the room is unbridged, and another for when it comes back to allow clients to show the disconnect and for us to debug it.

src/bridge/IrcHandler.ts
Half-Shot: More that we should just invoke a join, as this means the Matrix user is in the room but the bridge hasn't joined them.

src/bridge/IrcHandler.ts
Half-Shot: Renamed to be the same.

src/irc/IrcServer.ts
```diff
@@ -502,6 +504,21 @@ export class IrcServer {
...
 
+    /**
+     * Does the server/channel require all Matrix users to be joined?
+     * @param channel The IRC channel.
+     * @returns True if the server requires all Matrix users to be joined.
+     */
+    public shouldRequireMatrixUserJoined(channel: string) {
+        let shouldSync = this.config.membershipLists.global.ircToMatrix.requireMatrixJoined;
+        this.config.membershipLists.channels.forEach((chan) => {
+            if (chan.channel === channel && typeof chan.ircToMatrix?.requireMatrixJoined === "boolean") {
```
andir, line 515: Isn't this a bit expensive? It iterates through all the channels even if the channel has already been found.

Half-Shot: Good catch.

src/bridge/RoomConfig.ts
```diff
@@ -89,4 +91,23 @@ export class RoomConfig {
         }
         return Math.min(roomState.lineLimit, this.config?.lineLimitMax ?? roomState.lineLimit);
     }
+
+    /**
+     * Get the per-room configuration for the paste bin limit for a room.
```
ara4n, line 96: bogus comment

ara4n:
this generally lgtm from a quick skim, although I'm worried that i've never seen myself kicked from a bridged room when my IRC connection to the remote server drops.  So while this drops msgs while the server is reconnecting its clients, I'm worried we still have a hole where if my IRC client pings out, the Matrix client doesn't get soft-kicked from the room.

jasikpark:
> this generally lgtm from a quick skim, although I'm worried that i've never seen myself kicked from a bridged room when my IRC connection to the remote server drops. So while this drops msgs while the server is reconnecting its clients, I'm worried we still have a hole where if my IRC client pings out, the Matrix client doesn't get soft-kicked from the room.

#libera:libera.chat and #libera-matrix:libera.chat both seem to be affected by something like this - large rooms seem to constantly have at least one user trying to join, and so irc -> matrix messages rarely come through.

The irc<->matrix libera bridge is working great outside of those two channels for others and I.

Issue #186, opened 2016-09-24
Allow 1:1 channel name mapping


mweinelt:
I am trying to privde a matrix irc bridge for hackint and subsequently use `irc.hackint.org` as its domain. The proper channelnames in my opinion would be `#channel:irc.hackint.org` but that is currently impossible to achieve.

What I can get is `##channel:.irc.hackint.org` but that is very counterintuitive and it doesn't looky very pleasant tbh.

Kegsay:
This breaks because channels can start with characters other than `#` (notably `&`), so you need a double `#` somewhere.

mweinelt:
Indeed `&channels` exist, but they're usually local to a single ircd and as such, at least in my opinion, rarely used and of less relevance.

mweinelt:
Fwiw, we achieved this a long time ago by patching the schema like this

```diff
diff --git a/lib/config/schema.yml b/lib/config/schema.yml
index 4ca77f9..b7c5fc7 100644
--- a/lib/config/schema.yml
+++ b/lib/config/schema.yml
@@ -240,7 +240,7 @@ properties:
                                     type: "boolean"
                                 aliasTemplate:
                                     type: "string"
-                                    pattern: "^#.*\\$CHANNEL"
+                                    pattern: "^.*\\$CHANNEL"
                                 whitelist:
                                     type: "array"
                                     items:
```

With this change you can set 

```
        # The room alias template to apply when creating new aliases. This only
        # applies if createAlias is 'true'. The following variables are exposed:
        # $SERVER => The IRC server address (e.g. "irc.example.com")
        # $CHANNEL => The IRC channel (e.g. "#python")
        # This MUST have $CHANNEL somewhere in it.
        # Default: '#irc_$SERVER_$CHANNEL'
        aliasTemplate: "$CHANNEL"
```

Kegsay:
The bridge code will likely assume that if a string passes the `aliasTemplate` check that it starts with `#`. If this is not true for any reason, it's likely to cause it to crash, which would be disruptive for the entire bridge. I'm happy that your monkey-patch works, though I'd be even happier if you PR'd a more comprehensive fix so everyone can benefit :)

PR #1339, opened 2021-05-26
Allow aliases that start with a single #

### LINKED ISSUE #1338: Bridge requires that channel aliases begin with a redundant hash.
E.g. Your alias must be at least `#$CHANNEL` which looks like `##foo:matrix.org` for channel `#foo` rather than a more desirable `#foo:matrix.org`

Half-Shot:
Fixes #1338 

Allows us to do #foo:matrix.org rather than ##foo:matrix.org


config.schema.yml
```diff
@@ -304,7 +304,7 @@ properties:
                                     type: "string"
                                 aliasTemplate:
                                     type: "string"
-                                    pattern: "^#.*\\$CHANNEL"
+                                    pattern: "^(#.*)?\\$CHANNEL"
```
jaller94, line 307: * The pattern does not include the mandatory hash sign from a Matrix room ID?
* Why do we need the optional group? Can $CHANNEL not start with a hash sign?

mweinelt:
https://github.com/matrix-org/matrix-appservice-irc/issues/186

config.schema.yml
Half-Shot: Hmm, I guess it may come up. Will rethink this one

Half-Shot: Perhaps the solution is to have `($HASH_CHANNEL|#$CHANNEL)`.

Half-Shot: Practically speaking this appears to work. 

>The pattern does not include the mandatory hash sign from a Matrix room ID?

The $CHANNEL starts with a hash, so in the case of #matrix:libera.chat, the hash prefix would make this ##matrix:libera.chat.

>Why do we need the optional group? Can $CHANNEL not start with a hash sign?

Because it's valid to have #matrix, ##_irc_#matrix and ##matrix as your localpart, but never foo_#matrix.

mweinelt:
Sidequest: Is this cleaner form also possible for users? Especially when the homeserver is dedicated to bridging IRC and has no local users.

jaller94 (approved):
Implying you accept the suggestions, this looks good to me.

src/irc/IrcServer.ts
```diff
@@ -769,6 +776,16 @@ export class IrcServer {
...
         }
 
+        if (!this.aliasTemplateHasHashPrefix) {
+            if (this.config.dynamicChannels.aliasTemplate !== "$CHANNEL") {
+                throw Error(
+                    "If no hash prefix is given in 'aliasTemplate', then the aliasTemplate must be exactly '$CHANNEL'"
+                );
+            }
+            log.warn(`You have configured your aliasTemplate to not include a prefix hash. This means that only
+            channels starting with a hash are supported by the bridge.`)
```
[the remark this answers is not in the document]
Half-Shot: ah, good catch.

Half-Shot:
@mweinelt 

>Sidequest: Is this cleaner form also possible for users (@$USER:example.com)? Especially when the homeserver is dedicated to bridging IRC and has no local users.

That's already possible today :)

PR #1344, opened 2021-05-27
Add support for using the homeserver directory for bridged rooms

Half-Shot:
Fixes an issue where homeservers like libera.chat do not display published bridged rooms.


Half-Shot:
In practice it appears to work, merging.

PR #1356, opened 2021-05-28
Sync membership after plumbing a room

Half-Shot:


jluebbe:
It seems that this isn't yet fixed for all cases. Messages from #oe:matrix.org are shown in #oe on libera.chat, but not the other way.

PR #1366, opened 2021-06-04
Support zero width spaces

### LINKED ISSUE #1364: matrix-appservice-irc strips ZWSP characters leading extraneous pings
**Describe the bug**

Recent events have increased the popularity of relaybots and as some people tend to be on multiple sides of them, there is a defacto behaviour of adding a ZWSP as a second character of the sender nickname. For some reason `matrix-appservice-irc` removes that and causes me to get pinged whenever I type in some rooms, while that doesn't happen with IRC clients.

**To Reproduce**

1. Open https://testnet.ergo.chat/kiwi/#limnoria-bots with a nickname-that-pings-you-at-Matrix and `/join #limnoria-bots:libera.chat` on Matrix.
2. Say something on the first link.
3. You get pinged on Matrix regardless of there being a ZWSP.

Alternatively you can use https://ergo.chat/kiwi/#ergo / https://ergo.chat/gamja/#ergo and join `#ergo:libera.chat`, but please don't as that is an on-topic support channel while the first one is a bot testing channel that hopefully doesn't mind a bit of other testing too.

**Expected behavior**

I expect matrix-appservice-irc to not remove ZWSP characters and thus not get pinged when there is a ZWSP breaking my nicknames.

**Screenshots**

* IRC
  * Note how here only the last message is considered as a ping by WeeChat.
* Element
   * Messages 1 & 3 don't contain `Mikaela` they contain `M<ZWSP>ikaela` while we cannot see the ZWSP.

 - OS: Fedora 34
 - Browser: Firefox
 - Version: 88.0.1

**Additional context**

* [Zero-width space on Wikipedia](https://en.wikipedia.org/wiki/Zero-width_space)

Half-Shot:
Fixes #1364 

We previously filtered these out (https://github.com/matrix-org/matrix-appservice-irc/commit/72f9243025a4b3cacfbbec26c1035e0a570b82ed) but I think this was just to avoid breaking HTML formatting.

### DIFF
--- a/spec/unit/formatting.spec.js
+++ b/spec/unit/formatting.spec.js
@@ -51,6 +51,12 @@ describe("Formatting", function() {
             ).toBe("The quick brown <b>fox</b> jumps over the lazy <b>dog</b>.");
         });
     });
+    describe("stripIrcFormatting", function() {
+        it("should not strip ZWSP characters", () => {
+            const text = "Lorem​Ipsum​Dolor​Sit​";
+            expect(formatting.stripIrcFormatting(text)).withContext("ZWSPs missing").toBe(text);
+        });
+    });
     describe("markdownCodeToIrc", function() {
         it("should return null for a non-code input", function() {
             expect(
--- a/spec/unit/formatting.spec.js
+++ b/spec/unit/formatting.spec.js
@@ -53,7 +53,7 @@ describe("Formatting", function() {
     });
     describe("stripIrcFormatting", function() {
         it("should not strip ZWSP characters", () => {
-            const text = "Lorem​Ipsum​Dolor​Sit​";
+            const text = "Lorem\u200bIpsum\u200bDolor\u200bSit";
             expect(formatting.stripIrcFormatting(text)).withContext("ZWSPs missing").toBe(text);
         });
     });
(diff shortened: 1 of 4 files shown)


spec/unit/formatting.spec.js
```diff
@@ -51,6 +51,12 @@ describe("Formatting", function() {
             ).toBe("The quick brown <b>fox</b> jumps over the lazy <b>dog</b>.");
         });
     });
+    describe("stripIrcFormatting", function() {
+        it("should not strip ZWSP characters", () => {
```
jaller94, line 55: If there's a Zero Width Space in here, would you mind printing it as \u200B?
Otherwise I think it will get lost and a massive pain for the next person to figure out the `git diff`.

jaller94 (approved):
Please don't include the actual character in code.
Otherwise, looks good to me.

spec/unit/formatting.spec.js
Half-Shot: Oh, right yes. That's an obvious one :facepalm:

Mikaela:
screenshot

I think the might be a small problem with this pull request 😅

Issue #1359, opened 2021-05-29
Can't store username that contains underscore


loganrosen:
**Describe the bug**
I tried to store my username on Libera Chat (Logan_) for SASL authentication using `!username Logan_`, but I got the message that it contains invalid characters not supported by IRC:
```
!username Logan_

Username contained invalid characters not supported by IRC.
```

This message is definitely incorrect because my username is registered with Libera Chat with an underscore. It looks like [this is because `_` is used as a delimiter by the bridge](https://github.com/matrix-org/matrix-appservice-irc/blob/develop/src/irc/IdentGenerator.ts#L226). Does this mean I can't authenticate my Libera username using Matrix?

**To Reproduce**
1. Open a DM with the bridge user
2. Say `!username` and then a username containing an underscore
3. See error

**Expected behavior**
I expected to be able to store my IRC username with Matrix and automatically authenticate using SASL.

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Browser Firefox
 - Version 90.0a1

vberger:
It's maybe not related with the underscore, because I get the same error message with `!username Levans` (which is registered by me).

tim-seoss:
Upper case characters are (incorrectly) rejected apparently.

deepbluev7:
I have the same issue with "Nico[m]", presumably also because of the upper case character.

pwr22:
I hit both of these cases with `Peter_R` :laughing:

jonasbb:
I have the same problem with `|`. I have no problem setting my nick to `saarsec|jbushart` using `!nick`, but `!username` fails with "Username contained invalid characters not supported by IRC.".

4nd3r:
not all characters are equal. another example: #498

Mikaela:
Another incorrectly invalid character seems to be `@`. It's needed for specifying device ID for Ergo IRCd and possibly also ZNC.

* https://github.com/ergochat/ergo/blob/v2.7.0/docs/USERGUIDE.md#history (point 3. 1 & 2, point 3 is #1410.

emersonveenstra:
If you're trying to use !username for SASL, a workaround for Libera (and other networks with a standard services setup) is you can group a nick that satisfies the requirements of the bridge (so basically all lowercase letters) and use that as !username

llucax:
That seems to work for me, not ideal but at least is a workaround. Thanks!

progval:
[...]

PR #1367, opened 2021-06-04
Support trailing underscores in usernames

babolivier:
Since this is my first PR on this project, please let me know if this isn't the right way to do it, or if I've forgotten anything.

Fixes https://github.com/matrix-org/matrix-appservice-irc/issues/1359

### DIFF
--- a/src/bridge/AdminRoomHandler.ts
+++ b/src/bridge/AdminRoomHandler.ts
@@ -465,8 +465,10 @@ export class AdminRoomHandler {
         let notice;
 
         try {
-            // Allow passwords with spaces
+            // Allow usernames with spaces
             const username = args[0]?.trim();
+            // Allow usernames with trailing underscores
+            const usernameWithoutUnderscores = username.replace(/_/g, "");
             if (!username) {
                 notice = new MatrixAction(
                     "notice",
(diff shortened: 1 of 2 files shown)


jaller94:
Isn't it that underscores shouldn't get sanitised by `sanitiseUsername` if they don't fit the delimiter pattern of the bridge?

src/bridge/AdminRoomHandler.ts
```diff
@@ -480,7 +482,7 @@ export class AdminRoomHandler {
                     `Username is longer than the maximum permitted by the bridge (${SANE_USERNAME_LENGTH}).`
                 );
             }
-            else if (IdentGenerator.sanitiseUsername(username) !== username) {
+            else if (IdentGenerator.sanitiseUsername(username) !== usernameWithoutUnderscores) {
```
Half-Shot, line 485: I think the best approach might be to make santitiseUsername more lax to be honest. The linked issue seems to mention that more characters are supported than I expected, so perhaps just make that accept more?

Half-Shot (commented):
Thanks for the PR! I think the approach would fix this specific issue, but I'd like to improve our sanitise function :)

babolivier:
Superceded by https://github.com/matrix-org/matrix-appservice-irc/pull/1538

Half-Shot:
And now superseded by https://github.com/matrix-org/matrix-appservice-irc/pull/1719

PR #1369, opened 2021-06-04
Hs/improve blocking

Half-Shot:
This change supersedes #1368.

This PR merges the useful changes made to libera back into the `develop` branch (with the exception of #1339, #1356 which should be merged separately).

This PR adds:
- A metric for tracking which channels are blocked
- Kicking users from rooms when they fail to get connected to a IRC channel.
- Fixes a bug where only the first error would be caught for a IRC join failure, which could cause the join to hang indefinitely.
- Moves the privacy safety net into it's own class
- Adds caching of the member list to speed up slightly
- Handles `err_useronchannel` where we might try to join a channel and be told we are already in it.

### DIFF
--- a/src/irc/PrivacyProtection.ts
+++ b/src/irc/PrivacyProtection.ts
@@ -1,0 +1,155 @@
+import { MatrixRoom } from "matrix-appservice-bridge";
+import { IrcBridge } from "../bridge/IrcBridge";
+import { BridgeRequest } from "../models/BridgeRequest";
+import { IrcRoom } from "../models/IrcRoom";
+import { IrcServer } from "./IrcServer";
+
+/**
+ * This class manages the visiblity of IRC messages on Matrix. It will check upon each IRC message
+ * that all Matrix users are connected to the channel to avoid messages leaking to the Matrix side.
+ *
+ */
+export class PrivacyProtection {
+    private roomBlockedSet = new Set<string>();
+    private memberListCache = new Map<string, string[]>();
+    constructor(private ircBridge: IrcBridge) {
+
+    }
+
+    /**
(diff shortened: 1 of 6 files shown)


src/irc/PrivacyProtection.ts
```diff
@@ -0,0 +1,161 @@
...
+        const allRooms = await this.ircBridge.getStore().getMatrixRoomsForChannel(server, channel);
+        return (await Promise.all((allRooms).map(async (room) => {
+            const required = await this.shouldRequireMatrixUserJoined(server, channel, room.roomId);
+            req.log.debug(`${room.roomId} ${required ? "requires" : "does not require"} Matrix users to be joined`);
+            if (!required) {
+                return room;
+            }
+            const allowed = await this.areAllMatrixUsersJoined(req, server, channel, room.roomId);
+            // Do so asynchronously, as we don't want to block message handling on this.
+            this.setBlockedStateInRoom(req, room.roomId, new IrcRoom(server, channel), !allowed);
```
jaller94, line 157: While the function says it's always going to resolve.. there could be a coding error resulting in an unhandled promise.

To log potential errors, implying one can still log to `req.log` after a response has been sent:
```suggestion
            this.setBlockedStateInRoom(req, room.roomId, new IrcRoom(server, channel), !allowed).catch(req.log.error);
```

jaller94 (approved):
Looks good to me. Pretty complicated change to be honest.

src/irc/PrivacyProtection.ts
Half-Shot: Yep, seems fine to me.

PR #1371, opened 2021-06-07
Update wording

adrian5:
There was a typo right at the top, and the wording was a bit long winded. The section below already explains what the Admin Room does.

### DIFF
--- a/docs/admin_room.md
+++ b/docs/admin_room.md
@@ -1,12 +1,12 @@
 # Admin Room
 
-Most bridges will send you an invite to an admin room once you connectto control your connection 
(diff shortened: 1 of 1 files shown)


Half-Shot (approved):
Looks great, can I get you to do a changelog file & sign off as per https://github.com/matrix-org/matrix-appservice-irc/blob/develop/CONTRIBUTING.md#making-pull-requests ?

adrian5:
Ah sorry, I missed that. Some projects require signage, while others don't, so I don't expect it by default. My account doesn't even have a valid email address. I'll get around to that one day, but not right now, so I shall close this PR. Feel free to replicate the changes, or just the typo part.

PR #1372, opened 2021-06-07
Ensure admin room is created with the is_direct flag

### LINKED ISSUE #1370: "liberachat IRC Bridge status" is created as a room instead of a DM.
**Describe the bug**
Upon first join of a bridged channel, appservice invites the user to a room instead of a DM for "liberachat IRC bridge status".

**To Reproduce**
1. Join a bridged libera.chat IRC channel from Matrix for the first time.
2. Receive room invite from appservice.
3. Join Bridge status room.
4. Develop awareness that it is a room and not a DM.

**Expected behavior**
Seems like it would be more appropriate as a DM. Quoting Half-Shot from the bridge's IRC channel: "I believe the fact that it doesn't [categorize the room as a DM] is a bug."

**Smartphone (please complete the following information):**
 - Device: Ulefone Armor X7 Pro
 - OS: Android 10
 - Browser: (Does my browser matter in this instance? Matrix client instead:) Element Android
 - Version 1.1.8

**Additional context**
Using the Element web client, it is possible to issue a `/converttodm` command in the existing room in order to categorize it as a DM on other platforms of Element. I haven't been around Matrix long enough to really understand where this change takes effect, whether it's Element-side or Matrix-side.

Half-Shot:
Fixes #1370 

This:
- Marks the admin room as a DM
- Linearlises the admin room creation code to avoid duplicates

### DIFF
--- a/src/bridge/IrcHandler.ts
+++ b/src/bridge/IrcHandler.ts
@@ -1062,12 +1062,13 @@ export class IrcHandler {
                                     `letting you control the connection. Type !help for more information`
                     const notice = new MatrixAction("notice", newRoomMsg);
                     await this.ircBridge.sendMatrixAction(adminRoom, botUser, notice);
+                    // This is stored now so we can delete the promise.
+                    this.pendingAdminRooms.delete(userId);
                     return adminRoom;
                 })();
                 this.pendingAdminRooms.set(client.userId, adminRoomNewPromise);
                 adminRoom = await adminRoomNewPromise;
             }
-
         }
         else {
             adminRoom = fetchedAdminRoom;
--- a/src/bridge/IrcHandler.ts
+++ b/src/bridge/IrcHandler.ts
@@ -1059,7 +1059,7 @@ export class IrcHandler {
                                     `'${client.server.domain}', where you ` +
                                     `are now connected as ${client.nick}. ` +
                                     `This room shows any errors or status messages from IRC, as well as ` +
-                                    `letting you control the connection. Type !help for more information`
+                                    `letting you control the connection. Type !help for more information.`
                     const notice = new MatrixAction("notice", newRoomMsg);
                     await this.ircBridge.sendMatrixAction(adminRoom, botUser, notice);
                     // This is stored now so we can delete the promise.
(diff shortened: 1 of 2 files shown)


src/bridge/IrcHandler.ts
```diff
@@ -1014,45 +1017,58 @@ export class IrcHandler {
...
+                            invite: [userId]
+                        }
+                    });
+                    adminRoom = new MatrixRoom(response.room_id);
+                    await this.ircBridge.getStore().storeAdminRoom(adminRoom, userId);
+                    const newRoomMsg = `You've joined a Matrix room which is bridged to the IRC network ` +
+                                    `'${client.server.domain}', where you ` +
+                                    `are now connected as ${client.nick}. ` +
+                                    `This room shows any errors or status messages from IRC, as well as ` +
+                                    `letting you control the connection. Type !help for more information`
```
jaller94, line 1062:
```suggestion
                                    `letting you control the connection. Type !help for more information.`
```

src/bridge/IrcHandler.ts
```diff
@@ -1014,45 +1017,58 @@ export class IrcHandler {
...
+            else {
+                const adminRoomNewPromise = (async () => {
+                    req.log.info("Creating an admin room with %s", userId);
+                    const response = await this.ircBridge.getAppServiceBridge().getIntent().createRoom({
+                        createAsClient: false,
+                        options: {
+                            name: `${client.server.getReadableName()} IRC Bridge status`,
+                            topic:  `This room shows any errors or status messages from ` +
+                                    `${client.server.domain}, as well as letting you control ` +
+                                    "the connection. ",
```
jaller94, line 1049:
```suggestion
                                    "the connection.",
```

src/bridge/IrcHandler.ts
```diff
@@ -1003,6 +1005,7 @@ export class IrcHandler {
             // Probably the bot
             return undefined;
         }
+        const userId = client.userId;
```
jaller94, line 1008: I assume this is just for readability, and not because `client` will ever change. Or do we worry about it getting modified externally during an await?

jaller94 (approved):
Looks good to me.

src/bridge/IrcHandler.ts
Half-Shot: Typescript didn't like using client.userId, because it can be undefined and we had a arrow function further down which needed it to be defined.

PR #1373, opened 2021-06-07
Do not attempt to sync membership from a room we can't access

Half-Shot:
If the bridge cannot join a room, it will repeatedly fail to sync the room and cause the membership sync to fail. We should just ignore these in the membership sync stage.


jaller94 (approved):
Looks good to me. Good to not delay these, if we know that we don't have the permission.

Issue #1095, opened 2020-07-28
IRC username and IRC Realname


H7-25:
Should be fine can set a different template both for irc username and IRC Realname

PR #1374, opened 2021-06-08
Allow templating realname

3nprob:
Addresses #1095


config.schema.yml
```diff
@@ -379,7 +379,7 @@ properties:
                                     type: "boolean"
                                 realnameFormat:
                                     type: "string"
-                                    enum: ["mxid","reverse-mxid"]
+                                    pattern: "^template:(\\$USERID|\\$LOCALPART|\\$DISPLAY|\\$IRCUSER)|^mxid$|reverse-mxid$"
```
jaller94, line 382: The current pattern only allows four templates.
The value in your tests (`template:$IRCUSER_$DISPLAY_$LOCALPART_$USERID_suffix`) would not be allowed.

```suggestion
                                    pattern: "^template:|^mxid$|reverse-mxid$"
```

jaller94 (changes requested):
I think there's an issue with the template validation.

config.schema.yml
3nprob: Good catch, thanks!

config.schema.yml
```diff
@@ -379,7 +379,7 @@ properties:
                                     type: "boolean"
                                 realnameFormat:
                                     type: "string"
-                                    enum: ["mxid","reverse-mxid"]
+                                    pattern: "^template:|^mxid$|reverse-mxid$"
```
Half-Shot, line 382: I think that regex wouldn't match `template:TEMPLATE`.

config.sample.yaml
```diff
@@ -387,7 +387,9 @@ ircService:
         # through the bridge e.g. caller ID as there is no way to /ACCEPT.
         # Default: "" (no user modes)
         # userModes: "R"
-        # The format of the realname defined for users, either mxid or reverse-mxid
+        # The format of the realname defined for users.
+        # either "mxid", "reverse-mxid", or "template:TEMPLATE", where TEMPLATE must
+        # contain either of $USERID, $LOCALPART, $DISPLAY, $IRCUSER
```
Half-Shot, line 392: Is template limited to strictly "$USERID", "$LOCALPART", "$DISPLAY", "$IRCUSER" or can it contain prefix/suffix characters, as the help text isn't clear?

I think on the whole I would make `mxid` and `reverse-mxid` available as templating options inline with what you have there already (mxid is essentially $USERID) and then drop the whole `template:` thing entirely, so the setting looks very similar to `nickTemplate` in the end.

config.schema.yml
3nprob: it does actually; after the review suggestion from　@jaller94 it should now match anything that starts with `template:`

config.sample.yaml
3nprob: Agreed; if I did this from scratch that's exactly how I would do it as well.

The reason for doing it this way is to keep backwards compatibility so users who have already configured this will keep having it working as previously. Though naturally this is at the expense of added complexity.

What do you think - keep as is or introduce a breaking change?

Half-Shot: We tend to favour dropping support for legacy options in the config, and adding backwards compatibility (with noisy warnings in the code instead). https://github.com/matrix-org/matrix-appservice-irc/blob/401c1ac0d308f26c7cc9fca0aa51941fbcfd16e0/src/irc/IrcServer.ts#L784-L790

At a later point, we can remove the old setting.

3nprob: Sounds reasonable, I'll take a stab at that

Half-Shot: How did the stabbing go?

3nprob:
@Half-Shot Sorry for leaving this hanging. I have updated it in line with your feedback, hopefully we can have better engagement this round ;)

tadzik (approved):
Looks good to me :) Thanks!

3nprob:
@jaller94 concern addressed?

GovanifY:
Hi, is there anything stopping the inclusion of this PR?

The only last open review has been dealt with by adding a warning about using a legacy version of the configuration options, as far as I see, so everything seems to be good to go.

PR #1375, opened 2021-06-08
Allow passing any secure opts to the IRC connection

Half-Shot:
Which will allow us to use `servername` to force the correct server address. To avoid some duplication, this PR deprecates the top level `ca` key and instead includes it inside the tlsOptions object.

### DIFF
--- a/src/irc/IrcServer.ts
+++ b/src/irc/IrcServer.ts
@@ -34,6 +34,9 @@ export interface IrcServerConfig {
     ssl?: boolean;
     sslselfsign?: boolean;
     sasl?: boolean;
+    tlsOptions?: {
+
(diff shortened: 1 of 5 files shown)


src/irc/IrcServer.ts
```diff
@@ -34,6 +34,9 @@ export interface IrcServerConfig {
     ssl?: boolean;
     sslselfsign?: boolean;
     sasl?: boolean;
+    tlsOptions?: {
+
+    };
```
jaller94, line 39: Is there anything missing?
```suggestion
    tlsOptions?: {};
```

jaller94 (approved):
Looks good to me.

PR #1376, opened 2021-06-08
Exclusively allow additional addresses when connecting to IRC

Half-Shot:
This is so that we can connect to `irc.matrix.libera.chat` exclusively while retaining the same config key and name in bridge messages.

### DIFF
--- a/src/irc/IrcServer.ts
+++ b/src/irc/IrcServer.ts
@@ -37,6 +37,7 @@ export interface IrcServerConfig {
     password?: string;
     allowExpiredCerts?: boolean;
     additionalAddresses?: string[];
+    onlyAdditionalAddresses?: boolean;
     dynamicChannels: {
         enabled: boolean;
         published: boolean;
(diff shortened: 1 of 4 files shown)


src/irc/IrcServer.ts
```diff
@@ -770,7 +772,10 @@ export class IrcServer {
         }
 
         this.addresses = config.additionalAddresses || [];
-        this.addresses.push(this.domain);
+        // Don't include the original domain if not configured to.
+        if (!config.onlyAdditionalAddresses) {
+            this.addresses.push(this.domain);
+        }
```
jaller94, line 778: If you don't want runtime errors (division by zero), you need to throw here if `.length === 0`.

src/irc/IrcServer.ts
```diff
@@ -37,6 +37,7 @@ export interface IrcServerConfig {
     password?: string;
     allowExpiredCerts?: boolean;
     additionalAddresses?: string[];
+    onlyAdditionalAddresses?: boolean;
```
jaller94, line 40: It's optional? Even though we have a default?

jaller94 (changes requested):
Introduces a runtime error if no addresses are in the addtionalAddresses list and onlyAdditionalAddresses is true.

src/irc/IrcServer.ts
Half-Shot: point

jaller94 (approved):
Nice! Looks all good to me now.

PR #1379, opened 2021-06-09
Do not use roomConfigs for PMs

### LINKED ISSUE #1378: RoomConfig should not attempt to fetch state from PM rooms
Because the appservice user isn't joined, and there is little benefit to having per-room configs for PMs.

Half-Shot:
Fixes #1378


benparsons (approved):
lgtm

PR #1380, opened 2021-06-09
Inline "err_nononreg" errors for PMing a user

Half-Shot:
When a Matrix user attempts to PM a IRC user and that user has unregistered PMs turned off, the error would propagate to their admin room but not inline in the DM. This basically needs us to add this error type as a inlineable error.


benparsons (approved):
lgtm

PR #1381, opened 2021-06-09
Add support for headers in the admin help text

Half-Shot:
This PR adds headers to the help text to distinguish between commands, since our command list has grown somewhat:

image

I've also alphabetically ordered the list as it was bugging @Twi1ightSparkle that we didn't do that :)


benparsons (approved):
lgtm

Issue #1323, opened 2021-05-19
Unlinking a provisioned room still causes the bot to rejoin after leaving to set permissions


Half-Shot:
![image](https://user-images.githubusercontent.com/2072976/118798179-43ad4980-b895-11eb-8466-1ccf9da6b4bc.png)

PureTryOut:
This made it really hard to unbridge several Freenode rooms and bridge them to OFTC instead. Would be great to have this fixed quickly, now people are leaving Freenode en-mass.

aperezdc:
I suppose this is the reason for the following error when trying to setup the bridging after removing Freenode:

> Bridge returned HTTP 500 Internal Server Error: Room failed validation. You may be attempting to "double bridge" this room. Error: 2

Is there some workaround for this in the meantime?

ftyers:
Yes, and nope, just got to wait...

Half-Shot:
Looking into this one today.

liskin:
Someone somewhere mentioned that bridging to another room and unbridging may fix this issue. I've just tried it and it did work indeed! The steps I did:

1. join an empty IRC channel without ChanServ so that you're the only op there
2. bridge the Matrix room to this channel (this will require an op on IRC side to confirm)
3. leave that channel on IRC side so that when it's unbridged, there is nobody left in there (as you do it, appservice-irc will change the power level of your IRC puppet to Default)
4. remove the bridge on Matrix side (now appservice-irc has no reason to join again and change the power levels as there is nobody left to change)

ftyers:
Thanks for the suggestion! That didn't work for me.

srevinsaju:
The above workaround didnt work for me, but another did: Perhaps it might help anyone who is looking for a solution:
1. Make sure that no freenode users are on the matrix channel
2. Use the integration manager > Select IRC > Select Network Name "Freenode" > Select your old IRC channel name and complete the authorization
3. Once you are done, close the integration manager (this step is important)
4. Now open the integration manager again. You will see "IRC" in the "Existing integration" section
5. Click "IRC" and hit "Unlink" button. Automatically `@appservice:matrix.org` will leave the matrix channel.
6. Select new IRC network, eg: Libera, OFTC in the same integration manager, and complete the authorization
7. React with 👍 or :-1: if it worked for you or not :stuck_out_tongue:

kfiven:
Ok I was able to figure this out and bridge succeeded.

[...]

PR #1382, opened 2021-06-09
Await leaving users before removing the bridge bot

Half-Shot:
Fixes https://github.com/matrix-org/matrix-appservice-irc/issues/1323

The provisioner unbridges all IRC ghosts from a room and will unset the powerlevels for those users upon a unlink request, while also removing the bridge bot from the room. However, it would not await the removal of the ghost users / PL unset before trying to remove the bot user.
This lead to the bot rejoining the room as it would then try to set the PLs after leaving. Typically this meant a lot of users would find that the bridge bot became wedged in the room forever, meaning they could not then bridge to other networks.

### DIFF
--- a/changelog.d/1382.bugfix
+++ b/changelog.d/1382.bugfix
@@ -1,0 +1,1 @@
+Fix a bug where the bridge user would rejoin shortly after unbridging a room.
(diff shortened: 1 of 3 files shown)


benparsons (approved):
lgtm

PR #1383, opened 2021-06-10
Fix expecting advanced to be defined when hot reloading config

Half-Shot:
Fixes an issue where hot reloads would fail if advanced was not defined.

### DIFF
--- a/src/bridge/IrcBridge.ts
+++ b/src/bridge/IrcBridge.ts
@@ -209,7 +209,7 @@ export class IrcBridge {
 
         this.appservice = new AppService({
             homeserverToken,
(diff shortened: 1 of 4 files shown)


src/bridge/IrcBridge.ts
```diff
@@ -218,8 +218,8 @@ export class IrcBridge {
         log.info(`Bridge config was reloaded, applying changes`);
         const oldConfig = this.config;
 
-        if (oldConfig.advanced.maxHttpSockets !== newConfig.advanced.maxHttpSockets) {
```
jaller94, line 221: Note to myself: This seems to have been the critical line.

jaller94 (approved):
Looks good to me.

PR #1384, opened 2021-06-10
Always return empty object when handing secure opts

Half-Shot:
Ref #1375 

A falsey value of secureOpts turns off SSL, which is not what we want to be doing and was a breaking behaviour.

### DIFF
--- a/src/irc/IrcServer.ts
+++ b/src/irc/IrcServer.ts
@@ -323,7 +323,8 @@ export class IrcServer {
     }
 
     public getSecureOptions() {
-        return this.config.tlsOptions;
+        // Return undefined here, as a falsy secure opts will disable SSL.
+        return this.config.tlsOptions ?? {};
     }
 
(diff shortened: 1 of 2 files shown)


src/irc/IrcServer.ts
```diff
@@ -323,7 +323,8 @@ export class IrcServer {
     }
 
     public getSecureOptions() {
-        return this.config.tlsOptions;
+        // Return undefined here, as a falsy secure opts will disable SSL.
```
jaller94:
```suggestion
        // Don't return undefined here, as a falsy secure opts will disable SSL.
```

jaller94 (approved):
Except for the comment it all makes sense.

src/irc/IrcServer.ts
Half-Shot: Oh, I spotted that too :D.

PR #1388, opened 2021-06-11
Bump matrix-org-irc to 1.0.0

Half-Shot:
This fixes a bug which can cause the bridge to crash (https://github.com/matrix-org/node-irc/pull/69) so we will need it for RC3.

### DIFF
--- a/changelog.d/1388.bugfix
+++ b/changelog.d/1388.bugfix
@@ -1,0 +1,1 @@
+Update `matrix-org-irc` to `1.0.0` to fix a bug where the bridge can crash.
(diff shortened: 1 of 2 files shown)


jaller94:
What is this about?
You want a release because of the linked PR?
How comes this is 1.0.0 (commonly the first stable version) and why is the description talking about some Release Candidate No.3 (presumably 0.27.0-rc3) instead?

jaller94 (approved):
Looks good to me.

Half-Shot:
This is a release to bump the version of a package we depend on, because the version we currently depend on had an issue which can cause the bridge to crash (I linked https://github.com/matrix-org/node-irc/pull/69). `1.0.0` is the version of `matrix-org-irc` that contains the fix. I'm taking about RC3 because our current RC2 for this bridge is affected by the issue in the linked PR, and would like to get this change merged so we can release an RC3 before the weekend.

PR #1394, opened 2021-06-16
Fix IRC channel joins never completing.

Half-Shot:
Previously, it was possible for `_joinChannel` to never resolve/reject the promise if the join timed out. This was introduced in one of our PRs that changed how we handled deferred joins. This should hopefully fix the issue where we can have users who are unable to connect to channels.

### DIFF
--- a/src/irc/BridgedClient.ts
+++ b/src/irc/BridgedClient.ts
@@ -442,12 +442,17 @@ export class BridgedClient extends EventEmitter {
         });
     }
 
-    public leaveChannel(channel: string, reason = "User left") {
+    public async leaveChannel(channel: string, reason = "User left") {
         if (this.state.status !== BridgedClientStatus.CONNECTED) {
             return Promise.resolve(); // we were never connected to the network.
         }
         if (!channel.startsWith("#")) {
             return Promise.resolve(); // PM room
+        }
+        const deferredChannelJoin = this.channelJoinDefers.get(channel);
+        if (deferredChannelJoin) {
+            // We are in the process of joining this channel, so await the join before trying to leave.
+            await deferredChannelJoin;
         }
         if (!this.inChannel(channel)) {
             return Promise.resolve(); // we were never joined to it.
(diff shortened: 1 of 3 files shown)


src/irc/BridgedClient.ts
```diff
@@ -442,13 +442,18 @@ export class BridgedClient extends EventEmitter {
         });
     }
 
-    public leaveChannel(channel: string, reason = "User left") {
+    public async leaveChannel(channel: string, reason = "User left") {
         if (this.state.status !== BridgedClientStatus.CONNECTED) {
             return Promise.resolve(); // we were never connected to the network.
         }
         if (!channel.startsWith("#")) {
             return Promise.resolve(); // PM room
```
jaller94:
```suggestion
            return; // we were never connected to the network.
        }
        if (!channel.startsWith("#")) {
            return; // PM room
```

src/irc/BridgedClient.ts
```diff
@@ -442,13 +442,18 @@ export class BridgedClient extends EventEmitter {
...
         }
+        const deferredChannelJoin = this.channelJoinDefers.get(channel);
+        if (deferredChannelJoin) {
+            // We are in the process of joining this channel, so await the join before trying to leave.
+            await deferredChannelJoin;
+        }
         if (!this.inChannel(channel)) {
             return Promise.resolve(); // we were never joined to it.
```
jaller94:
```suggestion
            return; // we were never joined to it.
```

src/irc/BridgedClient.ts
```diff
@@ -442,17 +442,28 @@ export class BridgedClient extends EventEmitter {
         });
     }
 
-    public leaveChannel(channel: string, reason = "User left") {
+    public async leaveChannel(channel: string, reason = "User left"): Promise<void> {
         if (this.state.status !== BridgedClientStatus.CONNECTED) {
-            return Promise.resolve(); // we were never connected to the network.
+            return undefined; // we were never connected to the network.
```
jaller94, line 447: Does TypeScript prefer `return undefined` over `return`?

src/irc/BridgedClient.ts
Half-Shot: Aha, got there ahead of you :p

src/irc/BridgedClient.ts
Half-Shot: Our linter is a pain, if we return a defer.promise then it expects all the others to return something. And yes, eslint makes a distinction between `return undefined` and `return;`;

Half-Shot: I am wondering whether consistent return is useful given we have types now, so as long as we type the function responses it's probably overkill.

src/irc/BridgedClient.ts
```diff
@@ -971,51 +1027,17 @@ export class BridgedClient extends EventEmitter {
                     this, `Could not join ${channel} on '${this.server.domain}': ${err.command}`, true
                 );
             }
+            // Otherwise, not a failure we recognise. This will eventually timeout.
```
jaller94, line 1030:
```suggestion
            // Otherwise, not a failure we recognise. This will eventually time out.
```

src/irc/BridgedClient.ts
```diff
@@ -442,17 +442,28 @@ export class BridgedClient extends EventEmitter {
...
+        }
+        const deferredChannelJoin = this.channelJoinDefers.get(channel);
+        if (deferredChannelJoin) {
+            // We are in the process of joining this channel, so await the join before trying to leave.
+            try {
+                await deferredChannelJoin;
+            }
+            catch (ex) {
+                // Given we're trying to leave, this isn't critical.
+                this.log.debug(`Channel join failed to complete while leaving channel`, ex);
```
jaller94, line 460: Looks good. What a great way to deal with the case of an ongoing join.

jaller94 (approved):
Looks all good to me.
Hope you tested it well, as I haven't done that as part of the review.

Issue #1361, opened 2021-05-31
!help should mention that !username is used for SASL when the network supports it, possibly even make login help more clear


Mikaela:
**Describe the bug**

There is no indication that `!username` is used for SASL and `!nick` is not enough.

```
username [irc.example.net] username : Store a username to use for future connections.
!storepass [irc.example.net] passw0rd : Store a NickServ OR SASL password (server password)
```

**To Reproduce**

1. Send `!help`
2. `!storepass password`
3. Expect to be logged in, but you aren't

**Expected behavior**

The help for `!username` should mention that it's used and needed for SASL. Additionally it would be helpful if the help text began with something like `To identify with SASL, set your username with **!username username** and password with **!storepass password** so on the next **!reconnect** you will be automatically identified.` so the information doesn't have to be read very carefully from the long output.

**Screenshots**

N/A

Sysinfo: N/A

**Additional context**

This seems to be a frequently encountered problem on `#libera-matrix:libera.chat`

> BridgeVersion: 0.26.0

jluebbe:
Also, in #libera-matrix, a common confusion is that the username would refer to the matrix-id, not the nickserv account.

kerel-fs:
I had the same problem and found this issue via web search which solved my authentication problem, thanks!

The in-tree documentation describes the [SASL authentication](https://matrix-org.github.io/matrix-appservice-irc/latest/usage.html#sasl-authentication) and its needs already. But despite knowing the in-tree documentation I didn't know this specific page until much later, maybe because it is "hidden" from the list-of-contents in the main page of section `2. Usage` and doesn't have its own sub-section (slightly OT here).

Thus I propose:

**:bulb: Possible alternative solution**
The `!help` command should contain a link to the [Matrix IRC Bridge Usage Guide](https://matrix-org.github.io/matrix-appservice-irc/latest/usage.html#sasl-authentication) intended for new or experienced users in the in-tree documentation.

PR #1402, opened 2021-06-23
Add an in-tree documentation link to the !help command

kerel-fs:
This may be a seen as an alternative fix for #1361.

This is a  drive-through contribution / my first interaction with this repository and I do not know (yet?) how to test the actual formatting created by this code. In case it is wrong please let me know.


Half-Shot (approved):
No gripes from me. Thank you this is excellent!

PR #1405, opened 2021-06-29
Fix typo in registration documentation

lindhe:
The text was referring to something other than the previous command, and this change attempts to fix the text and the command to make more sense in context.

From the context of the sample output, I'm guessing that it should be `http://localhost:9999`, but I cannot really know. Since I don't know the real answer here (just trying to follow instructions and noticed issues), please think about what would make most sense and change it accordingly. :)

### DIFF
--- a/changelog.d/1405.doc
+++ b/changelog.d/1405.doc
@@ -1,0 +1,2 @@
+Fix typo regarding examples of hostname and port in Bridge Setup documentation (4. Registration)
+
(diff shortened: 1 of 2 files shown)


Half-Shot:
Thanks @lindhe, this looks sane to me :). Could you follow https://github.com/matrix-org/matrix-appservice-irc/blob/develop/CONTRIBUTING.md#making-pull-requests and add a signoff and newsfile. Afterwards I can merge this one.

Half-Shot (approved):

PR #1406, opened 2021-07-01
Fix inverted logic in privacy filter

Half-Shot:
This bites everyone who sets their rooms to not require membership to be synced to IRC.


jaller94 (approved):
Looks good.
I also re-reviewed `allowUnconnectedMatrixUsers` which looks good.

PR #1418, opened 2021-07-16
Add prometheus metrics for IRC connection times

tadzik:


src/bridge/IrcBridge.ts
```diff
@@ -728,6 +735,13 @@ export class IrcBridge {
         }
     }
 
+    public logTime(key: keyof Timers, time: number) {
```
Half-Shot, line 738: Yay for keyof!

Half-Shot (approved):
This will be super useful in measuring connection times. Woop

PR #1420, opened 2021-07-18
Document the Debug API in mdbook

Half-Shot:
This is intending to replace https://github.com/matrix-org/matrix-appservice-irc/wiki/Debug-API.


docs/debug_api.md
```diff
@@ -0,0 +1,165 @@
...
+}
+```
+
+#### Response Body
+
+The response body will contain a JSON array of stages that were successful and failed as
+it's possible for this command to only be partially successful.
+
+```json5
+error: [],
```
tadzik, line 89: Could we put some example errors and stages in there? From this sample it's hard to know what the JSON arrays will contain.

docs/debug_api.md
```diff
@@ -0,0 +1,165 @@
...
+Idleness is calculated by how long it has been since a user has sent a message/joined/left a room.
+This is calulated by whether the appservice bot or it's users have seen the user perform any actions 
+(i.e. left a IRC bridged room or sent a message). Due to limitations of Matrix, it is not possible to 
+discover "lurkers".
+
+#### Request Parameters
+
+- `server` is the servername you wish to disconnect users from. This is the key of 
+  your server configuration object in the config section.
+- `since` is the number of hours a user has been idle for to be considered `idle`.
```
tadzik, line 125: An integer, I assume? May be worth documenting that just in case :)

docs/administrators_guide.md
```diff
@@ -86,7 +87,4 @@ can be found in [GRAFANA.md](https://github.com/matrix-org/matrix-appservice-irc
 ## The Debug API
 
 The Debug API allows you to perform administrative actions on the bridge such as killing a IRC
-connection, inspecting connected users or unbridging a room. The API is currently loosely documented
-over in [the GitHub wiki]([https://github.com/matrix-org/matrix-appservice-irc/wiki/Debug-API).
-
-This is not enabled by default.
\ No newline at end of file
+connection, inspecting connected users or unbridging a room. You can learn more by reading [Bridge Setup](./debug_api.md)
```
jaller94, line 90:
```suggestion
connection, inspecting connected users or unbridging a room. You can learn more by reading [Debug API](./debug_api.md)
```

tadzik, line 90: This should probably be "by reading [Debug API]".

docs/debug_api.md
```diff
@@ -0,0 +1,165 @@
+Debug API
+=========
+
+
+The Debug API is a powerful HTTP API to make adjustments to the bridge at runtime, and is typically useful
+when administrating large bridges. Some functionality has moved to the [Admin Room](./admin_room.md)
+interface as sending commands over Matrix is preferred, however many powerful commands are exposed here.
```
jaller94, line 7:
```suggestion
interface as sending commands over Matrix is preferred, however, many powerful commands are exposed here.
```

docs/debug_api.md
```diff
@@ -0,0 +1,165 @@
...
+  }
+}
+```
+
+### `POST /killRoom`
+
+Stop a room from being bridged. This will remove IRC ghost users from the room
+and disconnect Matrix users from the channel.
+
+The [Admin Room](admin_room.html#unlink) features a less
```
jaller94, line 68: `.html`? Is this needed when using anchor links?

docs/debug_api.md
```diff
@@ -0,0 +1,165 @@
+Debug API
+=========
+
+
+The Debug API is a powerful HTTP API to make adjustments to the bridge at runtime, and is typically useful
+when administrating large bridges. Some functionality has moved to the [Admin Room](./admin_room.md)
```
tadzik, line 6: I wonder if it'd make sense to have an aggregate page that lists all the admin/debug functionality, so that a user trying to discover their admin powers won't have to switch between two separate book sections.

tadzik (changes requested):

docs/debug_api.md
```diff
@@ -0,0 +1,165 @@
...
+
+Stop a room from being bridged. This will remove IRC ghost users from the room
+and disconnect Matrix users from the channel.
+
+The [Admin Room](admin_room.html#unlink) features a less
+powerful version of this command.
+
+#### Request Body
+
+```json5
```
jaller94, line 73: TIL about JSON5.

docs/debug_api.md
```diff
@@ -0,0 +1,165 @@
...
+```
+
+#### Response Body
+
+The response body will contain a JSON array of stages that were successful and failed as
+it's possible for this command to only be partially successful.
+
+```json5
+error: [],
+stages: [],
```
jaller94, line 90:
```suggestion
{
  error: [],
  stages: [],
}
```

docs/debug_api.md
```diff
@@ -0,0 +1,165 @@
...
+  "remove_alias": true, // Should the room alias for the room be removed. Default: true.
+}
+```
+
+#### Response Body
+
+The response body will contain a JSON array of stages that were successful and failed as
+it's possible for this command to only be partially successful.
+
+```json5
```
jaller94, line 88: I expect the server response will be `application/json`. That's why the example below should preferably also be json, not json5.

docs/debug_api.md
```diff
@@ -0,0 +1,165 @@
...
+stages: [],
+```
+
+### `POST /killUser`
+
+This will kill a connection to IRC for a given user on all networks they are connected to.
+
+#### Request Body
+
+```js
```
jaller94, line 99:
```suggestion
```json
```

docs/debug_api.md
```diff
@@ -0,0 +1,165 @@
...
+{
+  "user_id": "@foo:bar",
+  "reason": "Trust nobody"
+}
+```
+
+#### Response Body
+
+If a disconnection was successful, the bridge will emit "null". Otherwise it may emit an error
```
jaller94, line 108:
```suggestion
If a disconnection was successful, the bridge will emit "null". Otherwise, it may emit an error
```

docs/debug_api.md
```diff
@@ -0,0 +1,165 @@
...
+
+To access the API over `curl`:
+
+```sh
+curl http://127.0.0.1:11100/inspectUsers?access_token=AS_TOKEN_FROM_REGISTRATION_FILE
+```
+
+## Endpoints
+
+<!-- no toc -->
```
jaller94, line 29: Did you build this manual Table of Content to ignore the `h4`s of each endpoint?

jaller94 (dismissed):

docs/debug_api.md
Half-Shot: It's a query parameter, so a string

docs/debug_api.md
Half-Shot: We're trying to move a lot of the Debug API functionality over to admin rooms, because it turns out Matrix is more convenient than HTTP calls. I'm hoping we can instead reduce the API over time.

docs/debug_api.md
Half-Shot: Nope, I'm assuming bad autocorrect.

docs/debug_api.md
Half-Shot: Basically JSON that allows comments, handy.

docs/debug_api.md
Half-Shot: JSON5 is just so mdbook doesn't get upset about commas or comments, the response will still be perfectly valid standard json.

docs/debug_api.md
Half-Shot: Indeed, I felt having a list of each header was overkill. We only need a list of each endpoint.

docs/debug_api.md
tadzik: Right – I meant it as „are floats allowed?”

Half-Shot: ahhhhhhhhhhhh no, will update

tadzik (approved):

Issue #1363, opened 2021-06-01
!storepass with 3 parameters applies to primary network when network isn't recognized


TheCycoONE:
**Describe the bug**
On my homeserver when I try to store my password for a network that isn't Freenode, Freenode's password get's changed instead:

**To Reproduce**
1. configure matrix-appserver-irc with second IRC network
2. open @appservice-irc@myserver.com room
2. type !storepass [network] password
3. observe response

**Expected behavior**
I expect the irc.libera.chat password to be updated (or else an error that the network name is invalid)

(Additionally I wouldn't expect @apservice-irc to be called "Freenode IRC Bridge status" unless there was another channel for the second network)

**Screenshots**
<img width="698" alt="Screen Shot 2021-06-01 at 4 12 20 PM" src="https://user-images.githubusercontent.com/823433/120384545-4c216d80-c2f4-11eb-8407-4083f70ae811.png">

**Desktop (please complete the following information):**
 - OS: MacOS and ArchLinux
 - Browser: Firefox
 - Version: 88.0.1

**Additional context**
```yaml
  servers:
    chat.freenode.net:
      name: "Freenode"

      additionalAddresses: [ "chat.freenode.net" ]
      port: 6697
      ssl: true
      sslselfsign: false
      sasl: true
      allowExpiredCerts: false
      sendConnectionMessages: true
      quitDebounce:
        enabled: false
        quitsPerSecond: 5
        delayMinMs: 3600000 # 1h
        delayMaxMs: 7200000 # 2h

      modePowerMap:
        o: 50
        v: 1

      botConfig:
        enabled: false
        nick: "MatrixBot"
        username: "matrixbot"
        joinChannelsIfNoUsers: true

      privateMessages:
        enabled: true
        federate: true

      dynamicChannels:
        enabled: true
        createAlias: true
        published: false
        joinRule: public
        groupId: +freenode:sealbaker.com
        federate: false
        aliasTemplate: "#freenode_$CHANNEL"
      membershipLists:
        enabled: true
        floodDelayMs: 10000

        global:
          ircToMatrix:
            initial: true
            incremental: true

          matrixToIrc:
            initial: true
            incremental: true

        rooms:
          - room: "!fuasirouddJoxtwfge:localhost"
            matrixToIrc:
              initial: false
              incremental: false

        channels:
          - channel: "#foo"
            ircToMatrix:
              initial: false
              incremental: false

        ignoreIdleUsersOnStartup:
          enabled: true
[...]

PR #1423, opened 2021-07-19
Require explicit server selection for !storepass when more than one possibility exists

tadzik:
This avoids the situation described in #1363 when a password could've
been accidentally set (and used) with the wrong server due to the
command parser trying to be too helpful.


changelog.d/1363.bugfix
```diff
@@ -0,0 +1,3 @@
+Require explicit server selection for !storepass when more than one possibility exists.
+
+This makes the command a bit more verbose, but avoids the situation where a password could've been accidentally specified for the wrong server.
```
Half-Shot, line 3: Try to keep to one line where possible, as it makes the changelog less wonky.

Half-Shot (approved):
One nit, otherwise looks good to me.

PR #1424, opened 2021-07-19
Change the reply rendering to something more IRCish (and configurable)

### LINKED ISSUE #1307: Reply format change proposal
**Is your feature request related to a problem? Please describe.**

Currently, the reply formatting in the bridge on the IRC side frustrates a lot of users, as it is difficult to read. I'd like to propose an alternative set of formatting rules, that should be relatively easy to implement as a first pass, and approximate IRC conventions well enough to prevent the worst frustrations.

**Describe the solution you'd like**

The reply formatting should vary depending on how old the replied-to message is; this is to mirror the different IRC conventions when replying to someone who has just talked, vs. someone who said something a while ago that you wanted to make a belated reply on.

For quick replies, it should translate to just a mention:

```
<you> otherperson: hi this is my reply
```

For belated replies, the (first line of) the original message is sent in IRC-quote-style first:

```
<you> <otherperson> the original message being replied to
<you> hi this is my reply
```

The exact cut-off between a quick and a belated reply is not too important, I think, so long as it's something vaguely reasonable. Ideally something along the lines of "more than 20 messages ago", but if this requires difficult state tracking, something like "more than 5 minutes ago" could also work as a first approximation.

These rules could undoubtedly be improved over time (to eg. account for things like "more than 5 messages from this specific person ago"), but that is outside of the scope of this proposal. The current priority is to fix the worst of the frustration caused by the current reply format, which is *definitely* bad.

**Describe alternatives you've considered**

None in too much depth. This is a first approximation to the problem which tries to approximate IRC conventions - better than nothing at all, and relatively easy to implement.

**Additional context**

 N/A

tadzik:
Fixes #1307

### DIFF
--- a/src/bridge/IrcBridge.ts
+++ b/src/bridge/IrcBridge.ts
@@ -193,9 +193,9 @@ export class IrcBridge {
             maxActionDelayMs: 5 * 60 * 1000, // 5 mins,
             defaultTtlMs: 10 * 60 * 1000, // 10 mins
         });
-        this.matrixHandler = new MatrixHandler(this, this.config.matrixHandler || {}, this.membershipQueue);
+        this.matrixHandler = new MatrixHandler(this, this.config.ircService.matrixHandler || {}, this.membershipQueue);
         this.privacyProtection = new PrivacyProtection(this);
-        this.ircHandler = new IrcHandler(this, this.config.ircHandler, this.membershipQueue, this.privacyProtection);
+        this.ircHandler = new IrcHandler(this, this.config.ircService.ircHandler, this.membershipQueue, this.privacyProtection);
 
         // By default the bridge will escape mxids, but the irc bridge isn't ready for this yet.
         MatrixUser.ESCAPE_DEFAULT = false;
@@ -238,8 +238,12 @@ export class IrcBridge {
             log.info(`Adjusted media_url to ${newConfig.homeserver.media_url}`);
         }
 
-        this.ircHandler.onConfigChanged(newConfig.ircHandler || {});
-        this.config.ircHandler = newConfig.ircHandler;
+        this.ircHandler.onConfigChanged(newConfig.ircService.ircHandler || {});
+        this.config.ircService.ircHandler = newConfig.ircService.ircHandler;
+
+        this.matrixHandler.onConfigChanged(newConfig.ircService.matrixHandler || {});
+        this.config.ircService.matrixHandler = newConfig.ircService.matrixHandler;
+
         this.config.ircService.permissions = newConfig.ircService.permissions;
         this.roomConfigs.config = newConfig.ircService.perRoomConfig;
 
--- a/src/bridge/MatrixHandler.ts
+++ b/src/bridge/MatrixHandler.ts
@@ -95,7 +95,7 @@ interface CachedEvent {
     timestamp: number;
 }
 
-interface Config {
+export interface MatrixHandlerConfig {
     eventCacheSize?: number;
     shortReplyTresholdSeconds?: number;
     shortReplyTemplate?: string;
@@ -117,7 +117,7 @@ export class MatrixHandler {
 
     constructor(
         private ircBridge: IrcBridge,
-        private config: Config = {},
+        private config: MatrixHandlerConfig = {},
         private readonly membershipQueue: MembershipQueue) {
         // maintain a list of room IDs which are being processed invite-wise. This is
         // required because invites are processed asyncly, so you could get invite->msg
@@ -1342,6 +1342,9 @@ export class MatrixHandler {
     }
 
     // EXPORTS
+    public onConfigChanged(config: MatrixHandlerConfig) {
+        this.config = config;
+    }
 
     public onMemberEvent(req: BridgeRequest, event: OnMemberEventData) {
         return reqHandler(req, this._onMemberEvent(req, event));
--- a/src/bridge/MatrixHandler.ts
+++ b/src/bridge/MatrixHandler.ts
@@ -10,6 +10,7 @@ import { IrcAction } from "../models/IrcAction";
 import { toIrcLowerCase } from "../irc/formatting";
 import { AdminRoomHandler } from "./AdminRoomHandler";
 import { trackChannelAndCreateRoom } from "./RoomCreation";
+import { renderTemplate } from "../util/Template";
 
 async function reqHandler(req: BridgeRequest, promise: PromiseLike<unknown>) {
     try {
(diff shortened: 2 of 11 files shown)


src/bridge/MatrixHandler.ts
```diff
@@ -1232,7 +1253,8 @@ export class MatrixHandler {
                     rplSource = eventContent.content.body;
                 }
                 rplSource = rplSource.substring(0, REPLY_SOURCE_MAX_LENGTH);
-                this.eventCache.set(eventId, {sender: rplName, body: rplSource});
+                cachedEvent = {sender: rplName, body: rplSource, timestamp: eventContent.origin_server_ts};
+                this.eventCache.set(eventId, cachedEvent); // TODO can overflow the cache
```
tadzik, line 1257: Fixme before merging, preferably wrap it in a helper method.

src/util/Template.ts
```diff
@@ -0,0 +1,8 @@
+export function renderTemplate(template: string, args: { [key: string]: string }): string {
```
tadzik, line 1: Does this need a copyright clause like in some other files? @Half-Shot 

Most seem to have it, but PackageInfo.ts does not, for instance.

tadzik, line 1: This should replace existing manual replaces scattered around the codebase.

Half-Shot: Yes, ones that don't are bad and wrong.

tadzik: Fixed in 6b6146bf. Should we add a linter rule/CI script that'll check for that? (Probably in a separate task).

src/bridge/MatrixHandler.ts
tadzik: Fixed in a332dfc8. We can consider using QuickLRU like the slack bridge does.

src/util/Template.ts
tadzik: Fixed in ff39e7fa7c20c6c2b76c93666a3b6b13712c30c

src/util/Template.ts
```diff
@@ -0,0 +1,26 @@
+/*
```
Half-Shot, line 1: Yay for a template function at long last!

src/bridge/MatrixHandler.ts
```diff
@@ -30,6 +31,12 @@ const MSG_PMS_DISABLED_FEDERATION = "[Bridge] Sorry, PMs are disabled on this br
 const DEFAULT_EVENT_CACHE_SIZE = 4096;
 /* Length of the source text in a formatted reply message */
 const REPLY_SOURCE_MAX_LENGTH = 32;
+// How many seconds needs to pass between a message and a reply to it to switch to the long reply format
+const DEFAULT_SHORT_REPLY_TRESHOLD_SECONDS = 5 * 60;
```
Half-Shot, line 35: We tend to store values in ms across the bridge. In this case, I guess it's fine because the config is in seconds?

src/bridge/MatrixHandler.ts
```diff
@@ -30,6 +31,12 @@ const MSG_PMS_DISABLED_FEDERATION = "[Bridge] Sorry, PMs are disabled on this br
...
+// Format of replies sent shortly after the original message
+const DEFAULT_SHORT_REPLY_TEMPLATE = "$NICK: $REPLY";
+// Format of replies sent a while after the original message
+const DEFAULT_LONG_REPLY_TEMPLATE = "<$NICK> \"$ORIGINAL\" <- $REPLY";
```
Half-Shot, line 39: I wish we had a default config somewhere rather than having consts at the top of the file, but I see I have also done this.

src/bridge/MatrixHandler.ts
```diff
@@ -97,7 +117,7 @@ export class MatrixHandler {
 
     constructor(
         private ircBridge: IrcBridge,
-        config: {eventCacheSize?: number} = {},
+        private config: Config = {},
```
Half-Shot, line 120: This config won't be hot-reloaded, might be worth doing the do now that the config is more useful. See `IrcHandler` / `IrcBridge` on how we do config reloads.

Half-Shot (commented):
A few thoughts, but generally this is super exciting!

src/bridge/MatrixHandler.ts
tadzik: Correct.

The config file uses a plethora of units, as in `floodDelayMs`, `quitsPerSecond` and even `idleForHours`. Interestingly, `ircClients.idleTimeout` is the only one that doesn't have its unit in the name, and its unit turns out to be seconds also :) It'd be hellish to fix it now, but it is at least nicely documented to be that way.

src/bridge/MatrixHandler.ts
tadzik: Yeah, I'm not super fond of it, but everything seems to be kept in sync manually between `config.sample.yaml` and the code – sometimes in `consts` like this, sometimes in good old magic numbers. I wouldn't mind fixing that also, but it's probably a better fit for another day :)

Half-Shot: In the teams / github bridges we have magic code to generate a config sample from defaults, with comments. I should port that to the common bridge lib some day.

src/bridge/IrcBridge.ts
```diff
@@ -193,9 +193,9 @@ export class IrcBridge {
             maxActionDelayMs: 5 * 60 * 1000, // 5 mins,
             defaultTtlMs: 10 * 60 * 1000, // 10 mins
         });
-        this.matrixHandler = new MatrixHandler(this, this.config.matrixHandler || {}, this.membershipQueue);
+        this.matrixHandler = new MatrixHandler(this, this.config.ircService.matrixHandler || {}, this.membershipQueue);
         this.privacyProtection = new PrivacyProtection(this);
-        this.ircHandler = new IrcHandler(this, this.config.ircHandler, this.membershipQueue, this.privacyProtection);
+        this.ircHandler = new IrcHandler(this, this.config.ircService.ircHandler, this.membershipQueue, this.privacyProtection);
```
Half-Shot, line 198: Oh YIKES

src/bridge/MatrixHandler.ts
Half-Shot: Fair, one day we should refactor the config to be sane in many ways, and just deprecate the old ways.

src/bridge/MatrixHandler.ts
tadzik: It wasn't actually loaded at all /o\ I fixed that and IrcHandler config (which was broken for similar reasons) in 48c6e48f.

Half-Shot (approved):
Looking amazing. Thanks for fixing yet another bug that makes me go "oh fuck".

PR #1427, opened 2021-07-21
Make the default port (9999) documented and used consistently

### LINKED ISSUE #1334: Default port in documentation and configs does not match the actual default listen port
In the configuration file (`config.sample.yaml`) `bindPort: 9999` is commented out by default, which makes it look like the bridge will listen on port `9999` if not specified otherwise. This is not the case, the bridge actually listens on port `8090`. This was introduced here: https://github.com/matrix-org/matrix-appservice-irc/pull/1274
The linked PR actually intended to change the default port to 8090 in all configurations, scripts and referencing documents, but this was reverted as this port change would break existing setups.

However, I find the current solution confusing as I was wondering quite some time why my homeserver was unable to reach the bridge. That's why I think that at least the sample config should match the actual port.

tadzik:
Fixes #1334


jaller94:
I'm all for this, but after a discussion this had previously not been changed to not break bridges for people who didn't configure a port explicitly.
https://github.com/matrix-org/matrix-appservice-irc/pull/1274#discussion_r602259333

jaller94:
@tadzik  It seems like an easy win to make 8090 the default port, sticking with the current default in code.

tadzik:
> @tadzik It seems like an easy win to make 8090 the default port, sticking with the current default in code.

Yeah, agreed. It's now 8090 everywhere, so the code is intact but the docs have been updated.

PR #1428, opened 2021-07-21
Truncate original messages more gently when replying

### LINKED ISSUE #801: reply prefixes should be rounded to the nearest word and ellipsis added if truncated

tadzik:
Fixes #801

 * No longer breaks unicode characters in half
 * No longer forgets to add the ellipsis in some cases
 * Tries to stop at the word boundary unless that'd make the result too short

### DIFF
--- a/spec/unit/TrimString.spec.js
+++ b/spec/unit/TrimString.spec.js
@@ -2,35 +2,27 @@
 const { trimString } = require("../../lib/util/TrimString.js");
 
 describe("trimString", function() {
-    it("should not cut unicode characters in half", (done) => {
+    it("should not cut unicode characters in half", () => {
         const input = "lol 😅";
         const result = trimString(input, 5);
         expect(result).toEqual(input);
-
-        done();
     });
 
-    it("should trim trailing whitespace by itself", (done) => {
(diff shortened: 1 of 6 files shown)


spec/unit/TrimString.spec.js
```diff
@@ -0,0 +1,36 @@
+"use strict";
+const { trimString } = require("../../lib/util/TrimString.js");
+
+describe("trimString", function() {
+    it("should not cut unicode characters in half", (done) => {
+        const input = "lol 😅";
+        const result = trimString(input, 5);
+        expect(result).toEqual(input);
+
+        done();
```
Half-Shot, line 10: Shouldn't need this for sync tests. Shouldn't need it for async tests either if you return a promise :)

Half-Shot (approved):
Looks good, you're overdoing it a bit on your tests though :)

spec/unit/TrimString.spec.js
tadzik: Indeed, good catch. Fixed in 9cbb200e

Half-Shot (approved):

PR #1430, opened 2021-07-22
Show message previews for messages too long to be bridged

tadzik:
This ports #529 to the recent codebase (credit goes to @14mRh4X0r++).

Requires https://github.com/matrix-org/node-irc/pull/72 (which requires a release).

Looks roughly like this:

image

### DIFF
--- a/spec/integ/matrix-to-irc.spec.js
+++ b/spec/integ/matrix-to-irc.spec.js
@@ -647,7 +647,7 @@ describe("Matrix-to-IRC message bridging", function() {
                 body: tBody.join("\n"),
                 msgtype: "m.text"
             },
-            user_id: testUser.id,
+            sender: testUser.id,
             room_id: roomMapping.roomId,
             type: "m.room.message"
         });
--- a/spec/integ/matrix-to-irc.spec.js
+++ b/spec/integ/matrix-to-irc.spec.js
@@ -625,8 +625,8 @@ describe("Matrix-to-IRC message bridging", function() {
     });
 
     it("should truncate multiline messages and include a full message URL", function(done) {
-        let tBody = "This\nis\na\nmessage\nwith\nmultiple\nline\nbreaks".split('\n');
-        let sdk = env.clientMock._client(config._botUserId);
+        const tBody = "This\nis\na\nmessage\nwith\nmultiple\nline\nbreaks".split('\n');
+        const sdk = env.clientMock._client(config._botUserId);
 
         sdk.uploadContent.and.returnValue(Promise.resolve("mxc://deadbeefcafe"));
 
--- a/spec/integ/matrix-to-irc.spec.js
+++ b/spec/integ/matrix-to-irc.spec.js
@@ -624,6 +624,69 @@ describe("Matrix-to-IRC message bridging", function() {
         await p;
     });
 
+    it("should truncate multiline messages and include a full message URL", function(done) {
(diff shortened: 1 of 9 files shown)


spec/integ/matrix-to-irc.spec.js
```diff
@@ -624,6 +624,69 @@ describe("Matrix-to-IRC message bridging", function() {
         await p;
     });
 
+    it("should truncate multiline messages and include a full message URL", function(done) {
+        let tBody = "This\nis\na\nmessage\nwith\nmultiple\nline\nbreaks".split('\n');
+        let sdk = env.clientMock._client(config._botUserId);
```
Half-Shot, line 629:
```suggestion
        const tBody = "This\nis\na\nmessage\nwith\nmultiple\nline\nbreaks".split('\n');
        const sdk = env.clientMock._client(config._botUserId);
```

spec/integ/matrix-to-irc.spec.js
```diff
@@ -624,6 +624,69 @@ describe("Matrix-to-IRC message bridging", function() {
...
+            expect(text.indexOf('deadbeefcafe')).not.toEqual(-1);
+            done();
+        });
+
+        env.mockAppService._trigger("type:m.room.message", {
+            content: {
+                body: tBody.join("\n"),
+                msgtype: "m.text"
+            },
+            user_id: testUser.id,
```
Half-Shot, line 650: (legacy tests might use user_id, they should all be sender as the Matrix spec changed several years ago)

```suggestion
            sender: testUser.id,
```

src/bridge/MatrixHandler.ts
```diff
@@ -1342,8 +1370,9 @@ export class MatrixHandler {
     }
 
     // EXPORTS
-    public onConfigChanged(config: MatrixHandlerConfig) {
-        this.config = config;
+    public onConfigChanged(config: MatrixHandlerConfig|undefined) {
+        this.config = DEFAULTS;
+        Object.assign(this.config, config);
```
Half-Shot, line 1375:
```suggestion
        this.config = {...DEFAULTS, ...config};
```

src/irc/BridgedClient.ts
```diff
@@ -1144,4 +1135,11 @@ export class BridgedClient extends EventEmitter {
         }
         return this.connectDefer.promise;
     }
+
+    private assertConnected(): Client {
+        if (this.state.status !== BridgedClientStatus.CONNECTED) {
+            throw Error('Client is not connected');
+        }
+        return this.state.client as Client;
```
Half-Shot, line 1143: Surprised you needed the `as`.

Half-Shot (approved):
Looks great, a few nits but nothing major.

src/bridge/MatrixHandler.ts
tadzik: Mmm, fancy!

src/irc/BridgedClient.ts
tadzik: I don't think it can now what `client` will be, since it depends on the `.status`, and we don't have a way (yet?) to represent that in the type system itself.

14mRh4X0r:
Thanks for taking the time to port this to the TS codebase! I already forgot about this :grimacing:

tadzik:
> Thanks for taking the time to port this to the TS codebase! I already forgot about this grimacing

My pleasure :) I can't wait to see this live. Thanks for doing all the hard work!

Issue #1431, opened 2021-07-23
AdminRoomHandler's bot requires a server address for operations that don't need one


tadzik:
**Describe the bug**
The admin room bot command parser gets very confused (and not very helpful) when multiple IRC servers are available in the bridge, asking for a server address.

**To Reproduce**
1. Configure the appservice with more than one IRC server
2. See the screenshot below

**Expected behavior**
Bot responding to commands as expected

**Screenshots**
![image](https://user-images.githubusercontent.com/86378/126765171-3b70c581-32ab-45fa-b4e4-e08d4cf5031f.png)

**Additional context**
It smells like a regression after https://github.com/matrix-org/matrix-appservice-irc/pull/1423, but it's actually not.

PR #1433, opened 2021-07-26
Refactor/ovehaul AdminRoom command handling

tadzik:
This makes them obtain information (like the IRC server or clientList)
lazily rather than eagerly, making sure that we don't require a server
to be specified for stuff that doesn't need it (like !help).

Also makes the command handler interface almost unified, in case this
comes in handy later on for some reason.

Fixes GH-1431.

### DIFF
--- a/src/bridge/AdminRoomHandler.ts
+++ b/src/bridge/AdminRoomHandler.ts
@@ -113,6 +113,10 @@ const COMMANDS: {[command: string]: Command|Heading} = {
     },
 };
 
+class ServerRequiredError extends Error {
+    notice = new MatrixAction("notice", "A server address must be specified.");
+}
+
 const USER_FEATURES = ["mentions"];
 export class AdminRoomHandler {
     private readonly botUser: MatrixUser;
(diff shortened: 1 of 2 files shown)


src/bridge/AdminRoomHandler.ts
```diff
@@ -731,4 +731,34 @@ export class AdminRoomHandler {
         }
         return new MatrixAction("notice", null, body + "</ul>");
     }
+
+    // will mutate args if sucessful
```
Half-Shot, line 735: I'll allow it for convenience, but usually this is a bit scary

Half-Shot (approved):
Very fancy, no complaints here.

src/bridge/AdminRoomHandler.ts
tadzik: Yeah, I'm not super-fond of it either, but convenience won the moment I learned that Typescript won't allow me to redefine `args` in a function like so:

```typescript
public async foo(args: string[]) {
    const [server, args] = this.extractServerFromArgs(args); // nope
    ...
}
```

I can't think of anything that'd be both nice to use like the above but not crazy action-at-a-distance like it currently is.

PR #1442, opened 2021-08-02
Add bridge prefix to IRC connection time metric

Half-Shot:


tadzik (approved):
LGTM.

May be worth adding this flavour of a timer to m-a-b so that we're not tempted to roll or own in the future.

Half-Shot:
Yeah, we need to upstream the customisable bucket sizes option.

Issue #1209, opened 2020-12-22
Double bridging is hampered by username uniqueness attempts


turt2live:
`@telegram_406edacted:t2bot.io failed to get a IRC connection. Kicking from room: Error: Ran out of entries: telegram40`

This leads to generic 'IRC Connection Failed' errors, effectively ruining the bridge's attempt to operate. We might need to do different detection on the usernames to give them sensible names as bridges commonly use similar prefixes for users, such as the Discord bridge using the guild ID or Telegram having huge swathes of numbered users

ftyers:
I'm getting the same problem:
```
@appservice-irc:matrix.org ha fet fora a thinkl33t (@_discord_126751079089045504:t2bot.io). Raó: IRC connection failure
```

I got the following feedback:
> that would be a bug with the IRC bridge - it doesn't do a very good job of picking usernames for users, so it thinks it has run out (it shortens to _discord_126 or so on IRC, and so it thinks it has run out of numbers to increment)

Evidlo:
I think this is the snippet in question from [IdentGenerator.js](https://github.com/matrix-org/matrix-appservice-irc/blob/develop/src/irc/IdentGenerator.ts#L206-L222)

``` javascript
    private async getSuffixForUsername(username: string, domain: string) {
        let suffixLength = 1;
        while (suffixLength <= Math.log10(IdentGenerator.MAX_USER_NAME_SUFFIX) + 1) {
            const prefix = username.substring(
            // myusername becomes myuserna
                0, username.length - IdentGenerator.USER_NAME_DELIMITER.length - suffixLength);
            // Look for all usernames starting with myuserna
            const countForPrefix = await this.dataStore.getCountForUsernamePrefix(domain, prefix);
            // is there myuserna_1 to myuserna_9
            if (countForPrefix < Math.pow(10, suffixLength) - 1) {
                // Take the next available suffix.
                return countForPrefix;
            }
            suffixLength++;
        }
        throw Error("Ran out of entries: " + username);
    }
```

So these lines are responsible for trimming the MXID length (

``` javascript
const prefix = username.substring(
// myusername becomes myuserna
    0, username.length - IdentGenerator.USER_NAME_DELIMITER.length - suffixLength);
```

However, the next bit checks for collisions with this trimmed name

``` javascript
// Look for all usernames starting with myuserna
const countForPrefix = await this.dataStore.getCountForUsernamePrefix(domain, prefix);
[...]

PR #1446, opened 2021-08-03
Do not generate a unique username for IPv6 bridges

### LINKED ISSUE #1445: Messages from libera.chat are not shown in a room with other bridges
First of all, thank you so much to the project contributors and maintainers.  Our project (github.com/mlpack/mlpack) runs its chat infrastructure on Matrix and all of the bridges are wonderful, as they let our users use their preferred mode of communication.

**Describe the bug**

Recently we bridged `#mlpack:matrix.org` into `#mlpack` on irc.libera.chat.  However, it seems that messages from irc.libera.chat are not always sent to `#mlpack:matrix.org`.

Fascinatingly, we did observe one user who came into the libera.chat room (to set up a logging system, on whitequark.org) and their messages were sent to Matrix.  But, other libera.chat users' messages are not sent to Matrix, and it's not clear why.  We've tried with different non-colliding usernames.

`support@matrix.org` suggested that this could be a manifestation of #1209, but I'm not entirely sure if that applies in this case.

**To Reproduce**
1. Join `#mlpack:matrix.org`.
2. Also fire up an IRC client and join `#mlpack` on libera.chat.
3. Send a message
4. Don't see message in Matrix

 - OS: Debian (though my personal OS probably does not make the difference here)
 - Browser: Firefox
 - Version: 78.11.0esr

**Smartphone (please complete the following information):**
 - Device: Pixel 4a 5G  (also probably does not make the difference here)
 - OS: LineageOS :smile:
 - I'm using the Element app

Thanks so much for any help! :+1:

Half-Shot:
Fixes https://github.com/matrix-org/matrix-appservice-irc/issues/1209
Fixes #1445 

This PR makes it so that bridges that exclusively use IPv6 addresses for users will no longer attempt to generate unique usernames, instead favouring to pick a sanitised version of the userID. This should help cases where unique usernames were hard to generate (E.g. Slack / TG identifiers)

### DIFF
--- a/changelog.d/1446.misc
+++ b/changelog.d/1446.misc
@@ -1,0 +1,1 @@
+Do not generate a unique username for users on a IPv6 bridge, as it's unnessacery.
--- a/spec/unit/IdentGenerator.spec.js
+++ b/spec/unit/IdentGenerator.spec.js
@@ -15,6 +15,7 @@ describe("Username generation", function() {
     let ircClientConfigsUsernames;
     const serverMock = {
         getRealNameFormat: () => "mxid",
+        getIpv6Only: () => false,
     };
 
     beforeEach(function() {
@@ -68,6 +69,7 @@ describe("Username generation", function() {
         var userId = "@myreallylonguseridhere:localhost";
         const info = await identGenerator.getIrcNames(ircClientConfig, {
             getRealNameFormat: () => "reverse-mxid",
+            getIpv6Only: () => false
         }, new MatrixUser(userId));
         expect(info.realname).toEqual("localhost:myreallylonguseridhere");
     });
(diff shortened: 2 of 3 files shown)


tadzik:
Looks sensible – I do wonder if we should try to do our best for IPv4 anyway. Do you have some (anonymized) sample of names that ended up in conflicts on the IRC side? Maybe we could have a TODO test for it or something.

Half-Shot:
>Do you have some (anonymized) sample of names that ended up in conflicts on the IRC side? Maybe we could have a TODO test for it or something.

You could generate a bunch of random Slack IDs for this:

`@_slack_1115551:matrix.org`
`@_slack_11115552:matrix.org`
`@_slack_11115553:matrix.org`
....

I'm minded *not* to fix this in this PR, as it's a relatively difficult problem to solve and only really hurts those running large enough bridges on IPv4 (not something we'd recommend you do).

I can raise an issue describing the  problem in the hopes that we come up with a solution to generate unique usernames for IPv4 bridges.

tadzik:
> I'm minded _not_ to fix this in this PR, as it's a relatively difficult problem to solve and only really hurts those running large enough bridges on IPv4 (not something we'd recommend you do).
> 
> I can raise an issue describing the problem in the hopes that we come up with a solution to generate unique usernames for IPv4 bridges.

Yeah, that sounds good. I'm not advocating for fixing it now, just to not lose the context and the problem that can arise for some :)

tadzik (approved):

PR #1448, opened 2021-08-03
Add remote_ghosts_max metric

### LINKED ISSUE #1440: Export max irc client limit as metric
So we can easily tell when any bridge is about to hit it's limit

Half-Shot:
Fixes #1440

### DIFF
--- a/src/bridge/IrcBridge.ts
+++ b/src/bridge/IrcBridge.ts
@@ -408,6 +408,12 @@ export class IrcBridge {
             labels: ["version"],
         }).inc({ version: getBridgeVersion()}, 1);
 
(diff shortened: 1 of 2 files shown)


src/bridge/IrcBridge.ts
```diff
@@ -418,6 +424,7 @@ export class IrcBridge {
                     {server: server.domain},
                     mxMetrics["connection_failure_kicks"] || 0
                 );
+                maxRemoteGhosts.set({server: server.domain}, server.getMaxClients());
```
Half-Shot, line 427: We set this inside the collector because hot reloading means the value can change.

tadzik (approved):

PR #1453, opened 2021-08-10
Support matrix-appservice-bridge@3.1.0

Half-Shot:
This factors in changes from the upcoming matrix-appservice-bridge 3.0.0 release


tadzik (approved):
Looks good \o/

PR #1455, opened 2021-08-16
Remove index for unique usernames

Half-Shot:
We didn't remove this index when working on #1446, which would mean that the username would still be treated as unique. There should be supporting code to ensure that we don't create duplicate usernames.


tadzik (approved):
LGTM

PR #1458, opened 2021-08-17
Update documentation and remove HOWTO.md content

Half-Shot:
We really want to encourage use of the documentation, to reduce the amount of places we put information. This change does the following:

- Updates our docker badge to be more useful
- Removes references to IRC networks where appropriate with links to the networks page
- A few wording improvements here and there


tadzik (approved):
Looks good to me :+1:

PR #1459, opened 2021-08-17
Make sure we don't exceed the line limit when trimming long messages

tadzik:
getMaxLineLength() doesn't actually include the channel name in its
calculations (though it does add required spaces around it), so we need
to take care of that bit ourselves here.

The API could be a bit more helpful here perhaps, but this at least
fixes the issue for now.

### DIFF
--- a/src/bridge/MatrixHandler.ts
+++ b/src/bridge/MatrixHandler.ts
@@ -1132,7 +1132,7 @@ export class MatrixHandler {
                 const explanation = renderTemplate(this.config.truncatedMessageTemplate, { url: httpUrl });
                 let messagePreview = trimString(
                     potentialMessages[0],
-                    ircClient.getMaxLineLength() - 4 /* "... " */ - explanation.length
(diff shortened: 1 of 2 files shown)


src/bridge/MatrixHandler.ts
```diff
@@ -1132,7 +1132,7 @@ export class MatrixHandler {
                 const explanation = renderTemplate(this.config.truncatedMessageTemplate, { url: httpUrl });
                 let messagePreview = trimString(
                     potentialMessages[0],
-                    ircClient.getMaxLineLength() - 4 /* "... " */ - explanation.length
+                    ircClient.getMaxLineLength() - 4 /* "... " */ - explanation.length - ircRoom.channel.length
```
Half-Shot, line 1135: This looks fine, are we sure this is the only place it is needed?

Half-Shot (approved):

src/bridge/MatrixHandler.ts
tadzik: It's the only place where we use `getMaxLineLength()` so it's probably the only wrong calculation thereof :)

Issue #490, opened 2017-08-16
m.notice is excempt from automatic pastebinning


Mikaela:
When user sends more than three lines, their message gets pastebinned. If a bot does the same, the messages go through triggering spam protections (https://github.com/matrix-org/matrix-appservice-irc/issues/471#issuecomment-322770808).

deepbluev7:
So, we accidentally triggered this, when a bot sent a notice containing the output of the unix "yes" command. This led to half an hour of IRC spam. I don't think that is that good of a behaviour, since it will easily get the bridge banned.

PR #1461, opened 2021-08-18
Make sure notices are also pastebinned when too long for IRC

tadzik:
Fixes GH-490.

### DIFF
--- a/changelog.d/1461.bugfix
+++ b/changelog.d/1461.bugfix
@@ -1,0 +1,1 @@
+Make sure notices are also pastebinned when too long for IRC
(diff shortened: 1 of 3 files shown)


changelog.d/1461.bugfix
```diff
@@ -0,0 +1 @@
+Make sure notices are also pastebinned when too long for IRC
```
Half-Shot, line 1: Little more detail
```suggestion
Make sure Matrix notice messages are also pastebinned when they exceed the line limit for IRC.
```

Half-Shot (approved):
Seems alright to me

Issue #481, opened 2017-08-02
Error "data too large for key size" when storing a password


Spydar007:
When attempting to `!storepass`, sometimes the password fails to store with the following error:

`Failed to store password: error:040A006E:rsa routines:RSA_padding_add_PKCS1_OAEP_mgf1:data too large for key size`

I have no idea why this is caused. A 73-character password stores fine for one network, but a different 73 character password on a different network does not. An 82 character password fails, but a 74 is fine. 

Freenode and Snoonet store fine. 
I have this issue with EsperNet, OFTC and Rizon.

There doesn't seem to be any sort of concrete reason for this (i.e. password over a set number of characters), but I haven't investigated much.

PR #1462, opened 2021-08-18
Migrate to symmetric encryption for passwords

3nprob:
I may be missing some context on why it was decided to use an RSA private key for symmetrically encrypting a database field.

While it's perfectly secure, it has drawbacks over AES:

* RSA is very slow/CPU intensive compared to cryptographic security
* It can't encrypt messages longer than the key length (manifesting in #481)
* PEM private keys are large and more cumbersome to manage than terse secret keys.

This changes the encryption scheme from RSA to AES-256-GCM.

To dodge implementing key rotation logic and make this change transparent for users, the AES key is constructed from the hash of the existing private key. Decryption of both modes works transparently.

The real motivation for this change is to facilitate introducing a new encrypted column for per-user SASL keys.,


tadzik:
Seems fine to me – I wonder if we want to actually migrate the old passwords though, to be able to cut out that code path. @Half-Shot?

3nprob:
@tadzik My 5c: Yes, ideally:

* Passwords should be migrated
* `passkey` in config  should be deprecated and replaced with an actual AES key (better inlined than as a separate file)

However;

* Switching to AES key in config will be a breaking change and need user intervention (either logging the secret key or having the process rewriting config under the user are Bad)
* Migration code will be a bigger and more change than this PR

So IMO it would be best to treat this in isolation from that and address migration/changing config in separate PR(s) to not have this be blocked.

3nprob:
Completely separate from the above, but a CLI utility for re-encrypting passwords with a new key will also be generally useful to do rotation of the encryption key, either as part of a security routine, or if the encryption key is suspected to be compromised.

andir:
Just my two cents on this (and they are really minor): We should not require inlining secrets into the configuration files. I'd much rather have those as a separate file that can be loaded and managed like a secret. I treat configuration files as mostly public in my deployments.

3nprob:
@tadzik @Half-Shot @jaller94  How would you like to move forward with this?

Issue #1026, opened 2020-04-10
Messages edits are too verbose on the IRC side


a-andreyev:
Hello! Sorry if a duplicate or not an issue, feel free to close or lead me to some other place to discuss it.

It could be very inconvenient for the IRC users to read my matrix messages if I'm using edits a lot during composing my sentences. With the protocols that support edits, I've used to re-read my messages and fix minor typos or rephrase my thoughts better while the message is not answered (especially at the english speaking groups since I'm not a native speaker).

AFAIK, since matrix supports edits on a protocol level, while IRC is not, my approach in composing the messages is very verbose on the IRC side. 

Of course, I could be more careful and re-check is the room connected to services that not support edits natively. Or try to change my messaging approach. But I'm afraid I'm not the only one who is typing like this.

Any hints on how could I contribute to solving the issue? :) Maybe it could be solved with some timer and text diff logic on the app services side or some warning on the client side?

Half-Shot:
Hi, currently the bridge isn't "edit aware" and it will faithfully send your edits as real messages (edits are just normal events with a '* ' on the front, with some metadata to link back to the original so smart clients can update the original).

I'm not adverse to the timer method, however I caution that this is unlikely to work in a federated context or when the service is just slow. For instance, sticking a 5 second timer on messages going out from Matrix would make a slow service even more slow, while still perhaps not being all that beneficial (for instance, most edits do not happen within the first 5 seconds).

At the moment, resending the message with a '*' is probably the best we can do for IRC because it allows the bridge to still be quick without causing too much spam on the IRC side. The only silver lining I could offer is perhaps underlining which bit of the message has changed, so IRC users don't need to re-read the whole message.

ljl-covid:
This really makes the bridge a huge liability in my channel... Just now a user sent a message long enough that it took up two IRC messages (which are 512-byte length limited), and then, they edited it *twice*, which resulted in a total of 6 messages on IRC, all quite long.

I understand you can only do what you can do, but I felt the need to "pile on" because this really is a huge usability issue from my point of view and those of the users in my channel, and I also wanted to point out how it interacts with long messages.

ljl-covid:
I would like to suggest a solution for at least one particular use case: long messages.

![longmessagespam](https://user-images.githubusercontent.com/62215844/92814182-0964e180-f3c3-11ea-8952-67b8a9529fc6.png)

[...]

PR #1465, opened 2021-08-18
Make message edits more compact when possible

tadzik:
I'll have the tests do the talking:

    [          
        'should generate a diff for short, multiline messages',          
        'one\nfoo\nthree', 'one\ntwo\nthree',          
        's/foo/two/',          
    ],          
    [          
        'should generate sed-like substitution when a short part of the message changes',          
        "Sounds good – I'll be there before 9", "Sounds good – I'll be there after 9",          
        's/before/after/'          
    ],          
    [          
        'should only show changes from the line that has changed in multiline messages',          
        'in a marmalade forest\nbetween the make-believe trees\nI forgot the third verse, sorry',          
        'in a marmalade forest\nbetween the make-believe trees\nin a cottage-cheese cottage...',          
        's/I forgot the third verse, sorry/in a cottage-cheese cottage.../',          
    ],
        [                   
        'should only show small portion of the message when a new word is added',          
        'this is a message where i a word', 'this is a message where i missed a word',          
        '* where i missed a word',          
    ],                  
    [                   
        'should only show small portion of the message when a new word is added at the beginning',          
        'get lunch now, be back a bit later', 'gonna get lunch now, be back a bit later',          
        '* gonna get lunch',          
    ],                  
    [                   
        'should only show small portion of the message when a new word is added at the beginning',          
        "I'm gonna get lunch now", "I'm gonna get lunch now, bbl",          
        '* lunch now, bbl',          
    ],                  
    [                   
        'should show word removals as s/foo//',          
        'I gotta go clean up my filthy room', 'I gotta go clean up my room',          
        's/filthy//'          
    ],                  

It tries its best, and falls back to the current behaviour if the result of generating "seds" would be considered worse.

I don't have a solution for pastebinned messages being edited: the URL will change with each edit, and some indication of what's changed is desirable. It gets *somewhat* improved for free with https://github.com/matrix-org/matrix-appservice-irc/pull/1430, which makes the leading part of the edited message shown before the URL, so it doesn't look like *total* link spam.

Fixes GH-1026.

### DIFF
--- a/changelog.d/1465.feature
+++ b/changelog.d/1465.feature
@@ -1,0 +1,1 @@
+Make message edits more compact when possible
--- a/changelog.d/1465.feature
+++ b/changelog.d/1465.feature
@@ -1,1 +1,1 @@
-Make message edits more compact when possible
+Render matrix message edits as sed-like diff statements, falling back to asterisk formatted messages
--- a/spec/integ/matrix-to-irc.spec.js
+++ b/spec/integ/matrix-to-irc.spec.js
@@ -392,7 +392,7 @@ describe("Matrix-to-IRC message bridging", function() {
                 expect(client.nick).toEqual(testUser.nick);
                 expect(client.addr).toEqual(roomMapping.server);
                 expect(channel).toEqual(roomMapping.channel);
-                expect(text).toEqual(`<${repliesUser.nick}> "This" <- Reply Text`);
+                expect(text).toEqual(`<${repliesUser.nick}> "This..." <- Reply Text`);
             }
         );
         const formatted_body = constructHTMLReply(
--- a/src/bridge/MatrixHandler.ts
+++ b/src/bridge/MatrixHandler.ts
@@ -19,6 +19,7 @@ import { AdminRoomHandler } from "./AdminRoomHandler";
 import { trackChannelAndCreateRoom } from "./RoomCreation";
 import { renderTemplate } from "../util/Template";
 import { trimString } from "../util/TrimString";
+import { niceDiff } from "../util/NiceDiff";
 
 async function reqHandler(req: BridgeRequest, promise: PromiseLike<unknown>) {
     try {
@@ -1058,6 +1059,8 @@ export class MatrixHandler {
         }
 
         let cacheBody = ircAction.text;
+
+        // special handling for replies
         if (event.content["m.relates_to"] && event.content["m.relates_to"]["m.in_reply_to"]) {
             const eventId = event.content["m.relates_to"]["m.in_reply_to"].event_id;
             const reply = await this.textForReplyEvent(event, eventId, ircRoom);
@@ -1066,6 +1069,35 @@ export class MatrixHandler {
                 cacheBody = reply.reply;
             }
         }
+
+        // special handling for edits
+        if (event.content["m.relates_to"] && event.content["m.relates_to"].rel_type === "m.replace") {
+            const originalEventId = event.content["m.relates_to"].event_id;
+            let originalBody = this.getCachedEvent(originalEventId)?.body;
+            if (!originalBody) {
+                try {
+                    // FIXME: this will return the new event rather than the original one
+                    // to actually see the original content we'd need to use whatever
+                    // https://github.com/matrix-org/matrix-doc/pull/2675 stabilizes on
+                    const eventContent = await this.ircBridge.getAppServiceBridge().getIntent().getEvent(
+                        event.room_id, originalEventId
+                    );
+                    originalBody = eventContent.content.body;
+                }
+                catch (_err) {
+                    req.log.warn("Couldn't find an event being edited, using fallback text");
+                }
+            }
+            req.log.debug(JSON.stringify(event.content));
+            const newBody = event.content["m.new_content"]?.body;
+            if (originalBody && newBody) {
+                const diff = niceDiff(originalBody, newBody);
+                if (diff) {
+                    ircAction.text = diff;
+                }
+            }
+        }
+
         let body = cacheBody.trim().substring(0, this.config.replySourceMaxLength);
         const nextNewLine = body.indexOf("\n");
         if (nextNewLine !== -1) {
@@ -1073,7 +1105,7 @@ export class MatrixHandler {
         }
         // Cache events in here so we can refer to them for replies.
         this.cacheEvent(event.event_id, {
-            body,
+            body: cacheBody,
             sender: event.sender,
             timestamp: event.origin_server_ts,
         });
--- a/src/bridge/MatrixHandler.ts
+++ b/src/bridge/MatrixHandler.ts
@@ -1071,7 +1071,7 @@ export class MatrixHandler {
         }
 
         // special handling for edits
-        if (event.content["m.relates_to"] && event.content["m.relates_to"].rel_type === "m.replace") {
+        if (event.content["m.relates_to"]?.rel_type === "m.replace") {
             const originalEventId = event.content["m.relates_to"].event_id;
             let originalBody = this.getCachedEvent(originalEventId)?.body;
             if (!originalBody) {
--- a/src/bridge/MatrixHandler.ts
+++ b/src/bridge/MatrixHandler.ts
@@ -1088,7 +1088,6 @@ export class MatrixHandler {
                     req.log.warn("Couldn't find an event being edited, using fallback text");
                 }
             }
-            req.log.debug(JSON.stringify(event.content));
             const newBody = event.content["m.new_content"]?.body;
             if (originalBody && newBody) {
                 const diff = niceDiff(originalBody, newBody);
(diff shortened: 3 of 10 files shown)


src/util/NiceDiff.ts
```diff
@@ -0,0 +1,60 @@
...
+        i++;
+    }
+
+    return changes.map(c => `s/${c[0]}/${c[1]}/`);
+}
+
+// tries to find a sensible representation of a message edit
+// returns undefined it it can't come up with anything better than
+// "just post the new message in its entirety"
+export function niceDiff(from: string, to: string): string|undefined {
```
tadzik, line 36: I'm open to suggestions for the better name for this :P

src/util/NiceDiff.ts
```diff
@@ -0,0 +1,60 @@
...
+    // don't bother if the message is short enough
+    if (to.length < 20 && !to.match(/\n/)) {
```
Half-Shot, line 38: Can the `20` be made into a const please :)

src/util/NiceDiff.ts
```diff
@@ -0,0 +1,60 @@
...
+        return undefined;
+    }
+
+    const changesets = [
+        formatChanges(Diff.diffWords(from, to)),
+        formatChanges(Diff.diffLines(from, to)),
+    ].filter(
+        // too many substitutions aren't that readable
+        diffs => diffs.length > 0 && diffs.length <= 3
```
Half-Shot, line 47: Same for the `3` here.

src/util/NiceDiff.ts
```diff
@@ -0,0 +1,60 @@
...
+            changes.push([diff[i].value, diff[i+1].value]);
+        }
+        i++;
+    }
+
+    return changes.map(c => `s/${c[0]}/${c[1]}/`);
+}
+
+// tries to find a sensible representation of a message edit
+// returns undefined it it can't come up with anything better than
```
Half-Shot, line 34:
```suggestion
// returns undefined if it can't come up with anything better than
```

src/util/NiceDiff.ts
Half-Shot: `generateTextualDiff` would be an improvement in my eyes

src/util/NiceDiff.ts
```diff
@@ -0,0 +1,60 @@
...
+// "just post the new message in its entirety"
```
Half-Shot, line 35: If you're going to add some comments, do one better and make it a docstring :innocent:

src/util/NiceDiff.ts
```diff
@@ -0,0 +1,60 @@
...
+export function niceDiff(from: string, to: string): string|undefined {
+    // don't bother if the message is short enough
+    if (to.length < 20 && !to.match(/\n/)) {
+        return undefined;
+    }
+
+    const changesets = [
+        formatChanges(Diff.diffWords(from, to)),
```
Half-Shot, line 43: Could somebody potentially exploit us and make a billion diffs appear, clogging the bridge?

src/bridge/MatrixHandler.ts
```diff
@@ -1066,14 +1069,43 @@ export class MatrixHandler {
...
+                    const eventContent = await this.ircBridge.getAppServiceBridge().getIntent().getEvent(
+                        event.room_id, originalEventId
+                    );
+                    originalBody = eventContent.content.body;
+                }
+                catch (_err) {
+                    req.log.warn("Couldn't find an event being edited, using fallback text");
+                }
+            }
+            req.log.debug(JSON.stringify(event.content));
```
Half-Shot, line 1091: Hello hello random debug statement

changelog.d/1465.feature
```diff
@@ -0,0 +1 @@
+Make message edits more compact when possible
```
Half-Shot, line 1: I'd go a bit further here: `Render matrix message edits as diff statements, falling back to asterisk formatted messages` or some such

spec/integ/matrix-to-irc.spec.js
```diff
@@ -392,7 +392,7 @@ describe("Matrix-to-IRC message bridging", function() {
                 expect(client.nick).toEqual(testUser.nick);
                 expect(client.addr).toEqual(roomMapping.server);
                 expect(channel).toEqual(roomMapping.channel);
-                expect(text).toEqual(`<${repliesUser.nick}> "This" <- Reply Text`);
+                expect(text).toEqual(`<${repliesUser.nick}> "This..." <- Reply Text`);
```
Half-Shot, line 395: Looks like a reply to me but I trust you :)

src/bridge/MatrixHandler.ts
```diff
@@ -1066,14 +1069,43 @@ export class MatrixHandler {
...
+        // special handling for edits
+        if (event.content["m.relates_to"] && event.content["m.relates_to"].rel_type === "m.replace") {
+            const originalEventId = event.content["m.relates_to"].event_id;
+            let originalBody = this.getCachedEvent(originalEventId)?.body;
+            if (!originalBody) {
+                try {
+                    // FIXME: this will return the new event rather than the original one
+                    // to actually see the original content we'd need to use whatever
+                    // https://github.com/matrix-org/matrix-doc/pull/2675 stabilizes on
+                    const eventContent = await this.ircBridge.getAppServiceBridge().getIntent().getEvent(
```
Half-Shot, line 1082: This might fail for DMs, as the bot can't read events in rooms it's not part of. You might have to determine the other user in a DM room and get an intent for them.

src/bridge/MatrixHandler.ts
```diff
@@ -1066,14 +1069,43 @@ export class MatrixHandler {
                 cacheBody = reply.reply;
             }
         }
+
+        // special handling for edits
+        if (event.content["m.relates_to"] && event.content["m.relates_to"].rel_type === "m.replace") {
```
Half-Shot, line 1074: I forget if TS likes these?

```suggestion
        if (event.content.["m.relates_to"]?.rel_type === "m.replace") {
```

Half-Shot (approved):
A few comments here and there on small things, but this looks great overall!

src/bridge/MatrixHandler.ts
tadzik: Gah, I thought I got them all >:(

spec/integ/matrix-to-irc.spec.js
tadzik: It is :) I sneaked in a bugfix [here](https://github.com/matrix-org/matrix-appservice-irc/pull/1465/files#diff-1461a1072635fdffa4ecfbfebe8aad698b2c3a406dd690a9546ff1f047fed580L1076) that makes the event cache actually remember the entire contents rather than just the first line – that's why this not includes the `...`, since it now properly remembers that the message that we're replying to didn't end there

src/bridge/MatrixHandler.ts
tadzik: Seems like it does – doesn'teven need a `.` Fixed in b7df6011

changelog.d/1465.feature
tadzik: Improved in 8ba51429

src/bridge/MatrixHandler.ts
tadzik: Fixed in c26080a4

src/util/NiceDiff.ts
tadzik: Did that in 627210a3 :)

src/util/NiceDiff.ts
tadzik: Obsoleted by 627210a3

src/util/NiceDiff.ts
tadzik: Also fixed in 093740bd

src/util/NiceDiff.ts
tadzik: Yep :) Fixed in 093740bd

src/bridge/MatrixHandler.ts
tadzik: Good call. Is there any reason we shouldn't just always use the user's intent here?

tadzik: Fixed, and answered my own question in 021d29aa :)

src/util/NiceDiff.ts
```diff
@@ -0,0 +1,73 @@
...
+const MAX_SUBSTITUTIONS = 3;
+
+/**
+ * Try to find a sensible representation of a message edit,
+ * or returns undefined if it deems posting the entire new message
+ * to be a better choice. Optimize for terseness, legibility
+ * and an IRC-native feel.
+ *
+ * @param {string} from : The original message
+ * @param {string} to : The new, edited version
```
Half-Shot, line 45: Excessive : colons

```suggestion
 * @param {string} from The original message
 * @param {string} to The new, edited version
```

src/util/NiceDiff.ts
tadzik: Went with `messageDiff` to describe the intent better.

Xiretza:
This should have a check to make sure the produced diff is shorter than just using the message body (as before).

Issue #1466, opened 2021-08-19
Inviting the first matrix user to a channel makes the bot join, but the invite is not sent


tadzik:
1. Create a new IRC channel
2. Invite a bridged matrix user
3. See matrix bot join
4. Bridged user sees no invite on matrix

**Expected behavior**
User should receive an invite straight away.

**Additional context**
Consecutive invites arrive just fine once the bot's on the channel.

PR #1467, opened 2021-08-19
Fallback to sending an invite as a bot if the regular invite fails

tadzik:
This works around an issue where an invite is sent before the room is
fully configured, and the power level of the actual inviter is not yet
know and set.

Fixes GH-1466.

### DIFF
--- a/src/bridge/IrcHandler.ts
+++ b/src/bridge/IrcHandler.ts
@@ -406,7 +406,15 @@ export class IrcHandler {
                 virtualMatrixUser.getId()
             ).invite(
                 room.getId(), invitee
-            );
+            ).catch(err => {
+                req.log.warn(
+                    `Failed to invite %s as the inviter user (reason: ${err}),
+                    inviting as a bot as fallback`
+                );
+                // TODO Note the original inviter in the reason
+                // once MSC 2367 stabilizes and is implemented by the SDK
+                return this.ircBridge.getAppServiceBridge().getIntent().invite(room.getId(), invitee);
+            });
         });
         await Promise.all(invitePromises);
         return undefined;
--- a/src/bridge/IrcHandler.ts
+++ b/src/bridge/IrcHandler.ts
@@ -408,7 +408,7 @@ export class IrcHandler {
                 room.getId(), invitee
             ).catch(err => {
                 req.log.warn(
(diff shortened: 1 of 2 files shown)


src/bridge/IrcHandler.ts
```diff
@@ -406,7 +406,15 @@ export class IrcHandler {
                 virtualMatrixUser.getId()
             ).invite(
                 room.getId(), invitee
-            );
+            ).catch(err => {
+                req.log.warn(
+                    `Failed to invite %s as the inviter user (reason: ${err}),
```
Half-Shot, line 411: You didn't provide an arg to %s, but we might as well use new template syntax anyway.
```suggestion
                    `Failed to invite ${invitee} as the inviter user (reason: ${err}),
```

src/bridge/IrcHandler.ts
```diff
@@ -406,7 +406,15 @@ export class IrcHandler {
...
+                    inviting as a bot as fallback`
+                );
+                // TODO Note the original inviter in the reason
```
Half-Shot, line 414: We could just set some room state here tbh, I don't think we should wait for the MSC to arrive :older_man:

Half-Shot (changes requested):

src/bridge/IrcHandler.ts
tadzik: What could we set in a room state that'll convey the relevant info?

Half-Shot: Oh, I mean literally send a m-room-member event with a membership of invite and a reason of whatever.

tadzik: Ah yes, that works. Though currently Synapse seems to just strip the reason anyway, so it doesn't actually show up – but we should at least be future-proof with this :) Done in 2e80b23e.

Half-Shot: :roll_eyes: synaaaaapse. Is there a bug for that? Otherwise, sounds good.

Half-Shot (approved):

src/bridge/IrcHandler.ts
tadzik: (Seems like it does not strip it after all, all works well, so this is all good).

PR #1472, opened 2021-09-02
Implement MAB's UserActivityTracker and blocking the bridge upon exceeding the user limit

tadzik:
Developed in tandem with https://github.com/matrix-org/matrix-appservice-bridge/pull/350

### DIFF
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -569,6 +569,12 @@ ircService:
 
   # Maximum number of montly active users, beyond which the bridge gets blocked (both ways)
   # RMAUlimit: 100
+  
+  # userActivity:
+    # The "grace period" before we start counting users as active
+    # minUserActiveDays: 1
+    # Time before users are considered inactive again
+    # inactiveAfterDays: 30
 
   ircHandler:
     # Should we attempt to match an IRC side mention (nickaname match)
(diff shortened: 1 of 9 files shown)


config.sample.yaml
```diff
@@ -556,6 +556,9 @@ ircService:
     # how much time needs to pass between the reply and the original message to switch to the long format
     shortReplyTresholdSeconds: 300
 
+  # Maximum number of montly active users, beyond which the bridge gets blocked (both ways)
+  # RMAUlimit: 100
```
Half-Shot, line 560: Needs a JSON schema change too :stuck_out_tongue:

Half-Shot (approved):
No concerns about this one

Half-Shot:
I believe we need #1453 for this to work?

tadzik:
> I believe we need #1453 for this to work?

I guess so – though locally I've also backported the MAB PR to the 2.6.1 that IRC bridge currently relies on, and it works too – we can use that if we're in a hurry.

config.sample.yaml
tadzik: Done in 26505376 :)

Half-Shot (approved):

PR #1478, opened 2021-09-14
Fix help not showing admin commands when the user is an admin

Half-Shot:
This was caused by us passing in the sender rather than the permission of the user. 

I've fixed this by using an enum type for permissions, so we don't make the mistake of passing in a string and expecting it to work.


tadzik (approved):
Mmm proper types

PR #1479, opened 2021-09-15
Fallback to generic profile displayname if we can't get the bot to join the proper room

tadzik:
This happens if a matrix users PMs an IRC user
while not being in any channel.


Half-Shot (approved):
Looks good (as per call)

Needs changelog pleeeease

tadzik:
> Needs changelog pleeeease

Should be good now :)

PR #1494, opened 2021-11-02
Make sure message diffs are not longer than the new message would be

### LINKED ISSUE #1477: New message edit logic makes some edits more verbose than before
**Describe the bug**
When a message like this:

```
Lorem ipsu dolor sit - an arbitrary amount of trailing text will be duplicated in the sed expression, even though it should only include a few words of context
```

is edited to this:

```
Lorem ipsum dolor sit amet - an arbitrary amount of trailing text will be duplicated in the sed expression, even though it should only include a few words of context
```

The following message is posted to IRC:

```
s/Lorem ipsu dolor sit - an arbitrary amount of trailing text will be duplicated in the sed expression, even though it should only include a few words of context/Lorem ipsum dolor sit amet - an arbitrary amount of trailing text will be duplicated in the sed expression, even though it should only include a few words of context/
```

This is much longer than what would have been produced before #1465.

**To Reproduce**
1. Post the first message above to an IRC-bridged matrix room.
2. Edit the message to resemble the second message above
3. Observe the overly long edit message in IRC

**Expected behavior**
A shorter edit summary should be produced - maybe something like `* Lorem ipsum dolor sit amet - an arbitrary`.

Note: this was observed in a libera.chat room.

tadzik:
Partially fixes #1477 – better diff detection would be better eventually, but this at least reduces the annoyance now

### DIFF
--- a/changelog.d/1477.bugfix
+++ b/changelog.d/1477.bugfix
@@ -1,0 +1,1 @@
+Make sure message diffs are not longer than the new message would be
--- a/changelog.d/1477.bugfix
+++ b/changelog.d/1477.bugfix
@@ -1,1 +1,1 @@
-Make sure message diffs are not longer than the new message would be
+Matrix message edits no longer bridge as a diff if it's longer than the new message
--- a/spec/unit/MessageDiff.spec.js
+++ b/spec/unit/MessageDiff.spec.js
@@ -26,7 +26,7 @@ describe('messageDiff', function() {
         ],
         [
             'should not use diffs with newlines in them',
-            'a\nb\nc', 'bla\nbleh\nc',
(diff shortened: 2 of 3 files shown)


Half-Shot (approved):
Seems sensible to me :+1:

changelog.d/1477.bugfix
```diff
@@ -0,0 +1 @@
+Make sure message diffs are not longer than the new message would be
```
Half-Shot, line 1: Oh, could we explain this a bit more? "Matrix message edits no longer bridge as a diff if it's longer than the new message"

tadzik: I like it. Changed in 1c526cb7

Issue #1414, opened 2021-07-12
Reactions & typing notifications aren't implemented


Mikaela:
**Describe the bug**

When an IRC or Matrix user reacts onto something, the reaction doesn't appear on the other side.

**To Reproduce**

1. Go to a network that supports IRCv3 message-tags on both IRC  and Matrix-appservice IRC 
2. React onto a message or start typing
3. See that it doesn't go through.

**Expected behavior**

Matrix and IRC both know when the other side has someone typing and the reactions go through as well

**Additional context**

* BridgeVersion: 0.27.0
* develop.element.io
* emoji reactions are supported by at least IRCCloud and Palaver
* typing notifications are supported by IRCCloud and WeeChat (git)
* I am not sure if this is more of a feature request or a bug, but the bug template asks whether this is connected to a bug so I erred on the side of bug.
* Specs: https://ircv3.net/irc/#client-only-tags
* I was testing on `ircs://irc.liberta.casa:6697` / `#irc_#libcasa.info:liberta.casa`

Half-Shot:
Rewriting title to be a more positive spin, as it sounded like a bug.

Mikaela:
I added replies to this issue considering they are technically in the same group as reactions and typing notifications

PR #1503, opened 2022-01-10
Bridge Matrix threads

tadzik:
This treats each thread reply as an old message reply, as in: always quotes the previous message from a thread, regardless of how long ago it happened – with the possibility of multiple threads going on concurrently, short replies just turn into a mess.

image

It is quite noisy, but I think it's the best we can do to preserve the context. Replies from IRC won't work, obviously :(

### DIFF
--- a/src/bridge/MatrixHandler.ts
+++ b/src/bridge/MatrixHandler.ts
@@ -48,6 +48,8 @@ export interface MatrixHandlerConfig {
     shortReplyTemplate: string;
     // Format of replies sent a while after the original message
     longReplyTemplate: string;
+    // Maximum number of requests we'll perform when looking up previous message in a thread
+    threadLookupRequestLimit: number;
     // Format of the text explaining why a message is truncated and pastebinned
     truncatedMessageTemplate: string;
 }
@@ -58,6 +60,7 @@ const DEFAULTS: MatrixHandlerConfig = {
     shortReplyTresholdSeconds: 5 * 60,
     shortReplyTemplate: "$NICK: $REPLY",
     longReplyTemplate: "<$NICK> \"$ORIGINAL\" <- $REPLY",
+    threadLookupRequestLimit: 10,
     truncatedMessageTemplate: "(full message at $URL)",
 };
 
@@ -126,6 +129,9 @@ export class MatrixHandler {
     // required because invites are processed asyncly, so you could get invite->msg
     // and the message is processed before the room is created.
     private readonly eventCache: Map<string, CachedEvent> = new Map();
+    // keep track of the last message sent to each thread
+    // so that we can form replies to the most recent one rather than the first one
+    private lastMessageInThreadCache: Map<string, string> = new Map();
     private readonly metrics: {[domain: string]: {
         [metricName: string]: number;
     };} = {};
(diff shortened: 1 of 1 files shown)


src/bridge/MatrixHandler.ts
```diff
@@ -126,6 +129,9 @@ export class MatrixHandler {
     // required because invites are processed asyncly, so you could get invite->msg
     // and the message is processed before the room is created.
     private readonly eventCache: Map<string, CachedEvent> = new Map();
+    // keep track of the last message sent to each thread
+    // so that we can form replies to the most recent one rather than the first one
+    private lastMessageInThreadCache: Map<string, string> = new Map();
```
Half-Shot, line 134: Do we want to make this a LRU to avoid this growing ever larger, or nah?

src/bridge/MatrixHandler.ts
```diff
@@ -1046,13 +1052,55 @@ export class MatrixHandler {
...
+                if (reply !== null) {
+                    ircAction.text = reply.formatted;
+                    cacheBody = reply.reply;
+                }
+            } else if (event.content["m.relates_to"]?.rel_type === "io.element.thread") {
+                const threadId = event.content["m.relates_to"].event_id;
+                let lastReplyId = this.lastMessageInThreadCache.get(threadId);
+                if (!lastReplyId) {
+                    try {
+                        const getThread = async (from?: string) => this.ircBridge.getAppServiceBridge().getIntent().matrixClient.doRequest(
```
Half-Shot, line 1069: Might be worth chucking a PR at the bot sdk too, even if it will come too late for this PR.

src/bridge/MatrixHandler.ts
```diff
@@ -1046,13 +1052,55 @@ export class MatrixHandler {
...
+                                } else {
+                                    lastReplyId = thread['chunk'][currentMessageIdx + 1].event_id;
+                                }
+                                break;
+                            }
+                        }
+                    } catch (err) {
+                        console.error('Error fetching thread from homeserver:', err);
+                    }
+                    lastReplyId ||= threadId; // fallback to the original message if we really have no better idea
```
Half-Shot, line 1096: Personal feeling is this logic feels too complex to sit in this function. Should at least be it's own function, if not in the bot/bridge sdk itself.

Half-Shot (approved):
Code itself seems reasonable, I suppose the jury is out on whether it's *useful* to have this. I think on the whole it looks okay, but I am leaning towards disabling support in portal rooms because I think it could get messy if Matrix users start threading.

tadzik:
> I suppose the jury is out on whether it's _useful_ to have this

This is what it goes down to, yeah. If it'll end up the giving the false impression that we can transparently bridge threads, and yet we can't bridge replies from IRC to Matrix it'll just become a mess. Perhaps it'd make more sense to do channels-as-threads on the IRC side?

Mikaela:
> It is quite noisy, but I think it's the best we can do to preserve the context. Replies from IRC won't work, obviously :(

Wouldn't it be best to just suppoot [IRCv3 reply client tag](https://ircv3.net/specs/client-tags/reply) as per https://github.com/matrix-org/matrix-appservice-irc/issues/1414?

Half-Shot:
> > It is quite noisy, but I think it's the best we can do to preserve the context. Replies from IRC won't work, obviously :(
> 
> Wouldn't it be best to just support [IRCv3 reply client tag](https://ircv3.net/specs/client-tags/reply) as per #1414?

I realise it's a bit chicken and egg, but atm only IRCCloud supports it...so we're unlikely to do the work to support it until Server/Client impls catch up.

Mikaela:
> > > It is quite noisy, but I think it's the best we can do to preserve the context. Replies from IRC won't work, obviously :(
> > 
> > 
> > Wouldn't it be best to just support [IRCv3 reply client tag](https://ircv3.net/specs/client-tags/reply) as per #1414?
> 
> I realise it's a bit chicken and egg, but atm only IRCCloud supports it...so we're unlikely to do the work to support it until Server/Client impls catch up.

It's also supported by Limnoria and Palaver has expressed interest towards implementing it. On servers at least Ergo has it and I would think InspIRCd to at least have it planned if not implemented already.

progval:
> until Server/Client impls catch up.

Servers don't need to implement it, it's a [client-tag](https://ircv3.net/specs/extensions/message-tags#client-only-tags).

Half-Shot:
> > until Server/Client impls catch up.
> 
> Servers don't need to implement it. It's a [client-tag](https://ircv3.net/specs/extensions/message-tags#client-only-tags), and servers just pass all client tags without explicitly implementing them. (Though UnrealIRCd has a whitelist, but it already has the tag on its whitelist.)

Mm, though client-tags is still lagging behind a bit as well (https://github.com/solanum-ircd/solanum/issues/167). Granted, we could probably implement given enough servers do support it.

Mikaela:
I am under impresson that Solanum is going to get a rewrite at some point and is not going to implement a lot of IRCv3 features before it happens, so I wouldn't include them in considerations how widely supported something is.

PR #1506, opened 2022-01-24
Remove legacy group chat check

### LINKED ISSUE #1505: Reject room invite instead of join->leaving, when being invited to a group chat.
At this point, we can probably say that if an invite doesn't contain the `is_direct` metadata, then it's probably not going to work out.

Half-Shot:
Fixes #1505 

This will break behaviors for clients which don't support `is_direct`, but this flag has been around for eons now and rejecting invites from large group rooms can reduce load on the HS.

### DIFF
--- a/src/bridge/MatrixHandler.ts
+++ b/src/bridge/MatrixHandler.ts
@@ -450,7 +450,7 @@ export class MatrixHandler {
         // invite yet!)
         this.processingInvitesForRooms[event.room_id + event.state_key] = req.getPromise();
         req.getPromise().finally(() => {
-            delete this.processingsForRooms[event.room_id + event.state_key];
+            delete this.processingInvitesForRooms[event.room_id + event.state_key];
         });
 
         // Check if this room is known to us.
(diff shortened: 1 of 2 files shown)


src/bridge/MatrixHandler.ts
```diff
@@ -457,7 +450,7 @@ export class MatrixHandler {
         // invite yet!)
         this.processingInvitesForRooms[event.room_id + event.state_key] = req.getPromise();
         req.getPromise().finally(() => {
-            delete this.processingInvitesForRooms[event.room_id + event.state_key];
+            delete this.processing#sForRooms[event.room_id + event.state_key];
```
tadzik, line 453: I don't think that `#` belongs here :)

tadzik (approved):
Fair enough! This is breaking however – does this warrant a major version bump, from 0 to 1? ;)

Half-Shot:
>Fair enough! This is breaking however – does this warrant a major version bump, from 0 to 1? ;)

I think so, probably coinciding with my plans for the provisioner :)

Issue #1507, opened 2022-01-24
Update minimum version to Node 14


Half-Shot:
[Node 12 will be EOL by April](https://nodejs.org/en/about/releases/), so we should make preparations to bump the minimum version before then.

PR #1515, opened 2022-01-31
Use GitHub Actions for tests (and bump minimum version to Node 14)

Half-Shot:
Fixes https://github.com/matrix-org/matrix-appservice-irc/issues/1507

This should hopefully move our CI pipelines back in tree. We're also bumping the minimum node version as per https://github.com/matrix-org/matrix-appservice-irc/issues/1507.

Once merged, let's kill off https://buildkite.com/matrix-dot-org/matrix-appservice-irc.


jaller94 (approved):
Looks good to me. :+1:

PR #1523, opened 2022-02-13
docs/bridged_networks.md: remove freenode

Mikaela:
[According to Half-Shot on #irc:matrix.org the bridge not being removed from documentation is an oversight,](https://matrix.to/#/!BAXLHOFjvDKUeLafmO:matrix.org/$xzZ6LqVUcb2q7o+TFzOFm1Mw1aXSSBpXVzEFfTTlehU?via=matrix.org&via=libera.chat&via=element.io)


docs/bridged_networks.md
```diff
@@ -53,10 +52,6 @@ A wishlist of IRC networks to be bridged is being collected [in a github issue](
```
Half-Shot, line 52: Just for the sake of tradition.
```suggestion
* The Freenode IRC bridge offically was shut down on 20-Dec 2021.
```

Half-Shot (approved):
One extra line, otherwise LGTM!

docs/bridged_networks.md
Mikaela: I changed the network name to lowercase as that is what they used to prefer and used ISO 8601 date format instead and for this suggestion made you a coauthor.

PR #1526, opened 2022-02-15
Fix user -> bot invites getting ignored

Half-Shot:
If `plumbedState` wasn't set, the code would throw and the invite would not be handled.

### DIFF
--- a/src/bridge/MatrixHandler.ts
+++ b/src/bridge/MatrixHandler.ts
@@ -169,19 +169,13 @@ export class MatrixHandler {
 
         // Do not create an admin room if the room is marked as 'plumbed'
         const matrixClient = this.ircBridge.getAppServiceBridge().getIntent();
-
-        try {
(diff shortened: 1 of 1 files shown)


src/bridge/MatrixHandler.ts
```diff
@@ -169,19 +169,13 @@ export class MatrixHandler {
...
-            if (plumbedState.status === "enabled") {
-                req.log.info(
-                    'This room is marked for plumbing (m.room.plumbing.status = "enabled"). ' +
-                    'Not treating room as admin room.'
-                );
-                return;
-            }
-        }
-        catch (err) {
-            req.log.debug(`Not a plumbed room: Error retrieving m.room.plumbing (${err.data.error})`);
```
Half-Shot, line 184: The bug was actually here, as we never checked if err.data was defined. However, since `getStateEvent` can return null on not found, the try/catch isn't needed.

PR #1528, opened 2022-02-15
Bridge.kill() should safely destroy the bridge

Half-Shot:
We seem to be hitting a bit of a race where user activity is calculated *after* we've finished running tests, causing the rig to fail.

### DIFF
--- a/src/bridge/IrcBridge.ts
+++ b/src/bridge/IrcBridge.ts
@@ -86,7 +86,7 @@ export class IrcBridge {
     private memberListSyncers: {[domain: string]: MemberListSyncer} = {};
     private joinedRoomList: string[] = [];
     private dataStore!: DataStore;
-    private startedUp = false;
+    private bridgeState: "not-started"|"running"|"killed" = "not-started";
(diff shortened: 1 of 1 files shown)


src/bridge/IrcBridge.ts
```diff
@@ -86,7 +86,7 @@ export class IrcBridge {
     private memberListSyncers: {[domain: string]: MemberListSyncer} = {};
     private joinedRoomList: string[] = [];
     private dataStore!: DataStore;
-    private startedUp = false;
+    private bridgeState: "not-started"|"running"|"killed" = "not-started";
```
tadzik, line 89: I know it's functionally identical, but any reason why it's not a proper typescript enum?

tadzik (approved):

src/bridge/IrcBridge.ts
Half-Shot: I tend to only use enums when it's getting used more widespread than this, I suppose it's a question of style. TS having strict string checks is good nuff for me.

PR #1530, opened 2022-02-16
Update packages

Half-Shot:
A bunch of our packages were out of date or unused, this PR just cleans up and checks that our tests keep working. I've checked that none of the major version bumps are breaking.


package.json
```diff
@@ -26,48 +26,44 @@
...
-    "nyc": "^14.1.1",
-    "prom-client": "13.1.0",
-    "proxyquire": "^1.4.0",
-    "typescript": "^4.4.3"
+    "@types/pg": "^8.6.4",
+    "@types/sanitize-html": "^2.6.2",
+    "@typescript-eslint/eslint-plugin": "^5.12.0",
+    "@typescript-eslint/parser": "^5.12.0",
+    "eslint": "^8.9.0",
+    "jasmine": "^4.0.2",
```
Half-Shot, line 63: Jasmine 4.0.2 breaks us currently.

src/bridge/PublicitySyncer.ts
```diff
@@ -32,10 +32,10 @@ export class PublicitySyncer {
...
         };
     } = {
-        mappings: {},
-        channelIsSecret: {},
-        roomVisibilities: {},
-    };
+            mappings: {},
+            channelIsSecret: {},
+            roomVisibilities: {},
+        };
```
jaller94, line 38: This looks wrong. Was it auto-"fixed"?

jaller94 (approved):
While I don't think that the code changes are good, the package upgrade seems fine.
Worst case, we fix/revert the ESLint indent rule afterwards.

src/irc/ConnectionInstance.ts
```diff
@@ -79,6 +79,7 @@ export class ConnectionInstance {
     private state: "created"|"connecting"|"connected" = "created";
     private pingRateTimerId: NodeJS.Timer|null = null;
     private clientSidePingTimeoutTimerId: NodeJS.Timer|null = null;
+    // eslint-disable-next-line no-use-before-define
     private connectDefer: Defer<ConnectionInstance>;
```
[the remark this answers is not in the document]
Half-Shot: It does...but I don't think we're going to fix this today, it would be a huge unpicking.

src/bridge/MemberListSyncer.ts
```diff
@@ -46,9 +46,9 @@ export class MemberListSyncer {
         irc: {[channel: string]: string[]};
         matrix: {[roomId: string]: RoomInfo};
     } = {
-        irc: {},
-        matrix: {},
-    }
+            irc: {},
+            matrix: {},
+        }
```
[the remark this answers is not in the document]
Half-Shot: Yeah, I'm trying to figure out why the indent is so wrong :(

package.json
Half-Shot: https://github.com/matrix-org/matrix-appservice-irc/issues/1533

PR #1532, opened 2022-02-16
Synchronise bridge with m.policy.rule.* updates

Half-Shot:
This PR aims to have the IRC bridge synchronize with https://spec.matrix.org/v1.2/client-server-api/#moderation-policy-lists entries so that when a user is banned via Mjolnir or similar, the bridge can take proactive action to ban them / filter their traffic out. While the bridge shoud not bundle too much in the way of moderation functionality (best left to other projects), this feature is aimed at trying to deal with attacks swiftly rather than waiting for processes to kick in.

### DIFF
--- a/docs/administrators_guide.md
+++ b/docs/administrators_guide.md
@@ -78,6 +78,28 @@ membershipLists:
 in the config. Users can choose to disable this on a per-room basis by modfiying their
 [room config](./room_configuration.md#allowunconnectedmatrixusers) options, if the bridge permits it.
 
+## Subscribing to Moderation policys
+
+Matrix has support for specifying [moderation policy lists](https://spec.matrix.org/v1.2/client-server-api/#moderation-policy-lists).
+Moderation policys are set by services such as [Mjolnir](https://github.com/matrix-org/mjolnir) and can be
+used to inform other services as to how to deal with content.
+
+The bridge can subscribe to rooms containing these policies and can then choose to filter out users and
+servers matching these rules.
+
+```yaml
+  banLists:
+    rooms:
+      - "#matrix-org-coc-bl:matrix.org" # The matrix.org code of conduct ban list
+```
+
+A `m.ban` rule will forcibly kill the connection of any user matching a ban, and will not
+allow them to reconnect. The rules are enforced on startup, when the ban list is modified
+and when the configuration file is reloaded.
+
+Administrators should beware that this configuration is very powerful, and any user caught
+on this list will not be able to use the bridge at all.
+
 ## Metrics / Grafana
 
 The bridge includes a prometheus compatible metrics endpoint which can be used to inspect
--- a/docs/administrators_guide.md
+++ b/docs/administrators_guide.md
@@ -97,7 +97,7 @@ A `m.ban` rule will forcibly kill the connection of any user matching a ban, and
 allow them to reconnect. The rules are enforced on startup, when the ban list is modified
 and when the configuration file is reloaded.
 
-Administrators should beware that this configuration is very powerful, and any user caught
+Administrators should beware that this configuration is **very** powerful, and any user caught
 on this list will not be able to use the bridge at all.
 
 ## Metrics / Grafana
--- a/docs/administrators_guide.md
+++ b/docs/administrators_guide.md
@@ -78,10 +78,10 @@ membershipLists:
 in the config. Users can choose to disable this on a per-room basis by modfiying their
 [room config](./room_configuration.md#allowunconnectedmatrixusers) options, if the bridge permits it.
 
-## Subscribing to Moderation policys
+## Subscribing to Moderation policies
 
 Matrix has support for specifying [moderation policy lists](https://spec.matrix.org/v1.2/client-server-api/#moderation-policy-lists).
-Moderation policys are set by services such as [Mjolnir](https://github.com/matrix-org/mjolnir) and can be
+Moderation policies are set by services such as [Mjolnir](https://github.com/matrix-org/mjolnir) and can be
 used to inform other services as to how to deal with content.
 
 The bridge can subscribe to rooms containing these policies and can then choose to filter out users and
--- a/spec/unit/MatrixBanSync.spec.js
+++ b/spec/unit/MatrixBanSync.spec.js
@@ -5,7 +5,7 @@ const BANNED_USER_STATE_EVENT = {
     type: "m.policy.rule.user",
     state_key: 'banned-user',
     content: {
-        reccomendation: 'm.ban',
+        recommendation: 'm.ban',
         entity: '@user:banned.com',
         reason: 'badly-behaved',
     }
@@ -21,7 +21,7 @@ const BANNED_SERVER_STATE_EVENT = {
     type: "m.policy.rule.server",
     state_key: 'banned-server',
     content: {
-        reccomendation: 'm.ban',
+        recommendation: 'm.ban',
         entity: 'banned-server.com',
         reason: 'badly-constructed',
     }
@@ -64,10 +64,10 @@ describe("MatrixBanSync", () => {
             expect(banSync.handleIncomingState(ev, '!valid:room')).toBeFalse();
             expect(banSync.bannedEntites.get(`!valid:room:banned-user`)).toBeUndefined();
         });
-        it("should skip unknown reccomendation", () => {
+        it("should skip unknown recommendation", () => {
             const ev = {...BANNED_USER_STATE_EVENT, content: {
                 ...BANNED_USER_STATE_EVENT.content,
-                reccomendation: 'm.tea-party',
+                recommendation: 'm.tea-party',
             }};
             expect(banSync.handleIncomingState(ev, '!valid:room')).toBeFalse();
             expect(banSync.bannedEntites.get(`!valid:room:banned-user`)).toBeUndefined();
@@ -121,7 +121,7 @@ describe("MatrixBanSync", () => {
                                 type: 'm.policy.rule.server',
                                 state_key: 'should-not-be-here',
                                 content: {
-                                    reccomendation: 'still-not-interested',
+                                    recommendation: 'still-not-interested',
                                     entity: 'foo.com',
                                     reason: 'foo',
                                 }
--- a/spec/unit/MatrixBanSync.spec.js
+++ b/spec/unit/MatrixBanSync.spec.js
@@ -1,0 +1,151 @@
+const { MatrixGlob } = require("matrix-bot-sdk");
+const { MatrixBanSync } = require("../../lib/bridge/MatrixBanSync");
+
+const BANNED_USER_STATE_EVENT = {
+    type: "m.policy.rule.user",
+    state_key: 'banned-user',
+    content: {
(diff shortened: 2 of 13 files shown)


docs/administrators_guide.md
```diff
@@ -78,6 +78,28 @@ membershipLists:
 in the config. Users can choose to disable this on a per-room basis by modfiying their
 [room config](./room_configuration.md#allowunconnectedmatrixusers) options, if the bridge permits it.
 
+## Subscribing to Moderation policys
```
tadzik, line 81:
```suggestion
## Subscribing to Moderation policies
```

tadzik: Or are both correct?

src/bridge/MatrixBanSync.ts
```diff
@@ -0,0 +1,133 @@
...
+        case "m.policy.rule.server":
+            return BanEntityType.Server
+        default:
+            return null;
+    }
+}
+
+export class MatrixBanSync {
+    private bannedEntites = new Map<string, BanEntity>();
+    private interestingRooms = new Set<string>();
```
tadzik, line 45: I'm not too sure about that name :P Then again, it's a private API (except for when the tests happily ignore that fact), so I don't really mind – in the context of this class it's obvious what it's for. Though maybe `eventSources` would be more representative of its uses?

src/bridge/MatrixBanSync.ts
```diff
@@ -0,0 +1,133 @@
...
+            }
+        }
+    }
+
+    /**
+     * Is the given room considered part of the bridge's ban list set.
+     * @param roomId A Matrix room ID.
+     * @returns true if state should be handled from the room, false otherwise.
+     */
+    public isInterestedInRoom(roomId: string): boolean {
```
tadzik, line 71: Ah, here's when "interesting" is a public API – but I think it makes sense here. Leaving all these comments as a food for thought though. Maybe `isTrackingRoomState()` would be more fitting here? I don't have strong opinions, just dropping it in here.

tadzik (approved):
Looks good!

src/bridge/MatrixBanSync.ts
```diff
@@ -0,0 +1,133 @@
...
+interface BanEntity {
+    matcher: MatrixGlob;
+    entityType: BanEntityType;
+    reason: string;
+}
+
+interface MPolicyContent {
+    entity: string;
+    reason: string;
+    reccomendation: "m.ban";
```
jaller94, line 29: For the entire PR: `reccomendation` -> `recommendation` (2c 1m -> 1c 2m)

src/bridge/MatrixBanSync.ts
```diff
@@ -0,0 +1,133 @@
...
+            return false;
+        }
+        const key = `${roomId}:${evt.state_key}`;
+        if (evt.content.entity === undefined) {
+            // Empty, delete instead.
+            log.info(`Deleted ban rule ${evt.type} matching ${content.entity}`);
+            this.bannedEntites.delete(key);
+            return false;
+        }
+        if (content.reccomendation !== "m.ban") {
```
Gnuxie, line 93: quite a lot of rules still have `org.matrix.mjolnir.ban`

src/bridge/IrcBridge.ts
```diff
@@ -949,6 +958,17 @@ export class IrcBridge {
         );
         for (const [userId, {display_name}] of Object.entries(members)) {
             try {
+                // If the user is banned, skip any connection attempts and go straight for a kick.
+                const banReason = this.matrixBanSyncer?.isUserBanned(userId);
+                if (banReason) {
+                    req.log.debug(`Not syncing ${userId} - user banned (${banReason})`);
+                    this.membershipQueue.leave(
+                        roomId, userId, req, true,
+                        "You are banned from interacting with this bridge.",
```
Gnuxie, line 967: Will they know why?

src/bridge/MemberListSyncer.ts
```diff
@@ -264,6 +264,11 @@ export class MemberListSyncer {
         const idleRegex = this.server.ignoreIdleUsersOnStartupExcludeRegex;
         for (const roomInfo of filteredRooms) {
             for (const uid of roomInfo.realJoinedUsers) {
+                const banReason = this.ircBridge.matrixBanSyncer?.isUserBanned(uid);
+                if (banReason) {
+                    log.debug(`Not syncing ${uid} - user banned (${banReason})`)
```
Gnuxie, line 269: I this just to avoid syncing the presence of the user in the room to the IRC side?

Gnuxie (commented):
I think the approach is the right one, especially checking the domain part of their user id for banned servers rather than using ACL.

The only thing I'm slightly worried about is how often you would have to call `checkForBannedConnectedUsers` and `syncMembersInRoomToIrc`?

Does `syncMembersInRoomToIrc` happen with each join/leave? FTR we also scan all members each time membership changes happen in Mjolnir at the moment, so it is probably OK. Though I have no idea if it always will be.

src/bridge/MatrixBanSync.ts
```diff
@@ -0,0 +1,133 @@
...
+            try {
+                const roomId = await intent.join(roomIdOrAlias);
+                this.interestingRooms.add(roomId);
+                const roomState = await intent.roomState(roomId, false) as WeakStateEvent[];
+                for (const evt of roomState) {
+                    this.handleIncomingState(evt, roomId);
+                }
+            }
+            catch (ex) {
+                log.error(`Failed to read ban list from ${roomIdOrAlias}`, ex);
```
jaller94, line 61: Optional feature request:
As an administrator of a bridge, I'd like this to be posted to an admin room.

jaller94 (changes requested):

src/bridge/MatrixBanSync.ts
Half-Shot: Ah good to know, I wasn't sure if we had ported them across. Will support both.

src/bridge/IrcBridge.ts
Half-Shot: We can use the `reason` field, if we think it's safe to explicitly say.

src/bridge/MatrixBanSync.ts
```diff
@@ -0,0 +1,133 @@
...
+    public handleIncomingState(evt: WeakStateEvent, roomId: string) {
+        const content = evt.content as unknown as MPolicyContent;
+        const entityType = eventTypeToBanEntityType(evt.type);
+        if (!entityType) {
+            return false;
+        }
+        const key = `${roomId}:${evt.state_key}`;
+        if (evt.content.entity === undefined) {
+            // Empty, delete instead.
+            log.info(`Deleted ban rule ${evt.type} matching ${content.entity}`);
```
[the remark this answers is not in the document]
Half-Shot: Of of course :facepalm:

docs/administrators_guide.md
Half-Shot: I think Policies is more correct than Policys

src/bridge/MatrixBanSync.ts
```diff
@@ -0,0 +1,133 @@
...
+            this.bannedEntites.delete(key);
+            return false;
+        }
+        if (content.reccomendation !== "m.ban") {
+            // We only deal with m.ban at the moment.
+            return false;
+        }
+        this.bannedEntites.set(key, {
+            matcher: new MatrixGlob(content.entity),
```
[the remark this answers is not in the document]
Half-Shot: I don't know if there is much more we can do than check for non-empty strings. I think we have to trust the input somewhat, as it's a state event sent into a room that is trusted by the bridge admin.

src/bridge/MemberListSyncer.ts
Half-Shot: Yep, it's just a shortcut to avoid having to do any processing on users we know will fail to get connected later.

Half-Shot:
>The only thing I'm slightly worried about is how often you would have to call checkForBannedConnectedUsers and syncMembersInRoomToIrc?

We can probably make this more efficient actually by only checking the user that was banned.

Half-Shot:
>Does syncMembersInRoomToIrc happen with each join/leave? FTR we also scan all members each time membership changes happen in Mjolnir at the moment, so it is probably OK. Though I have no idea if it always will be.

It's called when there is a known discrepancy between the membership of Matrix and IRC, so not super frequently enough to be a problem hopefully.

spec/unit/MatrixBanSync.spec.js
```diff
@@ -0,0 +1,151 @@
...
+
+const BANNED_SERVER_ENTITY = {
+    matcher: new MatrixGlob(BANNED_SERVER_STATE_EVENT.content.entity),
+    entityType: BANNED_SERVER_STATE_EVENT.type,
+    reason: BANNED_SERVER_STATE_EVENT.content.reason
+};
+
+let banSync;
+
+describe("MatrixBanSync", () => {
```
jaller94, line 38:
```suggestion
describe("MatrixBanSync", () => {
    let banSync;
```

spec/unit/MatrixBanSync.spec.js
```diff
@@ -0,0 +1,151 @@
...
+    beforeEach(() => {
+        banSync = new MatrixBanSync({ rooms: [] });
```
jaller94, line 40: I wonder if mocha tests can have local variables.
With `banSync` being reused and mutated the tests cannot run in parallel.
Then again, running tests in parallel is not a requirement we have to fulfill.

src/bridge/MatrixBanSync.ts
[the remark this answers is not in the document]
jaller94: This check is what I was looking for. :+1: 
https://github.com/matrix-org/matrix-appservice-irc/blob/8ab368f1c7c673c5e7191b78a4ed9c057c79fc2e/src/bridge/MatrixBanSync.ts#L104-L106

jaller94 (approved):

src/bridge/MatrixBanSync.ts
```diff
@@ -0,0 +1,143 @@
...
+        if (!supportedRecommendations.includes(content.recommendation)) {
+            return false;
+        }
+        if (typeof content.entity !== "string" || content.entity === "") {
+            throw Error('`entity` key is not valid, must be a non-empty string');
+        }
+        this.bannedEntites.set(key, {
+            matcher: new MatrixGlob(content.entity),
+            entityType,
+            reason: content.reason || "No reason given",
```
jaller94, line 110: Could be a number, an object, `true` or other(?) valid `content` value sub type.
But I don't think that actually breaks anything..
Worst case, your message could be that you were banned because of `[Object object]`.

Half-Shot: Yes, your admin is very silly if they put silly reasons in, I think.

spec/unit/MatrixBanSync.spec.js
Half-Shot: (it's not even mocha, it's :sparkle: jasmine :sparkle:)

But yeah, one day a big refactor will be due.

PR #1534, opened 2022-02-17
Remove old bridge version code

Half-Shot:
This was preventing the bridge from starting up due to a duplicate metric.


src/util/PackageInfo.ts
```diff
@@ -1,13 +0,0 @@
-
-let BridgeVersion: string;
```
Half-Shot, line 2: This is all implemented by the upstream library now, so no need to hang onto this.

tadzik (approved):

PR #1541, opened 2022-03-08
Prune non-production dependencies from runtime image

gdamjan:
prune dev packages from the runtime container image, to make it leaner

Before:
```
$ du -sh node_modules/
187M    node_modules
```

After:
```
$ du -sh node_modules/
89M     node_modules/
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # Freebind build
 # node:16-slim uses debian:stretch-slim as a base, so it's safe to build on here.
-FROM debian:stretch-slim as freebind 
+FROM debian:stretch-slim as freebind
 
 RUN apt-get update \
  && apt-get install -y git build-essential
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -13,13 +13,14 @@ FROM node:14-slim as builder
 
 WORKDIR /build
 
-COPY src/ /build/src/ 
+COPY src/ /build/src/
 COPY .eslintrc *json /build/
 
 RUN npm ci
 RUN npm run build
+RUN npm prune --production
```
gdamjan, line 21:
```suggestion
RUN npm prune --omit dev
```

gdamjan:
Do I need to rebase on top of `develop`, not sure what the relation between `master` and `develop` branches is.

gdamjan:
> these days we should be doing separate install steps for build and for production.

what do you mean by that?
something like `npm install --production /folder`

Half-Shot:
>is the sign-off check not recognizing my cyrillic name ? :)

Actually, I think GitHub actions has exploded (https://www.githubstatus.com/incidents/smn1qtqvbsb6). I don't think your characters broke the entirety of GitHub actions, but who's to say :p

gdamjan:
github actions is alive \o/

PR #1547, opened 2022-03-20
Surround URL in truncated messages with delimiters

wasamasa:
URLs use double quotes, angular brackets or whitespace as delimiters. Parentheses are URL constituent characters (see Wikipedia for example). While many URL regex treat closing delimiters at the end of a URL specially, far from all do, therefore URLs should be delimited with appropriate characters whenever possible.

See https://github.com/matrix-org/matrix-appservice-irc/blob/957e60c48afbfba60cc7fc0d8369bb8eeaea9e57/src/models/MatrixAction.ts#L202-L210 for existing code using a combination of angular brackets and whitespace.

### DIFF
--- a/changelog.d/1547.bugfix
+++ b/changelog.d/1547.bugfix
@@ -1,0 +1,1 @@
+Consistently use URL delimiters without spaces in truncated messages
--- a/src/bridge/MatrixHandler.ts
+++ b/src/bridge/MatrixHandler.ts
@@ -58,7 +58,7 @@ const DEFAULTS: MatrixHandlerConfig = {
     shortReplyTresholdSeconds: 5 * 60,
     shortReplyTemplate: "$NICK: $REPLY",
     longReplyTemplate: "<$NICK> \"$ORIGINAL\" <- $REPLY",
-    truncatedMessageTemplate: "(full message at $URL)",
+    truncatedMessageTemplate: "(full message at <$URL>)",
 };
 
 export interface MatrixEventInvite {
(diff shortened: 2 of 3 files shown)


wasamasa:
Sorry, I've overlooked the mandatory changelog message

Mikaela (approved):
👍🏻 enclosing URLs in `<>` is universally recommended and part of RFC 3986 appendix C

> In practice, URIs are delimited in a variety of ways, but usually within double-quotes "http://example.com/", angle brackets
<http://example.com/>, or just by using whitespace:
…
> Using <> angle brackets around each URI is especially recommended as a delimiting style for a reference that contains embedded whitespace.

I have heard that ending the URL to a closing bracket (`)`) that is not actually a part of it, breaks opening the link with some apps.

wasamasa:
Thank you for the review. For consistency's sake, would you agree to change the other location currently using angular brackets and spaces, too?

Mikaela:
>  For consistency's sake, would you agree to change the other location currently using angular brackets and spaces, too?

I don't see why not and I wasn't aware of it being in multiple places.

wasamasa:
Thanks, I've adjusted all locations and updated the changelog accordingly

wasamasa:
Not sure if @Mikaela needs to approve this for progress to be made or someone else.

tadzik (approved):
Makes sense - thanks for the patch!

I was about to ask about updating the sample config as well, but it seems like this option isn't actually configurable. Oops.

wasamasa:
Seems this is no longer needed. I've noticed IRC services on Libera do surround truncated messages correctly and this seems to be linked to this related change: https://github.com/matrix-org/matrix-appservice-irc/pull/1573

PR #1549, opened 2022-03-22
Make towncrier build the changelog instead of printing its version

jaller94:
Same as https://github.com/matrix-org/matrix-appservice-slack/pull/677

## Before
```
./scripts/towncrier.sh 
towncrier, version 21.9.0
```

## After
```
./scripts/towncrier.sh
Loading template...
Finding news fragments...
Rendering news fragments...
Writing to newsfile...
Staging newsfile...
Removing news fragments...
I want to remove the following files:
…
```


tadzik (approved):
Makes sense. Looks like CI is being picky about your changelog entry though :)

PR #1551, opened 2022-03-28
Check the banReason before banning a user.

Half-Shot:
This was causing users to get banned erroneously

### DIFF
--- a/src/irc/ClientPool.ts
+++ b/src/irc/ClientPool.ts
@@ -558,8 +558,10 @@ export class ClientPool {
             for (const [userId, client] of set.userIds.entries()) {
                 try {
                     const banReason = this.ircBridge.matrixBanSyncer?.isUserBanned(userId);
-                    log.warn(`Killing ${userId} client connection due - user is banned (${banReason})`);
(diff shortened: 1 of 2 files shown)


jaller94 (approved):

src/irc/ClientPool.ts
```diff
@@ -558,8 +558,10 @@ export class ClientPool {
             for (const [userId, client] of set.userIds.entries()) {
                 try {
                     const banReason = this.ircBridge.matrixBanSyncer?.isUserBanned(userId);
-                    log.warn(`Killing ${userId} client connection due - user is banned (${banReason})`);
-                    await client.kill('User was banned');
+                    if (banReason) {
```
jaller94, line 561: I'd prefer some test that also kills on empty string.
While the `matrixBanSyncer` will default empty strings to "No reason given", only the return value `false` states that the user isn't banned.

```suggestion
                    if (typeof banReason === 'string') {
```

Half-Shot:
(You were right about `typeof`, checking for false would have banned everyone)

PR #1559, opened 2022-04-29
Add useful information to the CTCP version request

Half-Shot:
This changes the string from something vague to the specific version running, and the homeserver handling the bridging. This obviously is a breakaway from obscuring our version, though given we already have a version command this doesn't feel like a major security footprint change.

### DIFF
--- a/src/irc/ConnectionInstance.ts
+++ b/src/irc/ConnectionInstance.ts
@@ -21,6 +21,7 @@ import * as logging from "../logging";
 import Bluebird from "bluebird";
 import { Defer } from "../promiseutil";
 import { IrcServer } from "./IrcServer";
+import { getBridgeVersion } from "matrix-appservice-bridge";
 
 const log = logging.get("client-connection");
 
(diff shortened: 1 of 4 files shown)


src/irc/ConnectionInstance.ts
```diff
@@ -42,7 +43,8 @@ const FLOOD_PROTECTION_DELAY_MS = 700;
 const THROTTLE_WAIT_MS = 20 * 1000;
 
 // String reply of any CTCP Version requests
-const CTCP_VERSION = 'matrix-appservice-irc, part of the Matrix.org Network';
+const CTCP_VERSION = 
+    (homeserverName: string) => `matrix-appservice-irc ${getBridgeVersion()} bridged via ${homeserverName}`;
```
Half-Shot, line 47: Do we need to escape these strings do you think?

tadzik: I don't think this can break anything on accident, and if someone wants to break it deliberately then there are easier options for someone controlling the HS and the bridge :)

tadzik (approved):
LGTM

jesopo:
for what it's worth, this looks good to me

PR #1566, opened 2022-05-16
Enable authenticateThirdpartyEndpoints

Half-Shot:
This has been a problem for years, and we might as well start enabling authentication for these endpoints now. https://github.com/matrix-org/matrix-appservice-bridge/issues/76


Half-Shot:
The next Synapse release will also support this change anyhow, so I'm going to opt for fixing this in matrix-appservice-bridge https://github.com/matrix-org/synapse/pull/12746

Issue #1563, opened 2022-05-04
"(IRC)" should only be part of the display, not of the message


pchampin:
**Is your feature request related to a problem? Please describe.**
When I use auto-completion to mention an IRC user (say 'john'), my matrix client auto-completes it as 'john (IRC)'.
While I don't mind it being displayed "john (IRC)" in Matrix (it is an interesting info), I don't think that this precision should be part of the actual message sent to IRC.
- for human IRC users, it might be surprising
- for some bots, it prevents them from working (because they expect something like `<their_name>, <command>` and don't recognize `<their_name> (IRC), <command>`).

**Describe the solution you'd like**
Either use "(IRC)" only in the Matrix client, but *not* in the messages sent to IRC,
or (if it is too complex), drop the "(IRC)" mention all together (or make it configurable).

Mikaela:
Which bridge and client are you seeing this with? It may be an issue with that particular instance as the "(IRC)" suffix was dropped back in August 2018 https://github.com/matrix-org/matrix-appservice-irc/issues/194#issuecomment-415734087 (and before that in 2016 "Riot" used to remove it automatically https://github.com/matrix-org/matrix-appservice-irc/issues/194#issuecomment-268208287).

pchampin:
> the "(IRC)" suffix was dropped back in August 2018
Was it? Interesting :thinking: 

I am using Element Desktop for Linux:
Element version: 1.10.11
Olm version: 3.2.8

My server is matrix.org.
An example IRC channel I use is https://matrix.to/#/#_w3c_#rdf-dev:matrix.org

tadzik:
This is indeed a misconfiguration of w3c bridge on matrix org, due to the fact that this is the IRC bridge's default behaviour. Dropping it from the example config is probably a sensible thing to default to.

pchampin:
Is it something that I should discuss with the people at Matrix.org? The people at W3C?

tadzik:
That's something that we, as a team maintaining matrix.org bridges, need to do – shouldn't be long now :)

Issue #194, opened 2016-09-27
Let's remove the ' (IRC)' suffix entirely.


ara4n:
It just annoys people, and the mxid is good enough, until we have groups/profiles

Kegsay:
What's the deal on this? We have 248823 `@freenode_` users on the bridge database. I can write a script to rename them all to non-(IRC). I can update the config so subsequent users won't have `(IRC)`. We're currently blocked on Synapse not sucking?

ara4n:
let's just fix the config for now for new users. if we can also update current ones when they next speak, that will probably help too.

Kegsay:
This has been mentioned again, so assigning someone so this doesn't get lost.

I'd rather we didn't have to add code to the bridge to migrate users to the new display name style (like you propose when they "next speak"), as that will inevitably bitrot. I'd much rather do a config change + one time script to enumerate existing users, but I'm concerned that Synapse won't be up to the task of updating that many users. We can't really heavily rate limit this because there's too many of them, 1 per 10s would take 28.8 **days** to adjust. 1 per second would take 2.88 days which may be good enough?

Kegsay:
This has been mitigated because Riot removes " (IRC)" when tab-completing now. We still need to update the names, but the urgency is much less now.

Croydon:
Couldn't that be confusing when when there is one Matrix user and one IRC user with the same nickname/display name and both without profil picture? The only way to distinguish them would be a look on their user information.

On the other hand this (IRC) indeed annoys myself a bit. So not sure what to think about removing it all together.

Kegsay:
@Croydon yes this was why the suffixes were originally added. Really the complete solution is to implement https://github.com/matrix-org/matrix-doc/issues/646

jcgruenhage:
@Croydon Well, if there are two users with the same displayname in one room, riot shows their complete MXID anyway, so that won't be more confusing I think.

Half-Shot:
This has been fixed in the bridge. There is a bug internally for removing the suffixes from existing users.

HackerFoo:
Is there a script to remove the suffix from existing databases? If so, could it be posted or linked here?

flackr:
Is it possible after changing the displayName on my bridge to purge the current users and have them recreated?

reuben:
[...]

PR #1567, opened 2022-05-18
Remove (IRC) as a default suffix for displayNames

tadzik:
This was decided and implemented on matrix.org long ago (with one
exception), but is still an unnecessary default.

See https://github.com/matrix-org/matrix-appservice-irc/issues/194#issuecomment-415734087

Will eventually fix https://github.com/matrix-org/matrix-appservice-irc/issues/1563 once the w3c on matrix.org gets the version with it (can be fixed in the configuration before then).


Half-Shot (approved):
Seems good to me. Make sure to call this out loudly in the changelog.

Issue #1397, opened 2021-06-18
Bridge shoud allow people to opt out of idle kicking


Half-Shot:
Or just perhaps a really long idleness period. Either way, I'd want to give a captcha or something meaningful to solve to prevent bots signing up and eating up our counts.

piegamesde:
How does this relate to #1268?

PR #1571, opened 2022-05-25
Add a DebugAPI to warn users that they will be idle kicked

Half-Shot:
Partial fix to https://github.com/matrix-org/matrix-appservice-irc/issues/1397 (doesn't allow opt out, but does warn you now)

This PR will allow services to advance-notify users that they are going to be idle kicked. Like user reaping, this is called from the DebugAPI. This also includes a dummy command "!active" to let the bridge know that you are active. I left this in just so the user has something tangible to do, but any messages sent to the bridge will reset your timeout anyway.

### DIFF
--- a/src/bridge/AdminRoomHandler.ts
+++ b/src/bridge/AdminRoomHandler.ts
@@ -76,8 +76,7 @@ const COMMANDS: {[command: string]: Command|Heading} = {
     },
     "!active": {
         example: "!active",
-        summary: "Informs the bridge that you do not wish to be idlekicked. This will reset your idleness timer " +
-                 "but you may need to run this again."
+        summary: "Mark yourself as active, which will exclude you from any idleness kicks."
     },
     'Authentication': { heading: true },
     "!storepass": {
--- a/src/bridge/AdminRoomHandler.ts
+++ b/src/bridge/AdminRoomHandler.ts
@@ -17,7 +17,7 @@ limitations under the License.
 import { BridgeRequest } from "../models/BridgeRequest";
 import { MatrixRoom, MatrixUser } from "matrix-appservice-bridge";
 import { IrcBridge } from "./IrcBridge";
-import { MatrixAction } from "../models/MatrixAction";
+import { ActionType, MatrixAction } from "../models/MatrixAction";
 import { IrcServer } from "../irc/IrcServer";
 import { BridgedClient } from "../irc/BridgedClient";
 import { IrcClientConfig } from "../models/IrcClientConfig";
@@ -119,7 +119,7 @@ const COMMANDS: {[command: string]: Command|Heading} = {
 };
 
 class ServerRequiredError extends Error {
-    notice = new MatrixAction("notice", "A server address must be specified.");
+    notice = new MatrixAction(ActionType.Notice, "A server address must be specified.");
 }
 
 const USER_FEATURES = ["mentions"];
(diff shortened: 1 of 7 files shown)


src/models/MatrixAction.ts
```diff
@@ -22,25 +22,42 @@ import escapeStringRegexp from "escape-string-regexp";
 import logging from "../logging";
 const log = logging("MatrixAction");
 
-const ACTION_TYPES = ["message", "emote", "topic", "notice", "file", "image", "video", "audio", "command"];
-const EVENT_TO_TYPE: {[mxKey: string]: string} = {
-    "m.room.message": "message",
-    "m.room.topic": "topic"
+export enum ActionType {
```
Half-Shot, line 25: I got annoyed that these wern't enums, so this PR includes a free refactor.

src/bridge/AdminRoomHandler.ts
```diff
@@ -74,6 +74,11 @@ const COMMANDS: {[command: string]: Command|Heading} = {
         summary: "Leave all bridged channels, on all networks, and remove your " +
                 "connections to all networks.",
     },
+    "!active": {
+        example: "!active",
+        summary: "Informs the bridge that you do not wish to be idlekicked. This will reset your idleness timer " +
+                 "but you may need to run this again."
```
tadzik, line 80: This reads to me a little like "try typing it in a few times just in case". We may want to clarify when's a good time for the "again".

src/bridge/IrcBridge.ts
```diff
@@ -1486,26 +1486,32 @@ export class IrcBridge {
...
+     * @param minIdleHours The minimum number of hours to be considered idle.
+     * @param defaultOnline Whether the user should be defaulted to online or offline if we hold no data for them.
+     * @param excludeRegex A regex of users to exclude from the check.
+     * @param maxIdleHours The maximum number of hours to be considered
+     *                     idle before they aren't considered part of the pool. By default, this isn't checked.
+     * @returns A ordered set of userIds by their idle time in ascending order.
+     */
+    public async calculateIdlenessPool(
+        server: IrcServer, minIdleHours: number,
+        defaultOnline?: boolean, excludeRegex?: string,
```
tadzik, line 1501:
```suggestion
        defaultOnline: boolean, excludeRegex?: string,
```

We're only calling it with a defined boolean – allowing undefined brings a question of "but what happens if you don't pass it in" which will be confusing since it defaults to true on the API side of things. (Same in the other methods).

src/bridge/IrcBridge.ts
```diff
@@ -1486,26 +1486,32 @@ export class IrcBridge {
...
+     * Calculate the number of idle users
+     * @param server The IRC server which we want to scope the idle check to.
+     * @param minIdleHours The minimum number of hours to be considered idle.
+     * @param defaultOnline Whether the user should be defaulted to online or offline if we hold no data for them.
+     * @param excludeRegex A regex of users to exclude from the check.
+     * @param maxIdleHours The maximum number of hours to be considered
+     *                     idle before they aren't considered part of the pool. By default, this isn't checked.
+     * @returns A ordered set of userIds by their idle time in ascending order.
+     */
+    public async calculateIdlenessPool(
```
tadzik, line 1499: I don't think this needs to be public.

tadzik (changes requested):

tadzik (approved):

PR #1573, opened 2022-05-29
Wrap URL in angle brackets in truncatedMessageTemplate

progval:
The closing parenthesis can confuse parsers, as this character is [allowed in IRI paths](https://datatracker.ietf.org/doc/html/rfc3987#section-2.2)

Angle brackets are a [common URL/IRI delimiter](https://datatracker.ietf.org/doc/html/rfc1738#page-3), which is [not allowed in IRIs](https://datatracker.ietf.org/doc/html/rfc3987#page-13) so they should work much better.

Signed-off-by: Valentin Lorentz \<progval plus git at progval dot net\>

### DIFF
--- a/changelog.d/1573.bugfix
+++ b/changelog.d/1573.bugfix
@@ -1,1 +1,1 @@
-truncatedMessageTemplate now defaults to wrapping URLs in angle brackets
+Truncated messages now default to wrapping URLs in angle brackets.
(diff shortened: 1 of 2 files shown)


Half-Shot (commented):
I can't see anything wrong with this change :+1: 

Just needs sign off and a changelog entry https://github.com/matrix-org/matrix-appservice-bridge/blob/develop/CONTRIBUTING.md#%EF%B8%8F-pull-requests

Half-Shot (approved):

PR #1596, opened 2022-08-22
Disable metrics by default

TheStranjer:
Says that metrics should be disabled by default, but are enabled instead.

Signed-off-by:  `Bob Arctor <neetzsche@tutanota.com>`

### DIFF
--- a/changelog.d/1596.bugfix
+++ b/changelog.d/1596.bugfix
@@ -1,0 +1,1 @@
+Disable metrics by default.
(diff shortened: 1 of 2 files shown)


Half-Shot (approved):
Looks good: Can you sign off as per https://github.com/matrix-org/matrix-appservice-bridge/blob/develop/CONTRIBUTING.md#%EF%B8%8F-pull-requests ?

TheStranjer:
I just got done driving for 7 hours so maybe I'm missing something. I think I did sign off. Is it missing the `changelog.d` entry? I put that in just now.

PR #1597, opened 2022-08-26
Fix typo

strugee:

### DIFF
--- a/docs/irc_operators.md
+++ b/docs/irc_operators.md
@@ -88,5 +88,5 @@ Bridged IRC rooms do not share history to Matrix users from before they have joi
 but history visibility can be changed by users with the correct power level on Matrix.
 
 The bridge can also be configured to stop bridging all traffic from a channel to Matrix if it
-cannot garuntee that a Matrix user is joined to the IRC channel, which is usually a step of last
+cannot guarantee that a Matrix user is joined to the IRC channel, which is usually a step of last
 resort should the bridge have failed to connect them. See the [administrators guide](administrators_guide.md#enforcing-matrix-users-to-be-connected-to-irc).


jaller94:
Hi @strugee, thanks for fixing this typo.
All matrix.org projects require a sign off, and I sadly don't know of an exception for small copy changes.

https://matrix-org.github.io/synapse/latest/development/contributing_guide.html#sign-off

jaller94 (changes requested):
Sign off seems to be missing.
I'll check with the team, if we can simplify this, by automating the request of a sign off and not requiring a sign off for minor contributions like typos.

morguldir:
> Sign off seems to be missing.
> I'll check with the team, if we can simplify this, by automating the request of a sign off and not requiring a sign off for minor contributions like typos.

I recommend checking out https://github.blog/changelog/2022-06-08-admins-can-require-sign-off-on-web-based-commits/ as well (the commit was gpg signed by github, so it was web based)

strugee:
Happy to sign DCO, just didn't know about it since I didn't see it in my quick scan of CONTRIBUTING.md - maybe link to https://github.com/matrix-org/matrix-appservice-bridge/blob/develop/CONTRIBUTING.md#%EF%B8%8F-pull-requests in the documentation section, since it's buried under "contributing code" which I dismissed as irrelevant? And it sounds like there are some other process improvements to be made here too.

In the meantime I'll push an amended patch with the signoff line!

strugee:
Note that a typo fix like this is probably not even copyrightable.

strugee:
Done!

strugee:
Tests are failing due to lack of a changelog entry. Does this really need one or is the status check wrong?

Half-Shot:
I don't think it's worth a changelog for this change :+1:

PR #1604, opened 2022-09-07
Strongly typed emitters

Half-Shot:
The other side of https://github.com/matrix-org/node-irc/pull/91. This ensures that the bridge is compatible with the type changes in matrix-org-irc


tadzik (approved):
Looks good, but CI seems unhappy.

PR #1608, opened 2022-09-20
Logging: Replace Object with Map

jaller94:
Replace Object with Map to avoid risk of prototype pollution by malicious logger names.

Risk: very low – requires malicious or ill-advised code changes


src/logging.ts
```diff
@@ -177,17 +175,15 @@ export function configure(opts: LoggerConfig) {
     // reconfigure any existing loggers. They may have been lazily loaded
     // with the default config, which is now being overwritten by this
     // configure() call.
-    Object.keys(loggers).forEach(function(loggerName) {
-        const existingLogger = loggers[loggerName];
-        // remove each individual transport
+    for (const [, existingLogger] of loggers) {
```
tadzik, line 178:
```suggestion
    for (const existingLogger of loggers.values()) {
```

tadzik (approved):
Nitpick, looks good.

src/logging.ts
jaller94: I've used the above syntax in a lot of places.
Because this is totally bike shedding territory (in my opinion), I wonder if there's an ESLint rule to at least enforce consistency.

PR #1610, opened 2022-09-21
IrcBridge.memberListSyncers: Object -> Map

jaller94:
* TypeScript pretends that for the Object `{[foo: string], Bar}` every key will return you an `Bar`, even though `undefined` is a possibility. The usage of `Map` shows us that we may need to validate if a MemberListSyncer was returned.
* Protects against the unlikely foot gun that someone configures a `__proto__` domain.


changelog.d/1610.misc
```diff
@@ -0,0 +1 @@
+IrcBridge.memberListSyncers: Object -> Map.
```
Half-Shot, line 1: This needs to be much more descriptive :). Needs to identify the change from the end user perspective.

PR #1611, opened 2022-09-21
Refactors: Reduce use of Bluebird, increase explicit types, use safer loops

jaller94:


src/DebugApi.ts
```diff
@@ -305,11 +300,11 @@ export class DebugApi {
         });
 
         // wait 3s to pool responses
-        return Bluebird.delay(3000).then(function() {
-            // unhook listener to avoid leaking
-            client.removeClientListener("raw", listener);
-            return buffer.join("\n") + "\n";
-        });
+        await new Promise(resolve => setTimeout(resolve, 3000));
```
Half-Shot, line 303: promiseutil.delay?

Half-Shot (approved):
No complaints from me.

src/DebugApi.ts
jaller94: Resolved in f87bdd11a12b4e9a093a1413d9e162ef8b4647e1

PR #1612, opened 2022-09-21
Fix distribution of IrcServer.randomDomain()

jaller94:
Does the former method have any advantages?

The larger the array of domains, the more it will tend to pick an item with a low index.
Extreme case: If the list has more than 1000 domains, only the first 1000 will be picked.


Half-Shot (approved):
Yes, I'm also struggling to see the purpose of the `100`. Approved.

PR #1614, opened 2022-09-22
PublicitySyncer: Replace Object with Map

jaller94:
Replace Object with Map to avoid risk of prototype pollution by malicious Matrix Room IDs.

Risk: low – Requires attacker to be able to add unvalidated Matrix Room IDs into the IRCchannel<->MatrixRoom mappings.

### DIFF
--- a/src/bridge/PublicitySyncer.ts
+++ b/src/bridge/PublicitySyncer.ts
@@ -19,19 +19,15 @@ export class PublicitySyncer {
 
     // Cache the mode of each channel, the visibility of each room and the
     // known mappings between them. When any of these change, any inconsistencies
-    // should be resolved by keeping the matrix side as private as necessary
+    // should be resolved by keeping the Matrix side as private as necessary.
     private visibilityMap: {
-        mappings: {
-            [roomId: string]: string[];
-        };
-        channelIsSecret: {
-            [networkIdChannel: string]: boolean;
-            // '$networkId $channel': true | false
-        };
-        roomVisibilities: {
-            [roomId: string]: "private"|"public";
-        };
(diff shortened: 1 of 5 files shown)


changelog.d/1614.bugfix
```diff
@@ -0,0 +1 @@
+PublicitySyncer: Replace Object with Map to avoid risk of prototype pollution by malicious Matrix Room IDs.
```
Half-Shot, line 1: Needs to describe the functional change in behaviour for the end-user/admin, rather than the code change we made

src/bridge/PublicitySyncer.ts
```diff
@@ -126,25 +122,27 @@ export class PublicitySyncer {
         const mappings = await this.ircBridge.getStore().getMatrixRoomsForChannel(server, channel);
         const roomIds = mappings.map((m) => m.getId());
 
-        this.visibilityMap.mappings = {};
+        this.visibilityMap.mappings = new Map();
 
         // Update rooms to correct visibilities
-        let currentStates: {[roomId: string]: "public"|"private"} = {};
+        let currentStates: Map<string, "public"|"private"> = new Map();
```
Half-Shot, line 128: Surprised we don't have a type for "public"|"private"

Half-Shot (changes requested):

src/bridge/PublicitySyncer.ts
jaller94: Added in https://github.com/matrix-org/matrix-appservice-irc/pull/1614/commits/7f58bfc1204703ee1b0d38d8dac215c1539764d5

Half-Shot (approved):

PR #1615, opened 2022-09-22
Replace uses of the deprecated String.prototype.substr()

jaller94:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr


Half-Shot (approved):
This is great, I wonder if we should protect ourselves by linting ala https://github.com/cloudflare/eslint-plugin-cflint/blob/master/docs/no-substr.md

jaller94:
I'll add an ESLint rule, if I can find a small config for it.
Adding an additional dependency for this rule sounds like a big risk for a tiny reward.

PR #1619, opened 2022-09-26
Prevent low-risk SQL injection from roomId values

jaller94:
Prevents low-risk SQL injection from roomId values when checking room visibility.

This fixes a hard to exploit SQL injection vector.
To abuse this, an attacker needs to be able to set malicious Matrix IDs in the room mappings.


Half-Shot (approved):
This does the things, though we might want to revise the changelog statement.

PR #1620, opened 2022-09-26
RoomAccessSyncer: Await processing of cached channel modes that changed.

jaller94:
`promises.concat()` does not mutate the array, so possible `this.onMode()` promises are not awaited by `Promise.all()`.

## Relevant code

https://github.com/matrix-org/matrix-appservice-irc/blob/ed0d64767db7fc7720d316326673e7c99f7947cf/src/bridge/RoomAccessSyncer.ts#L350-L364


changelog.d/1620.bugfix
```diff
@@ -0,0 +1 @@
+RoomAccessSyncer: Await processing of cached channel modes that changed.
```
Half-Shot, line 1: Suggest making this a misc, unless we can think of a user-facing name of a bug we've fixed?

PR #1622, opened 2022-09-28
E2E testing against IRCD/Homeserver

Half-Shot:
This work contains the initial parts required to construct full e2e tests. This uses a similar technique to the matrix-node-irc library of providing a specalised test class for creating all the dependencies needed for testing.

TODO:
- [x] matrix-node-irc release
- [x] CI support
- [x] actual tests

Outstanding bugs:

- [x] Inspircd won't let us connect when running via GH actions, possibly IP issues?
  - Fixed by running inspircd "on the host"
- Tests sit around forever after exiting, need to find what's holding the process open.
  - Bug in matrix-appservice-bridge, fixed in upstream but for now we just force exit. https://github.com/matrix-org/matrix-appservice-bridge/pull/439


src/irc/ClientPool.ts
```diff
@@ -146,8 +149,12 @@ export class ClientPool {
             await bridgedClient.connect();
         }
         catch (err) {
-            log.error("Bot failed to connect to %s : %s - Retrying....",
-                server.domain, JSON.stringify(err));
+            if (err instanceof IRCConnectionError && err.code === IRCConnectionErrorCode.Banned) {
+                throw Error('Bridge bot client is banned, cannot start');
```
Half-Shot, line 153: This would previously tightloop

src/bridge/IrcBridge.ts
```diff
@@ -1337,8 +1340,8 @@ export class IrcBridge {
     }
 
     public async connectToIrcNetworks(): Promise<void> {
-        await promiseutil.allSettled(this.ircServers.map((server) =>
-            Bluebird.cast(this.clientPool.loginToServer(server))
+        await Promise.all(this.ircServers.map((server) =>
```
Half-Shot, line 1343: We actually want to throw if this fails, otherwise we'd be running the IRC bridge with a possible configuration issue (if the bots cant connect).

src/bridge/IrcBridge.ts
```diff
@@ -544,9 +544,11 @@ export class IrcBridge {
             internalRoom = await this.dataStore.getAdminRoomByUserId("-internal-");
             if (!internalRoom) {
                 const result = await this.bridge.getIntent().createRoom({ options: {}});
+                console.log("Created room", result);
                 internalRoom = new MatrixRoom(result.room_id);
                 this.dataStore.storeAdminRoom(internalRoom, "-internal-");
             }
+            console.log(await this.bridge.getIntent().matrixClient.getJoinedRoomMembers(internalRoom.getId()));
```
Half-Shot, line 551: Naughty, should remove.

spec/integ/debug-api.spec.js
```diff
@@ -24,7 +24,7 @@ describe("Debug API", () => {
     it("should enable the debug API", async () => {
         // Revert to the default value for this one test
         expect(env.ircBridge.debugApi).not.toBeNull();
-        const res = await request(`http://localhost:${DEBUG_PORT}/version?access_token=${asToken}`);
+        const res = await (await crossFetch(`http://localhost:${DEBUG_PORT}/version?access_token=${asToken}`)).text();
```
Half-Shot, line 27: Just removing a dependency on the legacy request. Crossfetch is my fave since it's just `fetch` and soon enough Node will have that.

src/irc/IrcServer.ts
```diff
@@ -43,13 +43,13 @@ export interface IrcServerConfig {
     dynamicChannels: {
```
Half-Shot, line 43: A lot of these changes are making some of these configs optional, because we actually do want most of them to be optional.

spec/util/e2e-test.ts
```diff
@@ -0,0 +1,256 @@
...
+        console.log('Setting up IRC clients...');
+        await super.setUp(clients);
+        // Set up an IRC bridge.
+        this.dbPath = await mkdtemp('ircbridge-');
+
+        this.ircBridge = new IrcBridge({
+            homeserver: {
+                domain: this.homeserver.domain,
+                url: this.homeserver.url,
+                bindHostname: "0.0.0.0",
```
jaller94, line 88: Don't we expect these to always run on the same machine?
```suggestion
                bindHostname: "127.0.0.1",
```

spec/util/e2e-test.ts
```diff
@@ -0,0 +1,256 @@
...
+    }
+
+    public async waitForEvent<T>(
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        emitterType: string, filterFn: (...args: any[]) => T|undefined, timeoutMsg: string)
+    : Promise<T> {
+        return new Promise((resolve, reject) => {
+            // eslint-disable-next-line prefer-const
+            let timer: NodeJS.Timeout;
+            const fn = (...args: any[]) => {
```
jaller94, line 242: If it doesn't make TypeScript cry, `unknown` would be better.

```suggestion
        emitterType: string, filterFn: (...args: unknown[]) => T|undefined, timeoutMsg: string)
    : Promise<T> {
        return new Promise((resolve, reject) => {
            // eslint-disable-next-line prefer-const
            let timer: NodeJS.Timeout;
            const fn = (...args: unknown[]) => {
```

src/irc/IrcServer.ts
```diff
@@ -325,7 +325,7 @@ export class IrcServer {
     }
 
     public isInWhitelist(userId: string) {
-        return this.config.dynamicChannels.whitelist.includes(userId);
+        return !this.config.dynamicChannels.whitelist || this.config.dynamicChannels.whitelist.includes(userId);
```
jaller94, line 328: I find `!whitelist` pretty hard to understand.
```suggestion
        return this.config.dynamicChannels.whitelist?.includes(userId) ?? true;
```

jaller94 (approved):
A couple of fairly optional annotations, but in general this
Looks good to me.

Neustradamus:
@Half-Shot: Good job for IRC, can you merge for XMPP? :)

PR #1645, opened 2022-11-02
optionally make users wait to join bridged rooms

jesopo:
couple of `TODO`s still; looking for feedback on if i'm going in the right direction

### DIFF
--- a/src/bridge/MatrixHandler.ts
+++ b/src/bridge/MatrixHandler.ts
@@ -538,6 +538,35 @@ export class MatrixHandler {
             }
             // get the virtual IRC user for this user
             promises.push((async () => {
+                const entry = await this.ircBridge.getStore().getRoom(event.room_id, room.server.domain, room.channel);
+                // is this a portal room?
+                const delayTime = ["alias", "join"].includes(
+                    (entry?.data?.origin as string|null) ?? "unknown"
+                //TODO: pull these two numbers from the config file
+                ) ? 3600 : 0;
+
+                if (delayTime > 0) {
+                    let remaining = delayTime;
+                    const firstSeen = await this.ircBridge.getStore().getAccountFirstSeen(user.getId());
+                    if (firstSeen === null) {
+                        await this.ircBridge.getStore().setAccountFirstSeen(user.getId(), new Date());
+                    } else {
+                        remaining = Math.max(0, ((firstSeen.getTime() / 1000) + delayTime) - (new Date().getTime() / 1000));
+                    }
+
+                    if (remaining > 0) {
+                        await this.membershipQueue.leave(
+                            event.room_id,
+                            user.getId(),
+                            req,
+                            true,
+                            `Please wait ${remaining} seconds`,
+                            this.ircBridge.appServiceUserId,
+                        );
+                        return;
+                    }
+                }
+
                 let bridgedClient: BridgedClient|null = null;
                 try {
                     bridgedClient = await this.ircBridge.getBridgedClient(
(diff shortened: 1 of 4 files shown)


jesopo:
a more ideal future here would be one involving knocking, i.e. 60 minutes after you knock, you are invited, or something like that.

Yoric (commented):
As you mentioned, inviting people would probably make for better UX.

Is there a way to:
1. First time the user knocks, send them a message along the lines of "For user safety, this bridge is configured with a join delay. You will be allowed to join on ${date}."
2. After this delay, send an invite.

Maybe?

jesopo:
I don't believe knocking is widely implemented yet

src/datastore/DataStore.ts
```diff
@@ -192,5 +192,9 @@ export interface DataStore {
 
     getRoomCount(): Promise<number>;
 
+    getAccountFirstSeen(userId: string): Promise<Date|null>;
```
jesopo, line 195: I've noticed that there's a `data` json blob on users stored in postgres, but i also don't know how to get specifically at that blob of data, and i also don't know which db functions i should be using to specifically look up remote users

Half-Shot: Ugh the JSON blob is a bit of a sadness. 

For postgres, I'd probabaly crib from what we did [here](https://github.com/matrix-org/matrix-appservice-irc/blob/jess/delay-joins/src/datastore/postgres/schema/v7.ts) and create a new table that stores first seen times (OR: we just make the activity table store a first and latest time, which actually might be more sensible wdyt?)

For nedb: I hate NEDB. I'd crib from `updateLastSeenTimeForUser(userId: string)` and do exactly the same thing.

src/bridge/IrcBridge.ts
```diff
@@ -1072,8 +1072,13 @@ export class IrcBridge {
         if (userIds) {
             for (const userId of userIds) {
                 this.activityTracker.setLastActiveTime(userId);
-                this.dataStore.updateLastSeenTimeForUser(userId).catch((ex) => {
-                    log.warn(`Failed to bump last active time for ${userId} in database`, ex);
+                this.dataStore.getFirstSeenTimeForUser(userId).then((when) => {
```
Half-Shot, line 1075: I'd make this it's own function inside datastore personally, and have the logic for setting first seen time handled within. We might even be able to wrap the postgres stuff into one statement.

src/datastore/postgres/schema/v9.ts
```diff
@@ -0,0 +1,10 @@
+import { PoolClient } from "pg";
+
+export async function runSchema(connection: PoolClient) {
+    await connection.query(`
+        ALTER TABLE last_seen ADD COLUMN first BIGINT;
+        ALTER TABLE last_seen RENAME COLUMN ts TO last;
+        UPDATE last_seen SET first = last;
+        ALTER TABLE last_seen ALTER COLUMN first SET NOT NULL;
```
Half-Shot, line 8: That *might* explode for all the existing items that don't have a value set.

src/bridge/MatrixHandler.ts
```diff
@@ -538,6 +538,38 @@ export class MatrixHandler {
             }
             // get the virtual IRC user for this user
             promises.push((async () => {
+                const entry = await this.ircBridge.getStore().getRoom(event.room_id, room.server.domain, room.channel);
```
Half-Shot, line 541: This *sorta* feels like it could be it's own function, this async block is already chonky.

Half-Shot (changes requested):
Promising stuff, I've got some suggestions.

src/datastore/postgres/schema/v9.ts
jesopo: see the line above it; we backfill `first` from `last`

PR #1653, opened 2023-01-06
Update Dockerfile for stretch being EOL

benbz:
According to https://hub.docker.com/_/node, `16-slim` is now based on `debian:buster-slim`. Stretch isn't even `oldstable` anymore.


Half-Shot (approved):

Half-Shot:
Ahh, we don't have a Docker newsfile option. I'll make it a misc.

PR #1654, opened 2023-01-09
Fix port specified in config not being used in Docker

justinbot:
Fixes custom port specified in config under `homeserver.bindPort` not being used in Docker.
The run script for Docker was passing a port which would override this.

Also uses `node:16` instead of `node:16-slim` for the builder container because otherwise on ARM64 architecture it would error when trying to build `node-gyp` due to missing Python.


jaller94 (approved):
Looks good to me.

PR #1655, opened 2023-01-10
Refactor provisioning API

justinbot:
(Much of these changes derived from #1508)

Makes significant updates to the Provisioning API implementation, namely refactoring to use the `ProvisioningApi` class from `matrix-appservice-bridge`.

Also implements request validation and consistent error handling across the API.

PR #1656, opened 2023-01-23
Add new setup widget UI

justinbot:
Following #1655 which added support for widgets authenticating with the Provisioning API, this PR adds a new UI for configuring bridged rooms from within a widget:

Screenshot 2023-02-06 at 8 44 10 PM

Screenshot 2023-02-06 at 8 58 54 PM

This is an optional feature, enabled with the Provisioning API:
https://github.com/matrix-org/matrix-appservice-irc/blob/bb915f8ecfd4187ceb8ebd94c4b1a1cc4360d200/config.sample.yaml#L526-L528

## Implementation
At a high level, the widget is just a static webpage which interacts with the Provisioning API to perform configuration.

More specifically, this takes the same approach as the [setup widgets in `matrix-hookshot`](https://github.com/matrix-org/matrix-hookshot/blob/main/web/BridgeAPI.ts), but implemented with the consideration that much of it is generic and could be refactored out into a package for reuse in other bridges.

### Provisioning API
This part of the implementation is relevant for any API using the `ProvisioningApi` class from `matrix-appservice-bridge`. Includes:
* Parsing widget query parameters.
* Widget API setup and obtaining an OpenID token.
* Exchanging the OpenID token for a session token, storing and checking validity of the session token.
* Instantiating a generic Provisioning API client.

### Frontend
The same approach for building a frontend with Vite could be taken for other bridges, as well as the handful of reusable UI components which have been implemented with Tailwindcss.

### IRC Bridge
The IRC Provisioning API client and the actual configuration UI are specific to this IRC Bridge.


widget/src/IrcApp.tsx
```diff
@@ -0,0 +1,344 @@
...
+        { operatorNicks &&
+            <datalist id="operatorNicks">
+                { operatorNicks.map(nick => <option value={nick} key={nick}></option>) }
+            </datalist>
+        }
+        <Forms.Input
+            label="Channel key (optional)"
+            type="text"
+            value={channelKey}
+            onChange={e => setChannelKey(e.target.value)}
```
jaller94, line 251: useCallback to minimize DOM changes.

widget/src/components/buttons/Danger.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'preact';
+import classNames from 'classnames';
+
+const Danger = (props: React.ComponentProps<'button'>) =>
```
jaller94, line 4: To avoid the type cast below:
```suggestion
const Danger = (props: React.ComponentProps<'button'> & {className: string}) =>
```

justinbot: This was needed due to bad typings on `preact`.
Smaller bundle size would have been nice, but given the shortfalls I switched to `react` as the more reliable choice (especially as these components could be reused for other bridge UIs).

widget/src/IrcApp.tsx
```diff
@@ -0,0 +1,357 @@
...
+    client: IrcProvisioningClient
+    roomId: string,
+}) => {
+    const [error, setError] = useState('');
+    const [links, setLinks] = useState<ListLinksResponse>();
+
+    const getLinks = useCallback(async () => {
+        setError('');
+        try {
+            const _links = await client.listLinks(roomId);
```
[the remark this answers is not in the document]
justinbot: This repo has a linter rule that you can't shadow var names (otherwise I would just use the same name as the state).
I've often seen `_` prefix used in JS to indicate private vars so it seemed to make sense here.

widget/src/IrcApp.tsx
```diff
@@ -0,0 +1,357 @@
...
+            );
+        }
+    }, [client, roomId]);
+
+    useEffect(() => {
+        getLinks();
+
+        // Continue to poll linked channels
+        const interval = setInterval(async() => {
+            await getLinks();
```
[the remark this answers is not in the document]
justinbot: Addressed using recursive `setTimeout` instead, waiting for the previous call to finish so they don't overlap (e.g. if the network was slow)

widget/src/IrcApp.tsx
```diff
@@ -0,0 +1,369 @@
...
+            }
+        };
+        getNetworks();
+    }, [client]);
+
+    const [selectedServer, setSelectedServer] = useState('');
+
+    let content;
+    if (networks) {
+        if (networks.servers.length > 0) {
```
Half-Shot, line 320: My cheeky suggestion is that if the server length is 1 (which is true in nearly all cases), we should just render this as a readonly field. Something to make it look like not a dropdown.

widget/src/IrcApp.tsx
```diff
@@ -0,0 +1,369 @@
...
+
+import { useProvisioningContext } from './ProvisioningApp';
+import { ProvisioningError } from './ProvisioningClient';
+import { IrcProvisioningClient, ListLinksResponse, QueryNetworksResponse } from './IrcProvisioningClient';
+import * as Text from './components/text';
+import * as Buttons from './components/buttons';
+import * as Forms from './components/forms';
+import * as Alerts from './components/alerts';
+
+const pollLinkIntervalMs = 2000;
```
Half-Shot, line 11: 2s seems kinda high?

widget/src/IrcApp.tsx
```diff
@@ -0,0 +1,369 @@
...
+    }
+    else if (error) {
+        content = <Alerts.Danger>{ error }</Alerts.Danger>;
+    }
+    else {
+        content = <Text.Caption>Loading...</Text.Caption>;
+    }
+
+    return <div className="mb-6">
+        <Text.Title>Link a new channel</Text.Title>
```
Half-Shot, line 355: I feel like we need some text here to explain the approval process etc.

Half-Shot (changes requested):
This is actually looking amazing, you should be very proud of this work! I have a few minor concerns, but the bulk of this is great.

Half-Shot:
One final thought @justinbot, can you write some documentation on this feature in the hosted docs. It can probably be similar to the hookshot ones. Thanks!

widget/src/IrcApp.tsx
justinbot: Fixed in matrix-org/matrix-appservice-bridge#458

docs/setup_widget.md
```diff
@@ -0,0 +1,25 @@
...
+    # Whether to enable hosting the setup widget page. Default: false.
+    widget: true
+```
+It will be hosted on the same port as the appservice by default, but a different port can be specified.
+
+### Usage
+
+When enabled, the widget will be available at `/_matrix/provision/v1/static` by default and can be added to a Matrix room like this:
+```
+/addwidget https://example.com/_matrix/provision/v1/static/?roomId=$matrix_room_id&widgetId=$matrix_widget_id
```
Half-Shot, line 22: *where https://example.com is a public route to your IRC bridge's provisioning API

Half-Shot (approved):
Smashing, one note in the docs then this is ready to ship I feel :)

PR #1660, opened 2023-02-16
PgDataStore: Fix syntax of array parameters

heftig:
Arrays passed as query parameters become PostgreSQL arrays. However, `foo IN $1` is not valid syntax, as `IN` can only be used with subqueries (`foo IN (SELECT ...)`) and lists (`foo IN ($1, $2)`), which require parentheses.

`foo IN ($1)` is interpreted as a list and the expression is compared with the whole array, not checked against its members.

The correct syntax for checking if a value matches any member of an array parameter is `foo = ANY($1)`, described in
https://www.postgresql.org/docs/15/functions-comparisons.html#id-1.5.8.30.16.

Fix `getIrcChannelsForRoomIds` and `getRoomsVisibility`. While we're at it, replace `getMappingsForChannelByOrigin`'s workable but awkward building of a parameter list.


heftig:
Context: I stumbled on this because the bridge logs an error for each configured channel when starting:

```
ERROR:PublicitySyncer Failed to sync publicity for #foobar: syntax error at or near "$1"
```

Half-Shot:
N.B. Commit contains sign off.

Half-Shot:
Merged via #1698, thanks!

PR #1663, opened 2023-02-17
Define power levels to disable calls/reactions/redaction/stickers in PMs initiated from IRC

progval:
No one but the sender can see them because we cannot bridge them. Blocking them with PLs allows the clients to hide these features from their UI, so users do not mistakenly believe they will be received.


progval:
Signed-off-by: Valentin Lorentz <progval plus git at progval dot net>

progval:
ping?

Half-Shot (changes requested):
Looks fine, but needs to be targeting `develop`.

Half-Shot (approved):
Seems okay

PR #1671, opened 2023-03-04
config.sample.json: Document the meaning of both options in `bridgeInfoState`

progval:


progval:

PR #1672, opened 2023-03-05
Rename infoMapping to initialEvent

progval:
The info mapping is the second argument to createInitialState, not what it returns


progval:
Signed-off-by: Valentin Lorentz \<progval plus git at progval dot net\>

Half-Shot (approved):
No argument from me

PR #1555, opened 2022-04-18
Support monospace formatting

progval:
Signed-off-by: Valentin Lorentz \<progval plus git at progval dot net\>

PR #1682, opened 2023-03-11
Fix changelog

progval:
We created the wrong file name in #1555


progval:
Signed-off-by: Valentin Lorentz <progval plus matrixsignoff at progval dot net>

PR #1683, opened 2023-03-11
Implement MSC2346 (bridge info state event) for PMs

progval:
This PR depends on the following:

* https://github.com/matrix-org/matrix-appservice-bridge/pull/460 for typing
* https://github.com/matrix-org/matrix-appservice-irc/pull/1681 to allow tests to turn MSC2346 on

PR #1681, opened 2023-03-11
Make bridgeInfoState configuration reloadable

progval:
This will be useful for tests in #1683

PR #1686, opened 2023-03-12
Add support for MSC3968 ('poorer features')

progval:
[MSC3968](https://github.com/matrix-org/matrix-spec-proposals/pull/3968)

This includes a few commits from #1683 (those before the merge commit) because it would be a very messy merge conflict otherwise.

Depends on:

*  https://github.com/matrix-org/matrix-appservice-bridge/pull/460 (transitively, because of the above)
* https://github.com/matrix-org/matrix-appservice-irc/pull/1681 (ditto)
* https://github.com/matrix-org/matrix-appservice-bridge/pull/463

There are new tests only for PM rooms, because provisioning for non-PMs does not have unit tests


progval:

Half-Shot:
Thanks for this, it looks like something we had on our radar a while back and I'm glad you've picked up. I'll try to get a review in of these changes, because the idea is ace.

progval:
The merge conflicts of this PR are inherited from #1683; I fixed them there and will rebase this PR once #1683 is merged.

PR #1697, opened 2023-04-12
Handle thread fallbacks correctly

tadzik:
We were incorrectly handling these as regular messages because they didn't look like our idea of what a reply looks like on Matrix.


Half-Shot (approved):
Saddens me that we *still* have to parse these bodies with Regex. Also, I'm wondering if the bridge-sdk should be providing this functionality. However, won't block the PR on it.

PR #1698, opened 2023-04-12
Add tests to ensure publicity syncing works for +s modes

Half-Shot:
Extends #1660 to include tests which cover those functions.


tadzik (approved):
No complaints!

mweinelt:
Does this mean that matrix-appservice-irc before 0.38.0 does leak +s rooms into the channel directory?

PR #1700, opened 2023-04-20
changelog: Mention fixes to +s channel leaks

progval:
https://github.com/matrix-org/matrix-appservice-irc/releases/tag/0.38.0 will need to be updated accordingly if this is accepted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,8 +4,7 @@
 Bugfixes
 --------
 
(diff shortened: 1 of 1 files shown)


progval:

Half-Shot:
I've modified this a bit so we combine the lines (as really it's the same bug with two causes), and toned the language a bit for users/administrators. Otherwise, looks good to me. Thanks for keeping us honest!

mweinelt:
Thanks for clarifying that.

progval:
Thanks!

PR #1701, opened 2023-04-26
Support Node 20 (and drop Node 16)

Half-Shot:
See https://github.com/matrix-org/matrix-appservice-bridge/pull/466.

This also moves us onto yarn, to be in line with the rest of our ecosystem.


Dockerfile
```diff
@@ -8,20 +8,22 @@ RUN git clone https://github.com/matrix-org/freebindfree.git
...
 COPY src/ /build/src/
 COPY widget/ /build/widget/
-COPY package.json package-lock.json tsconfig.json .eslintrc /build/
+COPY package.json yarn.lock tsconfig.json .eslintrc /build/
 
-RUN npm ci
-RUN npm run build
-RUN npm prune --omit dev
+RUN yarn --strict-semver --frozen-lockfile
+RUN yarn build
```
AndrewFerr: I believe this is redundant since `package.json`'s `"prepare"` step runs a build.

package.json
```diff
@@ -5,7 +5,7 @@
   "main": "app.js",
   "bin": "./bin/matrix-appservice-irc",
   "engines": {
-    "node": ">=16"
+    "node": ">=18"
   },
   "scripts": {
     "prepare": "npm run build",
```
AndrewFerr: Can this (and other scripts) now use their `yarn` equivalents?

AndrewFerr (changes requested):
Looks good, just a few minor requests.

And FWIW the lint script fails on me locally until I remove the `import urlJoin` line in `widget/src/ProvisioningApp.tsx`, so if the next CI run fails on that, it should be fixable.

AndrewFerr:
Hmm, apparently the E2E tests fail non-deterministically...?

- Failed run: https://github.com/matrix-org/matrix-appservice-irc/actions/runs/4816225419/jobs/8575651161?pr=1701#step:14:817
- Passed re-run: https://github.com/matrix-org/matrix-appservice-irc/actions/runs/4816225419/jobs/8575756745#step:14:825

AndrewFerr (approved):
As long as:
- intermittent failures in the E2E CI tests are acceptable, and
- it's fine to no longer override the `matrix-bot-sdk` dependency only in `m-a-b`'s deps,

then this should be good to go.

PR #1702, opened 2023-04-26
Show helpful admin command error if it's a IrcProvisioningError

Half-Shot:
We now throw helpful error types when failing to run some provisioning commands but we are still showing "check logs for details". I, like many people get really pissed off when applications make me "check logs for details". For the sake of security this still only emits a error message in the case where we know the type of error.


AndrewFerr (approved):
LGTM

PR #1703, opened 2023-04-26
Fix documentation builds and pull in version picker

Half-Shot:
This largely cribs off the hookshot workflows, and blatantly steals the version picker code.


.github/workflows/docs.yml
```diff
@@ -1,43 +0,0 @@
-name: Build docs
-
-on:
-  push:
-    branches:
-      - master
```
Half-Shot, line 6: This was broken because we don't use `master` anymore.

.github/workflows/docker-hub-latest.yml
```diff
@@ -4,7 +4,8 @@ name: "Docker Hub - Latest"
 
 on:
   push:
```
Half-Shot, line 6: I believe push on it's own excludes forks, which we don't want to do.

AndrewFerr:
Are the E2E test failures the same ones caused by Redis sending an error on shutdown?

Half-Shot:
This flaked but I'm not sure why :/

PR #1704, opened 2023-04-26
Allow hot-reloading of the bridge logger

Half-Shot:
We already had it for the matrix-appservice-bridge logger but not the legacy logger. At some point we need to remove it, but for now this will do.


AndrewFerr (approved):
Looks simple enough

Issue #1695, opened 2023-03-28
libera.chat missing from "Bridges" section of "Add integrations"


allisonkarlitskaya:
I want to setup a new plumbed connection between a Matrix channel and an IRC channel on libera chat.  Using Element in the Matrix room, I click on "Add widgets, bridges & bots" and enable scalar.vector.im as my integrations manager.

I now see the ability to add some widgets, bidges for Slack, GIMPNet, OFTC, and some others, and couple of bots (RSS, GitHub) but no mention of Libera.chat anywhere.

According to everyone I've talked to and everything I've read online, this is the place to go to setup a new bridge.

allisonkarlitskaya:
Probably should have attached this in the original report: this is what I've seen for the past several days:

![image](https://user-images.githubusercontent.com/36541154/228441026-bee894ce-68ac-4e8a-828f-5b244fd56c26.png)

progval:
Not sure, but this might be a bug in https://github.com/vector-im/element-integration-manager/

allisonkarlitskaya:
I guess either way, @Half-Shot will have an opinion about where this issue belongs.

sanecito:
> According to everyone I've talked to and everything I've read online, this is the place to go to setup a new bridge.

I can confirm that this is how I've regularly set up bridges between Matrix and Libera.chat and that I don't see it either.

**Software info:**
Element version: 1.11.28
Olm version: 3.2.12

Half-Shot:
Hiya, this is on our roadmap to fix very soon. Unfortunately we're having to time restarts of the bridge quite carefully to avoid disruption.

sukso96100:
Hi, any updates regarding this issue? I'm waiting for Linera.chat bridge option to appear on integration manager screen :)

Half-Shot:
We're expecting this to happen when we deploy the 1.0.0 update in a few days or so.

Half-Shot:
Now fixed.

davidebeatrici:
Still missing:

![](https://github.com/matrix-org/matrix-appservice-irc/assets/5897523/914efd04-ccb8-4944-a2db-5e845a0e20b2)

progval:
@davidebeatrici different issue, the bridge is "temporarily" disabled since early august: https://libera.chat/news/matrix-bridge-disabled-retrospective

PR #1706, opened 2023-04-27
Fix linking rooms from setup widget when a `networkId` is configured

justinbot:
Fixes an error when trying to link a room on a server which has a `networkId` configured, by making the request with the domain instead of the `networkId` (which defaults to the domain if not configured).


progval:
Is this related to #1695 or afh's report on #matric-irc on 2023-04-24?

Half-Shot (approved):

PR #1709, opened 2023-05-04
Fix a couple of issues in the RC

Half-Shot:
This is:

- Actually providing a hostname to the pool.
- Remembering to return TLS sockets.

### DIFF
--- a/src/pool-service/IrcConnectionPool.ts
+++ b/src/pool-service/IrcConnectionPool.ts
@@ -74,7 +74,6 @@ export class IrcConnectionPool {
     }
 
     private async createConnectionForOpts(opts: ConnectionCreateArgs): Promise<Socket> {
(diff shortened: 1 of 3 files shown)


src/pool-service/IrcConnectionPool.ts
```diff
@@ -89,11 +88,12 @@ export class IrcConnectionPool {
                 };
             }
 
-            socket = await new Promise((resolve, reject) => {
+            return await new Promise((resolve, reject) => {
```
AndrewFerr, line 91: Is it possible for line 122 to be reached without any of the `resolve`s having been hit?

Half-Shot: 122 yes, 127 I don't think so.

PR #1715, opened 2023-05-10
Sort by channel name in !listrooms

ilmari:
It's stored as a set in the client object, which doesn't guarantee anything about the order, so sort it for consistent output.


Half-Shot (approved):
Looks good, just needs a signoff

ilmari:
> Looks good, just needs a signoff

Done

PR #1719, opened 2023-05-13
Allow usernames to include more characters when using the `!username` command

### LINKED ISSUE #1359: Can't store username that contains underscore

progval:
Fixes https://github.com/matrix-org/matrix-appservice-irc/issues/1359
Supercedes https://github.com/matrix-org/matrix-appservice-irc/pull/1367 and https://github.com/matrix-org/matrix-appservice-irc/pull/1538, which were both too restrictive

This change allows for a much more permissive username when users are setting their own username.


progval:

src/bridge/AdminRoomHandler.ts
```diff
@@ -39,6 +39,15 @@ enum CommandPermission {
...
 const SANE_USERNAME_LENGTH = 64;
 
+// Technically, anything but \0 is allowed as username (aka. authcid and authzid):
+// https://www.rfc-editor.org/rfc/rfc4616#section-2
+//
+// However, IRC services are very unlikely to allow the username to contain CR or LF
+// (because they would not fit in the wire format) or spaces (because usernames are
+// usually followed by passwords in "PRIVMSG NickServ :REGISTER" commands and
+// IRCv3 draft/account-registration.
+const SASL_USERNAME_INVALID_CHARS_PATTERN = /[ \0\r\n]+/;
```
Half-Shot: Wondering if we at least limit to printable characters. I can't see the value in allowing most of these.

```suggestion
const SASL_USERNAME_INVALID_CHARS_PATTERN = /[\0-\38\7F]+/;
```

src/bridge/AdminRoomHandler.ts
```diff
@@ -503,10 +513,10 @@ export class AdminRoomHandler {
                     `Username is longer than the maximum permitted by the bridge (${SANE_USERNAME_LENGTH}).`
                 );
             }
-            else if (IdentGenerator.sanitiseUsername(username) !== username) {
+            else if (invalidChars !== null) {
                 notice = new MatrixAction(
                     ActionType.Notice,
-                    `Username contained invalid characters not supported by IRC.`
+                    `Username contained invalid characters not supported by IRC (${JSON.stringify(invalidChars.join(""))}).`
```
Half-Shot: Wonder if we should make the unprintable, printable?

Half-Shot (commented):
Looking really nice, just a couple of nits.

src/bridge/AdminRoomHandler.ts
progval: I think you mean `[\0-\40\x7F]` (38 isn't valid octal, the octal for space is 40, and 7F is the hex code for DEL, not the octal code).

src/bridge/AdminRoomHandler.ts
progval: Doesn't `JSON.stringify` do it?

src/bridge/AdminRoomHandler.ts
Half-Shot: yeah, bad copy paste. That one ^

src/bridge/AdminRoomHandler.ts
Half-Shot: So it does!

Half-Shot:
Assuming tests pass this will be good enough for a merge.

Half-Shot (approved):

Half-Shot:
I'm happy with the contrib requirements and I believe docker builds are broken for non-collaborators atm (due to needing a login). @progval if you can make the lint happy I'll chuck this in the release tomorrow.

Half-Shot:
:star: let's go!

PR #1720, opened 2023-05-15
Add passkey checker on startup.

### LINKED ISSUE #1718: Check if the passkey is able to unlock user passwords on startup
If you pass in a passkey that is wrong, the bridge will fail to connect users and **kick them from rooms** if the passkey isn't valid. We should:

A) Check the validity of the passkey by doing an unlock of any password in the datastore.
B) Ideally not kick users on password unlock fails.

The kick failure is that we are unable to load the ClientConfig, and the failure triggers the client kick code.

Half-Shot:
Fixes #1718

### DIFF
--- a/src/bridge/IrcBridge.ts
+++ b/src/bridge/IrcBridge.ts
@@ -617,7 +617,6 @@ export class IrcBridge {
             const pgDs = new PgDataStore(this.config.homeserver.domain, dbConfig.connectionString, pkeyPath);
             await pgDs.ensureSchema();
             this.dataStore = pgDs;
-
         }
         else if (dbConfig.engine === "nedb") {
             await this.bridge.loadDatabases();
--- a/src/bridge/IrcBridge.ts
+++ b/src/bridge/IrcBridge.ts
@@ -618,9 +618,6 @@ export class IrcBridge {
             await pgDs.ensureSchema();
             this.dataStore = pgDs;
 
-            if (pkeyPath) {
-                await pgDs.ensurePasskeyCanDecrypt();
-            }
         }
         else if (dbConfig.engine === "nedb") {
             await this.bridge.loadDatabases();
@@ -653,6 +650,7 @@ export class IrcBridge {
             throw Error("Incorrect database config");
         }
 
+        await this.dataStore.ensurePasskeyCanDecrypt();
 
         await this.dataStore.removeConfigMappings();
 
--- a/src/bridge/IrcBridge.ts
+++ b/src/bridge/IrcBridge.ts
@@ -617,6 +617,10 @@ export class IrcBridge {
             const pgDs = new PgDataStore(this.config.homeserver.domain, dbConfig.connectionString, pkeyPath);
             await pgDs.ensureSchema();
             this.dataStore = pgDs;
+
+            if (pkeyPath) {
+                await pgDs.ensurePasskeyCanDecrypt();
+            }
         }
         else if (dbConfig.engine === "nedb") {
             await this.bridge.loadDatabases();
@@ -648,6 +652,7 @@ export class IrcBridge {
         else {
             throw Error("Incorrect database config");
         }
+
 
         await this.dataStore.removeConfigMappings();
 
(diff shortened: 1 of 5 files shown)


src/bridge/IrcBridge.ts
```diff
@@ -617,6 +617,7 @@ export class IrcBridge {
             const pgDs = new PgDataStore(this.config.homeserver.domain, dbConfig.connectionString, pkeyPath);
             await pgDs.ensureSchema();
             this.dataStore = pgDs;
+
```
AndrewFerr:
```suggestion
```

src/datastore/NedbDataStore.ts
```diff
@@ -610,6 +610,29 @@ export class NeDBDataStore implements DataStore {
...
+        });
+        for (const { id: userId, data } of docs) {
+            for (const [domain, clientConfig] of Object.entries(data.client_config)) {
+                if (clientConfig.password) {
+                    try {
+                        this.cryptoStore.decrypt(clientConfig.password);
+                    }
+                    catch (ex) {
+                        log.error(`Failed to decrypt password for ${userId} on ${domain}`, ex);
+                        throw Error('Cannot decrypt user password, refusing to continue');
```
AndrewFerr: How about setting [`Error.cause`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause#examples)?
```suggestion
                        throw Error('Cannot decrypt user password, refusing to continue', { cause: ex });
```

src/bridge/IrcBridge.ts
```diff
@@ -649,6 +650,8 @@ export class IrcBridge {
             throw Error("Incorrect database config");
         }
 
+        await this.dataStore.ensurePasskeyCanDecrypt();
```
AndrewFerr, line 652: If this throws, the bridge aborts startup altogether, right?

src/datastore/postgres/PgDataStore.ts
```diff
@@ -545,6 +551,24 @@ export class PgDataStore implements DataStore, ProvisioningStore {
...
+        }
+        const res = await this.pgPool.query<{password: string, user_id: string, domain: string}>(
+            "SELECT password, user_id, domain FROM client_config WHERE password IS NOT NULL");
+        for (const { password, user_id, domain } of res.rows) {
+            try {
+                this.cryptoStore.decrypt(password);
+            }
+            catch (ex) {
+                log.error(`Failed to decrypt password for ${user_id} on ${domain}`, ex);
+                throw Error('Cannot decrypt user password, refusing to continue');
```
AndrewFerr: Ditto re: `Error.cause`.

src/datastore/postgres/PgDataStore.ts
```diff
@@ -516,7 +516,13 @@ export class PgDataStore implements DataStore, ProvisioningStore {
         const row = res.rows[0];
         const config = row.config || {}; // This may not be defined.
         if (row.password && this.cryptoStore) {
-            config.password = this.cryptoStore.decrypt(row.password);
+            // NOT fatal, but really worrying.
```
AndrewFerr, line 519: 1. Why is this not fatal here?
2. Should `NedbDataStore.getIrcClientConfig` also wrap its decryption call with a try-catch?

AndrewFerr (changes requested):
Have some minor questions/requests (requestions?)

src/bridge/IrcBridge.ts
Half-Shot: Yep!

src/datastore/NedbDataStore.ts
Half-Shot: I like this!

src/datastore/postgres/PgDataStore.ts
Half-Shot: 1. Failing to load a password shouldn't stop you from being able to use the bridge, but it will annoy the user a bit.

AndrewFerr (approved):

Issue #1731, opened 2023-06-02
"No channel data for channel/nick" errors when trying to handle mode changes


Half-Shot:
We fetch the BridgedClient of the user being opped, and then get the prefix's for that username. It sounds like the `.chanData` request on the client is failing, and this is causing the bridge to not handle the mode change. I'd expect to see some data for this channel as we only emit a `+mode` if we know the channel. https://github.com/matrix-org/node-irc/blob/985b3380857e5d816934bf1c1bfc44ed60e70081/src/irc.ts#L513-L517

Half-Shot:
This should be fixed in #1732 .

Issue #1725, opened 2023-05-23
Making myself an op in a bridged IRC channel does not make me a moderator of the Matrix channel


daviwil:
Apologies if this has been reported before, I wasn't able to find my exact issue.

I've got a matrix.org room (!EoRhMvNpnWxCMTMPeP:libera.chat) bridged over to a libera.chat channel (#systemcrafters).  The bridged room was probably created a year or more ago (room version is 6).  I am an op of #systemcrafters but for some reason I am not a moderator of the Matrix channel.  If I use ChanServ to deop and re-op myself, my power level is not set to moderator level.

However, I created a bridged channel for a new libera.chat channel last week and making myself in op in that channel *does* give me moderator power level.

Is there something wrong with the channel configuration?  Is there anyway to reset its state using appservice?

Half-Shot:
Hey there, this seems to be caused by https://github.com/matrix-org/matrix-appservice-irc/issues/1731.

Half-Shot:
@daviwil are you still seeing this?

PR #1732, opened 2023-06-02
Add a test for mode changes correctly applying to users.

Half-Shot:
As well as adding some improved logging around the area.

### DIFF
--- a/src/bridge/IrcHandler.ts
+++ b/src/bridge/IrcHandler.ts
@@ -906,7 +906,7 @@ export class IrcHandler {
      * @return {Promise} which is resolved/rejected when the request finishes.
      */
     public async onMode(req: BridgeRequest, server: IrcServer, channel: string, by: string,
-                        mode: string, enabled: boolean, arg: string|null): Promise<BridgeRequestErr|void> {
+                        mode: string, enabled: boolean, arg: string|null): Promise<BridgeRequestErr|undefined> {
         this.incrementMetric("mode");
         req.log.info(
             "onMode(%s) in %s by %s (arg=%s)",
--- a/src/bridge/IrcHandler.ts
+++ b/src/bridge/IrcHandler.ts
@@ -906,14 +906,14 @@ export class IrcHandler {
      * @return {Promise} which is resolved/rejected when the request finishes.
      */
     public async onMode(req: BridgeRequest, server: IrcServer, channel: string, by: string,
-                        mode: string, enabled: boolean, arg: string|null) {
+                        mode: string, enabled: boolean, arg: string|null): Promise<BridgeRequestErr|void> {
         this.incrementMetric("mode");
         req.log.info(
             "onMode(%s) in %s by %s (arg=%s)",
             (enabled ? ("+" + mode) : ("-" + mode)),
             channel, by, arg
         );
-        await this.roomAccessSyncer.onMode(req, server, channel, by, mode, enabled, arg);
+        return this.roomAccessSyncer.onMode(req, server, channel, by, mode, enabled, arg);
     }
 
     /**
--- a/src/bridge/RoomAccessSyncer.ts
+++ b/src/bridge/RoomAccessSyncer.ts
@@ -1,6 +1,6 @@
 import { getLogger } from "../logging";
 import { IrcBridge } from "./IrcBridge";
-import { BridgeRequest } from "../models/BridgeRequest";
+import { BridgeRequest, BridgeRequestErr } from "../models/BridgeRequest";
 import { IrcServer } from "../irc/IrcServer";
 import { MatrixRoom, PowerLevelContent } from "matrix-appservice-bridge";
 import { BridgedClientStatus } from "../irc/BridgedClient";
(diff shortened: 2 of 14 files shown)


progval:
Shouldn't the changelog mention it's a bugfix for #1731? (and #1725?)

Half-Shot:
Test failures seem related to concurrency, again.

src/bridge/RoomAccessSyncer.ts
```diff
@@ -310,9 +318,18 @@ export class RoomAccessSyncer {
...
+                req.log.info(`Granting PL${level} to ${userId} in ${roomId}`);
+                await this.setPowerLevel(roomId, userId, level, req);
+            }
+            catch (ex) {
+                req.log.warn(`Failed to grant PL in ${roomId}`, ex);
+                didFail = true;
+            }
+        }
+        if (didFail) {
+            throw new Error('Failed to update PL in soime rooms');
```
AndrewFerr:
```suggestion
            throw new Error('Failed to update PL in some rooms');
```
...to ease log searching.

src/bridge/IrcHandler.ts
```diff
@@ -906,14 +906,14 @@ export class IrcHandler {
      * @return {Promise} which is resolved/rejected when the request finishes.
      */
     public async onMode(req: BridgeRequest, server: IrcServer, channel: string, by: string,
-                        mode: string, enabled: boolean, arg: string|null) {
+                        mode: string, enabled: boolean, arg: string|null): Promise<BridgeRequestErr|void> {
```
AndrewFerr: Nitpick: since the return type is sometimes used, should `void` be `undefined` instead?  That would also let it match the return type of `RoomAccessSyncer.onMode`.

This applies to `complete` in `IrcEventBroker.ts` as well.

On that note, some calls of `complete` (all in `IrcEventBroker.ts`) are neither `return`ed nor `await`ed -- is that intentional?

src/bridge/RoomAccessSyncer.ts
```diff
@@ -215,7 +214,8 @@ export class RoomAccessSyncer {
 
         // Bridge usermodes to power levels
         const modeToPower = server.getModePowerMap();
-        if (!Object.keys(modeToPower).includes(mode)) {
+        if (typeof modeToPower[mode] !== "number") {
```
AndrewFerr: Given `modeToPower` is typed as `{[mode: string]: number}`, is it sufficient to just check for `mode in modeToPower`?  Or was there a case where a non-number would sneak in?

src/bridge/RoomAccessSyncer.ts
AndrewFerr: How about also storing which exception was thrown (if any), and setting it as the `cause` of this `Error`?

src/irc/BridgedClient.ts
```diff
@@ -808,7 +808,6 @@ export class BridgedClient extends EventEmitter {
                                 identResolver: () => void) {
         // If this state has carried over from a previous connection, pull in any channels.
         [...connInst.client.chans.keys()].forEach(k => this.chanList.add(k));
-        console.log('Adding existing channels', this.chanList.entries());
```
AndrewFerr, line 811: Any reason why this was removed?  Was it too noisy?

src/bridge/RoomAccessSyncer.ts
```diff
@@ -226,7 +226,7 @@ export class RoomAccessSyncer {
         );
         if (matrixRooms.length === 0) {
             req.log.info("No mapped matrix rooms for IRC channel %s", channel);
-            return;
+            return BridgeRequestErr.ERR_NOT_MAPPED;
```
AndrewFerr, line 229: Given the pattern of returning undefined vs an error enum appears elsewhere, I'm confident that this works, but where are these enum values handled?

AndrewFerr (changes requested):

src/bridge/IrcHandler.ts
Half-Shot: Undefined makes more sense, yeah. At the risk of potentially ignoring a problem, I'm going to avoid diving into IrcEventBroker right now. It's fragile but works.

src/bridge/RoomAccessSyncer.ts
Half-Shot: I think that's sufficient yep, lemme change it.

src/irc/BridgedClient.ts
Half-Shot: Console.log is disallowed except for debugging, this shouldn't have landed in develop. Also, it's more noise than helpful now we can just inspect the bridge via redis.

src/bridge/RoomAccessSyncer.ts
Half-Shot: It's ultimately handled in https://github.com/matrix-org/matrix-appservice-irc/blob/3e2dabe2be11cb7fd3065c69ad1a51d11431ecb4/src/bridge/IrcBridge.ts#L889.

src/bridge/RoomAccessSyncer.ts
```diff
@@ -214,7 +214,7 @@ export class RoomAccessSyncer {
 
         // Bridge usermodes to power levels
         const modeToPower = server.getModePowerMap();
-        if (typeof modeToPower[mode] !== "number") {
+        if (mode in modeToPower === false) {
```
AndrewFerr, line 217: TIL that `in` has higher precedence than equality operators: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence#table

AndrewFerr (approved):

PR #1737, opened 2023-06-05
Add simple client discovery phase for pool configurations

Half-Shot:
The intention here is to basically iterate through all currently connected clients and reconnect them, as we've found that despite doing a room membership scan on startup to catch all matrix users, we were missing a few of them each time.


src/irc/ClientPool.ts
```diff
@@ -128,6 +128,34 @@ export class ClientPool {
...
+     * Discover any clients connected from a previous session when
+     * using the pool
+     */
+    public async discoverPoolConnectedClients() {
+        if (!this.redisPool) {
+            log.warn("discoverPoolConnectedClients called but redisPool wasn't set. No clients discovered");
+            return;
+        }
+
+        for await (const connection of this.redisPool.getPreviouslyConnectedClients()) {
```
Half-Shot, line 140: N.B. This loop is slow at scale, we should run this while also fetching rooms on startup.

src/bridge/MemberListSyncer.ts
```diff
@@ -301,36 +307,34 @@ export class MemberListSyncer {
...
+
+                // We only need to delay if the connection is "fresh" i.e. makes an impact on the ircd.
+                // If this is just reestablising a connection to the proxy then we can go as fast as
+                // we like!
+                if (freshConnection) {
+                    await promiseutil.delay(this.server.getMemberListFloodDelayMs());
+                }
+            }
+            catch {
                 // discard error, this will be due to timeouts which we don't want to log
```
AndrewFerr: This seems like an unreliable guarantee to uphold, particularly because `injectJoinFn` is set elsewhere.  Can we at least `log.debug` the error?

src/bridge/MemberListSyncer.ts
```diff
@@ -301,36 +307,34 @@ export class MemberListSyncer {
...
-            }).catch(() => {
+            try {
+                const freshConnection =
+                    await injectJoinFn(entry.roomId, entry.userId, entry.displayName, entry.frontier);
+
+                // We only need to delay if the connection is "fresh" i.e. makes an impact on the ircd.
+                // If this is just reestablising a connection to the proxy then we can go as fast as
+                // we like!
+                if (freshConnection) {
+                    await promiseutil.delay(this.server.getMemberListFloodDelayMs());
```
AndrewFerr: Is this delay literally just a delay between join attempts?  Are any of the joins for ircd, or only for Matrix?

Half-Shot: These are joins *to* the IRCD from Matrix, and for cold starts this will impact the ircd (hence the need for a delay).

My previous version of the code cocked up though and delayed all connections for a few MS, when actually we should be racing (and doing these concurrently).

src/irc/ClientPool.ts
Half-Shot: Now completed in [9c574a9](https://github.com/matrix-org/matrix-appservice-irc/pull/1737/commits/9c574a9181b471220eb1179a0e4939e632a457fe)

src/bridge/MemberListSyncer.ts
Half-Shot: The previous code didn't log, but tend to agree now that we don't have the possibility of a timeout error we should just carry on.

src/bridge/IrcBridge.ts
```diff
@@ -828,6 +836,7 @@ export class IrcBridge {
             log.error(err.stack);
         });
 
+        await discoveringClientsPromise;
```
AndrewFerr: Is this redundant with awaiting on `memberlistPromises` since it contains Promises that were `then`ed on `discoveringClientsPromise`?

src/bridge/MemberListSyncer.ts
```diff
@@ -321,20 +334,24 @@ export class MemberListSyncer {
...
+                // Inject a join to connect the user. We wait up til the delay time,
+                // and then just connect the next user.
+                // If the user connects *faster* than the delay time, then we need
+                // to delay for the remainder.
+                const delayPromise = promiseutil.delay(floodDelayMs);
+                await Promise.race([
+                    delayPromise,
+                    this.injectJoinFn(entry.roomId, entry.userId, entry.displayName, entry.frontier),
+                ]);
+                await delayPromise;
```
AndrewFerr, line 346: As long as the fulfillment value isn't used, would it be simpler to use `Promise.all` / `Promise.allSettled` instead?

(Note that with `allSettled`, an error in `injectJoinFn` wouldn't reduce the await time from taking the whole `floodDelayMs`; and the catch block would have to log `ex[1]` instead of `ex` (as long as there's no reason to believe `delayPromise` would ever reject).)

src/bridge/IrcBridge.ts
Half-Shot: Not *quite*. It's possible that the bridge will find no members to sync yet will have previous clients in the proxy. It's probably trending close to 0% chance, but awaits are cheap and we might as well.

src/bridge/MemberListSyncer.ts
Half-Shot: So, the idea in this section is to await the delayPromise *or* the inject, whichever is faster. We don't want to wait for a join to complete if it takes too long.

If it's too *quick* however, we do want to stagger it. That's the idea behind this section anyway.

- I think using `all` would make us wait for the join, which we don't want to do. We could potentially do a Promise.all([Promise.race, delay]), but that's probably trending close to ugly. I don't think waiting for the delay is that bad?

AndrewFerr: Ah, that makes sense --  I didn't consider the case of the join taking too long.

Then what's the consequence of not awaiting a slow join?  Does it just eventually resolve asynchronously?

src/bridge/IrcBridge.ts
AndrewFerr: Fine to leave it as is -- but to clarify, does `memberListPromises` contain one entry per IRC server (which AIUI is a hard-coded list of which networks to bridge), or one per client connection?  I thought it was the former due to the `push` on line 800 being the only `push` into that list.

src/bridge/IrcBridge.ts
```diff
@@ -774,7 +774,6 @@ export class IrcBridge {
             const syncer = this.memberListSyncers[server.domain] = new MemberListSyncer(
                 this, this.membershipQueue, this.bridge.getBot(), server, this.appServiceUserId,
                 async (roomId: string, joiningUserId: string, displayName: string, isFrontier: boolean) => {
-                    await discoveringClientsPromise;
```
AndrewFerr: Since the remaining `await` on this comes later, is it still true (as per the comment above) that "we will be waiting for discoveringClientsPromise to complete before we execute our first join"?

src/bridge/MemberListSyncer.ts
Half-Shot: Yeah, it'll resolve. If it fails, it doesn't kill the bridge but it'll need to be resolved by the user somehow. We're not going to fix that in this PR, and it's no worse than present day.

src/bridge/IrcBridge.ts
```diff
@@ -802,9 +801,9 @@ export class IrcBridge {
                 // all the clients already connected to avoid races.
                 syncer.sync().then(() =>
                     discoveringClientsPromise
-                ).finally(() => {
+                ).finally(() =>
```
AndrewFerr, line 804: I'll test this myself, but may as well ask now: does awaiting on `promise.finally(callback)` await on the original `promise` (meaning the callback is always resolved asynchronously), or on the Promise returned by the callback (as long as it's non-rejected)?

src/bridge/IrcBridge.ts
Half-Shot: out of band I realised that indeed things are fine, since we wait on `discoveringClientsPromise` after calling .sync(), but before trying any injects. TL;DR it's fine *now*.

src/bridge/IrcBridge.ts
Half-Shot: `.finally` awaits on the promise it's a member of. I.e. in this case `discoveringClientsPromise`

src/bridge/IrcBridge.ts
Half-Shot: Still true but I can highlight it.

AndrewFerr (approved):

PR #1744, opened 2023-06-08
Warn if the bridge is reloaded before it has finished starting

### LINKED ISSUE #1741: Config reload via SIGHUP seems to be broken
```
INFO 21:48:28:210 [cli] Got SIGHUP, reloading config file
WARN 21:48:28:237 [cli] Failed to reload config file: TypeError: Cannot read properties of undefined (reading 'onConfigChanged')
```

Half-Shot:
Fixes #1741

### DIFF
--- a/app.js
+++ b/app.js
@@ -39,6 +39,10 @@ new Cli({
         },
     },
     onConfigChanged: function(config) {
+        if (!bridge) {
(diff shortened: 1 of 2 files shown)


AndrewFerr (approved):
Integration tests are somehow failing on this, but that reflects more on the tests than on this PR :)

PR #1745, opened 2023-06-09
Add trace logs for E2E tests

Half-Shot:
Which should help us diagnose the flakes a bit.

### DIFF
--- a/spec/e2e/setup.ts
+++ b/spec/e2e/setup.ts
@@ -1,4 +1,4 @@
-import fs from "node:fs/promises";
+import * as fs from "node:fs/promises";
 
 export default async function() {
     console.log('Cleaning .e2e-traces')
--- a/spec/e2e/setup.ts
+++ b/spec/e2e/setup.ts
@@ -1,5 +1,8 @@
-import fs from "node:fs";
+import fs from "node:fs/promises";
 
-console.log('Cleaning .e2e-traces')
-fs.rmSync('.e2e-traces', { recursive: true, force: true })
-fs.mkdirSync('.e2e-traces');
+export default async function() {
+    console.log('Cleaning .e2e-traces')
+    await fs.rm('.e2e-traces', { recursive: true, force: true })
+    await fs.mkdir('.e2e-traces');
+}
+
--- a/spec/e2e/setup.ts
+++ b/spec/e2e/setup.ts
@@ -1,4 +1,5 @@
 import fs from "node:fs";
 
+console.log('Cleaning .e2e-traces')
 fs.rmSync('.e2e-traces', { recursive: true, force: true })
 fs.mkdirSync('.e2e-traces');
--- a/spec/e2e/setup.ts
+++ b/spec/e2e/setup.ts
@@ -1,0 +1,4 @@
+import fs from "node:fs";
+
+fs.rmSync('.e2e-traces', { recursive: true, force: true })
+fs.mkdirSync('.e2e-traces');
--- a/spec/util/e2e-test.ts
+++ b/spec/util/e2e-test.ts
@@ -7,7 +7,10 @@ import { IrcServer } from "../../src/irc/IrcServer";
 import { MatrixClient } from "matrix-bot-sdk";
 import { TestIrcServer } from "matrix-org-irc";
 import { IrcConnectionPool } from "../../src/pool-service/IrcConnectionPool";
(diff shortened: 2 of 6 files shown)


spec/util/e2e-test.ts
```diff
@@ -7,7 +7,10 @@ import { IrcServer } from "../../src/irc/IrcServer";
 import { MatrixClient } from "matrix-bot-sdk";
 import { TestIrcServer } from "matrix-org-irc";
 import { IrcConnectionPool } from "../../src/pool-service/IrcConnectionPool";
+import { expect } from "@jest/globals";
 import dns from 'node:dns';
+import fs from "node:fs/promises";
```
AndrewFerr: Looks like `fs` is never used (but `WriteStream` & `createWriteStream` are).

spec/util/e2e-test.ts
```diff
@@ -7,7 +7,10 @@ import { IrcServer } from "../../src/irc/IrcServer";
 import { MatrixClient } from "matrix-bot-sdk";
 import { TestIrcServer } from "matrix-org-irc";
 import { IrcConnectionPool } from "../../src/pool-service/IrcConnectionPool";
+import { expect } from "@jest/globals";
 import dns from 'node:dns';
```
AndrewFerr: Though this isn't a new line, my IDE is giving a warning for it:
> Module '"node:dns"' has no default export.

Using a namespace import should be a viable alternative:
```suggestion
import * as dns from 'node:dns';
```

spec/util/e2e-test.ts
```diff
@@ -289,7 +299,26 @@ export class IrcBridgeE2ETest {
         public readonly registration: AppServiceRegistration,
         readonly postgresDb: string,
         public readonly ircTest: TestIrcServer,
-        public readonly pool?: IrcConnectionPool) {
+        public readonly pool?: IrcConnectionPool,
+        private traceLog?: WriteStream,
+    ) {
+        const startTime = Date.now();
```
AndrewFerr, line 305: Nitpick: this can be moved to inside the `if` block, since it's unused otherwise.

spec/util/e2e-test.ts
```diff
@@ -160,6 +166,10 @@ export class IrcBridgeE2ETest {
     }
 
     static async createTestEnv(opts: Opts = {}): Promise<IrcBridgeE2ETest> {
+        const testName = expect.getState().currentTestName?.replace(/[^a-zA-Z]/g, '-');
+        const tracePath = `${traceFilePath}/${testName}.log`;
+        const traceStream = createWriteStream(tracePath, 'utf-8');
```
AndrewFerr: Consider renaming this to `traceLog` for consistency with the `IrcBridgeE2ETest.traceLog`.

AndrewFerr (changes requested):
This is a good idea!  Just some nitpick requests from me & we can go ahead with this.

spec/util/e2e-test.ts
Half-Shot: traceLog is no longer optional, so I removed the `if`.

spec/e2e/setup.ts
```diff
@@ -0,0 +1,8 @@
+import fs from "node:fs/promises";
```
AndrewFerr: Again, not sure if it matters (or if it's a config error on my end), but this is giving me a "no default export" warning, so use this if appropriate:
```suggestion
import * as fs from "node:fs/promises";
```
This follows the ESM-style imports shown in [Node docs](https://nodejs.org/api/module.html).

Though it must not matter after all, since the tests passed in CI, so feel free to ignore this.

spec/e2e/setup.ts
```diff
@@ -0,0 +1,8 @@
...
+
+export default async function() {
```
AndrewFerr, line 3: To clarify, this runs _before_ each group of tests to ensure that the trace directory is empty before writing files into it, right?

AndrewFerr (approved):

spec/e2e/setup.ts
Half-Shot: Yeah, this apparently works. Weird.

spec/e2e/setup.ts
Half-Shot: Definitely, I found out my earlier attempts did not, much to my frustration. This only runs on Jest startup!

PR #1751, opened 2023-07-24
Performance improvements for pooling stream reader

Half-Shot:
This change should ideally improve performance and do less redundant work for bridges in the IRC pool configuration.


tadzik (approved):
No complaints.

PR #1752, opened 2023-07-25
Ensure QUIT messages are sent in pooling configuration

Half-Shot:
Failing to specify a readyState means we missed this one.

### DIFF
--- a/spec/e2e/quit.spec.ts
+++ b/spec/e2e/quit.spec.ts
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/no-non-null-assertion */
 import { TestIrcServer } from "matrix-org-irc";
 import { IrcBridgeE2ETest } from "../util/e2e-test";
 import { describe, it } from "@jest/globals";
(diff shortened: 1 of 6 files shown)


spec/e2e/quit.spec.ts
```diff
@@ -0,0 +1,43 @@
...
+            matrixLocalparts: ['alice'],
+            ircNicks: ['bob'],
+            traceToFile: true,
+        });
+        await testEnv.setUp();
+    });
+    afterEach(() => {
+        return testEnv?.tearDown();
+    });
+    it('should correctly send a quit message on ', async () => {
```
tadzik:
```suggestion
    it('should correctly send a quit message on !reconnect', async () => {
```

src/pool-service/RedisIrcConnection.ts
```diff
@@ -11,12 +11,18 @@ export type RedisIrcConnectionEvents = {
...
 export class RedisIrcConnection extends (EventEmitter as unknown as
     new () => TypedEmitter<RedisIrcConnectionEvents&IrcConnectionEventsMap>) implements IrcConnection {
-    private readonly log = new Logger(`RedisIrcConnection:${this.clientId}`);
+    private readonly log: Logger;
 
     public get connecting() {
         return this.isConnecting;
     }
 
+    public get readyState() {
```
tadzik: It'd be nice to have this typed, but seems like it's not typed in node-irc either :/ Next time.

tadzik (approved):
Just nitpicks.

src/pool-service/RedisIrcConnection.ts
Half-Shot: I can type it here at least.

PR #1753, opened 2023-07-25
Add a test to ensure that the irc client state doesn't bloat on VERSION

Half-Shot:
Requires https://github.com/matrix-org/node-irc/pull/108

### DIFF
--- a/spec/e2e/pooling.spec.ts
+++ b/spec/e2e/pooling.spec.ts
@@ -72,4 +72,31 @@ describeif('Connection pooling', () => {
         bob.say(channel, "Hi alice!");
         await bobMsg;
     });
+
+    it.only('should store the IRC client state once', async () => {
+        const channel = `#${TestIrcServer.generateUniqueNick("test")}`;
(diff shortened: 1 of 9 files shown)


spec/e2e/pooling.spec.ts
```diff
@@ -72,4 +72,31 @@ describeif('Connection pooling', () => {
         bob.say(channel, "Hi alice!");
         await bobMsg;
     });
+
+    it.only('should store the IRC client state once', async () => {
```
tadzik:
```suggestion
    it('should store the IRC client state once', async () => {
```

tadzik: (There's some eslint rule for that)

tadzik (changes requested):

spec/e2e/pooling.spec.ts
Half-Shot: aha, `eslint-plugin-jest`

tadzik (approved):

PR #1755, opened 2023-07-26
Sync bans from legacy Mjolnir-y rules

Half-Shot:
Mjolnir used to use these old event types for moderation, but we implemented the Matrix-spec and MSC variants. This total miss meant we didn't catch older rules.

### DIFF
--- a/src/irc/ClientPool.ts
+++ b/src/irc/ClientPool.ts
@@ -137,16 +137,25 @@ export class ClientPool {
             return;
         }
 
+        // XXX: This is a safe assumption *for now* but when the proxy supports multiple
+        // servers this will break!
+        const server = this.ircBridge.getServers()[0];
         for await (const connection of this.redisPool.getPreviouslyConnectedClients()) {
-            if (connection.clientId === 'bot') {
+            if (connection.clientId) {
                 // The bot will be connected via the usual process.
                 continue;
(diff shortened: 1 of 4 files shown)


src/irc/ClientPool.ts
```diff
@@ -137,16 +137,26 @@ export class ClientPool {
...
+        const server = this.ircBridge.getServers()[0];
         for await (const connection of this.redisPool.getPreviouslyConnectedClients()) {
             if (connection.clientId === 'bot') {
                 // The bot will be connected via the usual process.
                 continue;
             }
-            // XXX: This is a safe assumption *for now* but when the proxy supports multiple
-            // servers this will break!
-            const server = this.ircBridge.getServers()[0];
+            const mxUser = new MatrixUser(connection.clientId) ;
```
Half-Shot, line 148: What is this all doing, I hear you cry?

We're setting up IRC clients for existing users who are still connected before a bridge restart. We don't use the getBridgedClient function because it does things we don't need to do for existing connections. We're also **not** checking for excluded or blocked users here, because we **always** want to reconnect a user on startup without fail.

Once we've reconnected everyone, we may end up kicking a few people off the network if they are on the ban list.

AndrewFerr (approved):

PR #1756, opened 2023-07-27
Refactor the command parser

Half-Shot:
We wrote a command parser twice, and it's about time we reconciled it a bit.

### DIFF
--- a/src/bridge/AdminRoomHandler.ts
+++ b/src/bridge/AdminRoomHandler.ts
@@ -48,7 +48,7 @@ export function parseCommandFromEvent(event: { content?: { body?: unknown }}, pr
         return null;
     }
     // First line only
-    const body = event.content.body.trim().split("\n")[0];
+    const body = event.content.body.trim().split(/\n|\000/)[0];
     if (!body.startsWith(prefix)) {
         return null;
     }
--- a/src/bridge/AdminRoomHandler.ts
+++ b/src/bridge/AdminRoomHandler.ts
@@ -54,6 +54,10 @@ export function parseCommandFromEvent(event: { content?: { body?: unknown }}, pr
     }
     // Assumes all commands have the form "!wxyz [irc.server] [args...]"
     const [cmd, ...args] = body.slice(prefix.length).split(" ");
+    if (cmd === '') {
+        // Command must have length.
+        return null;
+    }
     return { cmd, args };
 }
 
--- a/src/bridge/AdminRoomHandler.ts
+++ b/src/bridge/AdminRoomHandler.ts
@@ -108,7 +108,7 @@ const COMMANDS: {[command: string]: Command|Heading} = {
                 "connections to all networks.",
     },
     "active": {
-        example: "active",
+        example: "!active",
         summary: "Mark yourself as active, which will exclude you from any idleness kicks."
     },
     'Authentication': { heading: true },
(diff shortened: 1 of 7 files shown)


src/bridge/AdminRoomHandler.ts
```diff
@@ -127,7 +149,7 @@ const COMMANDS: {[command: string]: Command|Heading} = {
         requiresPermission: CommandPermission.Admin,
     },
     '!unlink': {
-        example: "!unlink !room:example.com irc.example.net #foobar",
+        example: "unlink !room:example.com irc.example.net #foobar",
```
progval: looks like you forgot to remove the `!` prefix from `!plumb` and `!unlink`, and removed it from the example instead

Half-Shot: ta, this'll be because we decided to use different forms of quote.

src/bridge/AdminRoomHandler.ts
```diff
@@ -35,6 +36,27 @@ enum CommandPermission {
...
+    if (typeof event.content?.body !== "string") {
+        return null;
+    }
+    // First line only
+    const body = event.content.body.trim().split("\n")[0];
+    if (!body.startsWith(prefix)) {
+        return null;
+    }
+    // Assumes all commands have the form "!wxyz [irc.server] [args...]"
+    const [cmd, ...args] = body.slice(prefix.length).split(" ");
```
AndrewFerr, line 56: Should this check if `cmd` ends up being empty?  i.e. should `null` be returned for an `event.content.body` of `"!"`?

src/provisioning/Provisioner.ts
```diff
@@ -860,9 +870,15 @@ export class Provisioner extends ProvisioningApi {
 
         const ircDomain = body.remote_room_server;
         const ircChannel = body.remote_room_channel;
+
+
```
AndrewFerr: Nitpick, is one newline sufficient? :slightly_smiling_face:

src/bridge/AdminRoomHandler.ts
Half-Shot: Yes! Good point.

src/provisioning/Provisioner.ts
Half-Shot: It looks like I do more work! But yep, removed.

AndrewFerr (approved):

src/models/IrcRoom.ts
```diff
@@ -18,6 +18,17 @@ import { RemoteRoom } from "matrix-appservice-bridge";
...
 
+/**
+ * Validate that a channel name is in the correct form.
+ * NOTE: This accepts both # and & as channel prefixes.
+ * @param channel Channel name
+ * @returns True if the channel is valid, or false otherwise.
+ */
+export function validateChannelName(channel: string) {
+    // According to https://www.rfc-editor.org/rfc/rfc1459#section-1.3
+    return /^(#|&)([^\s\cG,]){1,199}$/.test(channel);
```
progval: I don't think AS-irc supports `&`-channels, so you can probably disallow those.

(also non-space whitespaces are technically allowed in practice, eg. on OFTC, but that's just asking for trouble so it's fine to disallow them too)

Half-Shot: I left them in on a RFC technicality, but tbh it's probably not useful. :+1: 

and uuuurgghhhh whitespaces are something we'll care about later if people are desperate.

Half-Shot: `\cG matches the control sequence CTRL+G (ASCII 7)` according to my notes.

PR #1757, opened 2023-07-28
Support authenticating via certfp

### LINKED ISSUE #1483: !storepass doesn't work at all on OFTC
**Describe the bug**

It's impossible to store a nickserv password or certificate on OFTC.

OFTC doesn't support using server password to identify to services, which means currently there's no way to automatically reidentify after connection blips between the oftc bridge and their network. OFTC does support cert auth.

**To Reproduce**
1. Connect to OFTC. Register your nickname with NickServ.
2. Tell @oftc-irc:​matrix.org to `!storepass foo`
3. `!reconnect`

**Expected behaviour**

Nickname gets authenticated on NickServ (through certfp) on reconnection.

**Actual behaviour**

Nickname doesn't get authenticated, making it impossible to rejoin some channels. No error messages.

**Additional context**

There was #459 which got closed even though the problem still exists.

#1128 to add certfp was abandoned. #1331 which was done instead doesn't fix the issue - authentication fails even with `!username`.

#747 tracks certfp support, but doesn't make clear that authentication is currently completely broken on OFTC.

Half-Shot:
Fixes https://github.com/matrix-org/matrix-appservice-irc/issues/1483, https://github.com/matrix-org/matrix-appservice-irc/issues/747

Requires https://github.com/matrix-org/node-irc/pull/110

Most of the work was already done in node-irc, so this just needs the bridge to store and pass through the certificate. I haven't yet implemented a way for users to pass the cert to the bridge. I am wary of passing it through the admin room directly as it would leave the cert in Matrix.

Perhaps we may have a HTTP submission page of some kind.

### DIFF
--- a/src/bridge/AdminRoomHandler.ts
+++ b/src/bridge/AdminRoomHandler.ts
@@ -200,7 +200,7 @@ class ServerRequiredError extends Error {
 const USER_FEATURES = ["mentions"];
 export class AdminRoomHandler {
     private readonly botUser: MatrixUser;
-    private readonly roomIdsExpectingCertFp = new Map<string, (certificate: string) => void>();
+    private readonly roomIdsExpectingCertFp: Map<string, (certificate: string) => void> = new Map();
     constructor(private ircBridge: IrcBridge, private matrixHandler: MatrixHandler) {
         this.botUser = new MatrixUser(ircBridge.appServiceUserId, undefined, false);
     }
@@ -316,7 +316,8 @@ export class AdminRoomHandler {
         req.log.info(`${sender} is attempting to store a cert for ${server.domain}`);
         await this.ircBridge.sendMatrixAction(
             adminRoom, this.botUser, new MatrixAction(
-                ActionType.Notice, `Please enter your certificate and private key (without formatting) for ${server.getReadableName()}.`
+                ActionType.Notice,
+                `Please enter your certificate and private key (without formatting) for ${server.getReadableName()}.`
             )
         );
         let certfp: string;
@@ -676,7 +677,6 @@ export class AdminRoomHandler {
         const server = this.extractServerFromArgs(args);
 
         const domain = server.domain;
-        const store = this.ircBridge.getStore();
         let notice;
 
         try {
--- a/src/bridge/AdminRoomHandler.ts
+++ b/src/bridge/AdminRoomHandler.ts
@@ -28,6 +28,7 @@ import { getBridgeVersion } from "matrix-appservice-bridge";
 import { Provisioner } from "../provisioning/Provisioner";
 import { IrcProvisioningError } from "../provisioning/Schema";
 import { validateChannelName } from "../models/IrcRoom";
+import { X509Certificate, createPrivateKey } from "node:crypto";
 
 const log = logging("AdminRoomHandler");
 
@@ -61,6 +62,33 @@ export function parseCommandFromEvent(event: { content?: { body?: unknown }}, pr
     return { cmd, args };
 }
 
+
+
+export function getKeyPairFromString(keypairString: string) {
+    const keyStart = keypairString.indexOf('-----BEGIN PRIVATE KEY-----');
+    const keyEnd = keypairString.indexOf('-----END PRIVATE KEY-----');
+    const certStart = keypairString.indexOf('-----BEGIN CERTIFICATE-----');
+    const certEnd = keypairString.indexOf('-----END CERTIFICATE-----');
+    if (certStart === -1) {
+        throw Error('Missing BEGIN CERTIFICATE');
+    }
+    if (certEnd === -1) {
+        throw Error('Missing END CERTIFICATE');
+    }
+    if (keyStart === -1) {
+        throw Error('Missing BEGIN PRIVATE KEY');
+    }
+    if (keyEnd === -1) {
+        throw Error('Missing END PRIVATE KEY');
+    }
+    const certStr = keypairString.slice(certStart, certEnd + '-----END CERTIFICATE-----'.length);
+    const privateKeyStr = keypairString.slice(keyStart, keyEnd + '-----END CERTIFICATE-----'.length);
+    const privateKey = createPrivateKey(privateKeyStr);
+    const cert = new X509Certificate(certStr);
+    return { cert, privateKey };
+
+}
+
 // This is just a length to avoid silly long usernames
 const SANE_USERNAME_LENGTH = 64;
 
@@ -75,6 +103,8 @@ const SANE_USERNAME_LENGTH = 64;
 // (0x00 to 0x1F, plus DEL (0x7F)), as they are most likely mistakes.
 const SASL_USERNAME_INVALID_CHARS_PATTERN = /[\x00-\x20\x7F]+/; // eslint-disable-line
 
+const CERT_FP_TIMEOUT_MS = 90 * 1000;
+
 interface Command {
     example: string;
     summary: string;
@@ -87,6 +117,10 @@ interface Heading {
 
 const COMMANDS: {[command: string]: Command|Heading} = {
     'Actions': { heading: true },
+    "certfp": {
+        example: `!certfp [irc.example.net]`,
+        summary: `Store a certificate for authenticating with the remote network`,
+    },
     "cmd": {
         example: `!cmd [irc.example.net] COMMAND [arg0 [arg1 [...]]]`,
         summary: "Issue a raw IRC command. These will not produce a reply." +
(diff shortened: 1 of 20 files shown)


src/datastore/postgres/PgDataStore.ts
```diff
@@ -524,6 +524,23 @@ export class PgDataStore implements DataStore, ProvisioningStore {
...
+        config.certificate = {
+            cert: row.cert,
+            key: row.key,
+        };
+        // TODO: Testing
+        if (row.cert && row.key && this.cryptoStore) {
+            // NOT fatal, but really worrying.
+            try {
+                config.certificate = {
+                    cert: this.cryptoStore.decrypt(row.cert),
```
progval: why encrypt the certificate? it's supposed to be public

src/datastore/postgres/PgDataStore.ts
```diff
@@ -524,6 +524,23 @@ export class PgDataStore implements DataStore, ProvisioningStore {
...
+                    key:  this.cryptoStore.decrypt(row.key)
+                };
+            }
+            catch (ex) {
+                log.warn(`Failed to decrypt certificate for ${userId} ${domain}`, ex);
```
progval:
```suggestion
                log.warn(`Failed to decrypt TLS key for ${userId} ${domain}`, ex);
```

src/datastore/postgres/PgDataStore.ts
```diff
@@ -536,14 +553,23 @@ export class PgDataStore implements DataStore, ProvisioningStore {
         // We need to make sure we have a matrix user in the store.
         await this.pgPool.query("INSERT INTO matrix_users VALUES ($1, NULL) ON CONFLICT DO NOTHING", [userId]);
         let password = config.getPassword();
+        const cert: {cert?: string, key?: string} = { };
```
progval: to be pedantic it's more of a keypair (plus de certificate) than just a cert

src/datastore/postgres/PgDataStore.ts
Half-Shot: fair point.

justinbot (approved):

src/bridge/AdminRoomHandler.ts
```diff
@@ -252,6 +299,107 @@ export class AdminRoomHandler {
...
+                ActionType.Notice, 'Could not parse validFrom / validTo dates.',
+            );
+        }
+
+        const clientConfig = await this.getOrCreateClientConfig(sender, server);
+        clientConfig.setCertificate({
+            cert: cert.toString(),
+            key: privateKey.export({type: 'pkcs8', format: 'pem'}).toString(),
+        });
+        await this.ircBridge.getStore().storeIrcClientConfig(clientConfig);
```
justinbot: Needs error handling?

src/datastore/postgres/PgDataStore.ts
```diff
@@ -565,16 +564,7 @@ export class PgDataStore implements DataStore, ProvisioningStore {
...
-            let key = config.certificate.key;
-            // We can't store these as our encryption system doesn't support spaces.
-            key = key.replace('-----BEGIN PRIVATE KEY-----\n', '').replace('-----END PRIVATE KEY-----\n', '');
-            while (key.length > 0) {
-                const part = key.slice(0, 64);
-                cryptoParts.push(this.cryptoStore.encrypt(part));
-                key = key.slice(64);
-            }
-            keypair.key = cryptoParts.join(',');
+            keypair.key = this.cryptoStore.encryptLargeString(config.certificate.key);
```
justinbot: I thought we couldn't encrypt spaces?

Half-Shot: I found a way 👾

src/datastore/StringCrypto.ts
```diff
@@ -14,11 +14,16 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-import * as crypto from "crypto";
+import { createCipheriv, createDecipheriv, privateDecrypt, publicEncrypt, scrypt as scryptCb } from "node:crypto";
 import * as fs from "fs";
 import { getLogger } from "../logging";
+import { randomBytes } from "node:crypto";
+import { promisify } from "node:util";
```
AndrewFerr: Nitpick
```suggestion
import { createCipheriv, createDecipheriv, privateDecrypt, publicEncrypt, randomBytes, scrypt as scryptCb } from "node:crypto";
import { promisify } from "node:util";
import * as fs from "fs";
import { getLogger } from "../logging";
```

src/datastore/StringCrypto.ts
```diff
@@ -48,19 +53,59 @@ export class StringCrypto {
     }
 
     public encrypt(plaintext: string): string {
-        const salt = crypto.randomBytes(16).toString('base64');
-        return crypto.publicEncrypt(
+        const salt = randomBytes(16).toString('base64');
```
AndrewFerr: It would help readability / maintainability to define variables for the salt length & related values.

Example:

```ts
const SALT_ENCODING = 'base64';
const SALT_BYTE_LENGTH = 16;
const SALT_STRING_LENGTH = Buffer.alloc(SALT_BYTE_LENGTH).toString(SALT_ENCODING).length;

const ENCRYPTED_ENCODING = 'base64';

export class StringCrypto {
    // snip
    public encrypt(plaintext: string): string {
        const salt = randomBytes(SALT_BYTE_LENGTH).toString(SALT_ENCODING);
        return publicEncrypt(
            this.privateKey,
            Buffer.from(salt + ' ' + plaintext)
        ).toString(ENCRYPTED_ENCODING);
    }
    public decrypt(encryptedString: string): string {
        const decryptedPass = privateDecrypt(
            this.privateKey,
            Buffer.from(encryptedString, ENCRYPTED_ENCODING)
        ).toString();
        // Extract the password by removing the prefixed salt and seperating space
        return decryptedPass.slice(SALT_STRING_LENGTH+1);
    }
    // snip
}
```

src/datastore/StringCrypto.ts
```diff
@@ -29,7 +34,7 @@ export class StringCrypto {
 
             // Test whether key is a valid PEM key (publicEncrypt does internal validation)
```
AndrewFerr, line 47: This probably warrants another PR, but should symmetric encryption be used instead of RSA?  Since encryption & decryption are done by the same entity (i.e. the bridge), there doesn't seem to be an advantage to using asymmetric encryption here.

Granted, changing to symmetric encryption would require migrating any existing secrets stored in the database & making `ircService.passwordEncryptionKeyPath` expect some kind of password/secret other than an RSA privkey, so maybe it's not worth the friction at this point.

Half-Shot: Yeah, I agree. Unfortunately I think that would be a (necessary) breaking change for the moment.

src/datastore/StringCrypto.ts
```diff
@@ -48,19 +66,66 @@ export class StringCrypto {
...
+            Buffer.from(encryptedString, ENCRYPTED_ENCODING)
         ).toString();
         // Extract the password by removing the prefixed salt and seperating space
-        return decryptedPass.split(' ')[1];
+        return decryptedPass.slice(SALT_STRING_LENGTH + 1);
+    }
+
+    public async encryptLargeString(plaintext: string): Promise<string> {
+        const password = randomBytes(32).toString(ENCRYPTED_ENCODING);
+        const key = await scrypt(password, 'salt', 32) as Buffer;
```
AndrewFerr, line 87: As long as you're generating a random password, there's probably not much of a point in passing to scrypt, as it can be used directly as an AES key.

src/datastore/StringCrypto.ts
```diff
@@ -48,19 +66,66 @@ export class StringCrypto {
...
+            cipher.on('error', (err) => reject(err));
+            cipher.on('end', () => resolve(
+                `lg:${secret}:${encrypted}`
+            ));
+        });
+
+        cipher.on('data', (chunk) => { encrypted += chunk });
+        cipher.write(plaintext);
+        cipher.end();
+        return streamPromise;
```
AndrewFerr, line 107: Since the data to be written into the cipher is known to be just the provided `plaintext`, it'd be more direct to use `cipher.update` and `cipher.final` instead of bothering with stream callbacks.

src/datastore/StringCrypto.ts
```diff
@@ -48,19 +66,66 @@ export class StringCrypto {
...
+    }
+
+    public async decryptLargeString(encryptedString: string): Promise<string> {
+        if (!encryptedString.startsWith('lg:')) {
+            throw Error('Not a large string');
+        }
+        const [, keyPlusIvEnc, data] = encryptedString.split(':', 3);
+        const [keyB64, ivB64] = this.decrypt(keyPlusIvEnc).split('_');
```
AndrewFerr, line 115: Is it guaranteed that the ciphertexts of `encrypt()` or the AES cipher won't contain underscores or colons that would interfere with these splits?

AndrewFerr (changes requested):

programmerjake:
are you still intending on working on this? having storepass work on oftc would be really nice!

progval:
This is an alternative to `!storepass`, it won't make `!storepass` itself work.

funderscore1:
Any update?

PR #1758, opened 2023-07-30
move dependencies to devDependencies

gdamjan:
* typescript was already in devDependencies
* dependencies used only by the frontend widget are only devDependencies, and are not required by the backend (at runtime):
  - react
  - react-dom
  - matrix-widget-api
  - url-join
  - classnames
* cross-fetch is only used in tests
* 
Signed-off-by: Damjan Georgievski <gdamjan@gmail.com>


Half-Shot (approved):
Nice catches!

I think this will be good to go as soon as reviews pass.

gdamjan:
is, the sign-off check having issues with my cyrillic name!? :D

Half-Shot:
> is, the sign-off check having issues with my cyrillic name!? :D

The regex *should* be reasonably sensible these days...ugh. https://github.com/matrix-org/backend-meta/blob/release/v2/.github/workflows/sign-off.yml#L27

PR #1759, opened 2023-07-31
Don't await connection when starting up in pooling mode

Half-Shot:
Introduced in #1755, hence the same changelog text to merge them.

### DIFF
--- a/src/irc/ClientPool.ts
+++ b/src/irc/ClientPool.ts
@@ -136,6 +136,7 @@ export class ClientPool {
         if (!this.redisPool) {
             return;
         }
+        log.info(`Discovering connected pool clients`);
 
         // XXX: This is a safe assumption *for now* but when the proxy supports multiple
         // servers this will break!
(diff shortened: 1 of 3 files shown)


src/irc/ClientPool.ts
```diff
@@ -156,8 +158,11 @@ export class ClientPool {
                         mxUser, server.domain
                     );
                 const bridgeClient = await this.createIrcClient(config, mxUser, false, false);
-                await bridgeClient.connect();
-                log.info(`Connected previously connected user ${connection.clientId}`);
+                bridgeClient.connect().then(() => {
+                    log.info(`Connected previously connected user ${connection.clientId}`);
+                }).catch((ex) => {
+                    log.warning(`Connected previously connected user ${connection.clientId}`, ex);
```
tadzik: Shouldn't this say "Not connected" or something like that?

Half-Shot: Yes, ideally!

PR #1763, opened 2023-08-01
Fix Redis <=6.2 throwing errors when trying to clear the command queue in pooling mode.

### LINKED ISSUE #1761: Failed to trim commands from the IN/OUT stream ReplyError: ERR syntax error
```
WARN 09:58:51:848 [IrcConnectionPool] Failed to trim commands from the IN stream ReplyError: ERR syntax error
...
     args: [ 'ircbridge.stream.command.in', 'MINID', '1690797483729-0' ]
```

Looks like we're failing to specify the min ID without the suffix, upsetting the TRIM syntax.

Half-Shot:
Fixes #1761 

Ultimately the problem here is that we failed to specify a minimum Redis version. If you are using less than 6.2, you won't be able to clear the command stream effectively. To get around this for now, we're reverting to previous behaviour of just setting an absolute maximum message count for the stream when running on older Redis versions.

### DIFF
--- a/src/pool-service/CommandReader.ts
+++ b/src/pool-service/CommandReader.ts
@@ -1,5 +1,4 @@
 import { Redis } from "ioredis";
-import { READ_BUFFER_MAGIC_BYTES } from "./types";
 import semver from "semver";
 import { Logger } from "matrix-appservice-bridge";
 
@@ -9,7 +8,7 @@ const TRIM_MAXLEN_COUNT = 100_000;
 
 const log = new Logger('RedisCommandReader');
 
-export class RedisCommandReader<CommandType, PayloadType> {
+export class RedisCommandReader {
 
     private shouldRun = true;
     private commandStreamId = "$"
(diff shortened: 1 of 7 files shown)


src/pool-service/CommandReader.ts
```diff
@@ -0,0 +1,120 @@
...
+                    .catch(ex => log.warn(`Failed to handle msg ${msgId} (${cmdType}, ${payload})`, ex)
+                    ),
+            );
+        }
+
+    }
+
+    public async getSupported() {
+        const serverLines = (await
+        this.redis.info("Server")).split('\n').filter(v => !v.startsWith('#')).map(v => v.split(':', 2)
```
tadzik: Ouch. Awkward that there's no easier way to do this.

src/pool-service/CommandReader.ts
```diff
@@ -0,0 +1,120 @@
...
+        ) as [string, string][];
+        const options = new Map(serverLines);
+        const version = options.get('redis_version');
+        if (!version) {
+            log.warn(`Unable to identify Redis version, assuming unsupported version`);
```
tadzik: Perhaps we should also do this if parsing the server info failed for any reason.

tadzik (changes requested):
Some concerns. I like the refactoring!

src/pool-service/CommandReader.ts
Half-Shot: I was quite annoyed the library didn't parse this for us, but oh well.

tadzik (approved):

PR #1764, opened 2023-08-01
Test mass joining / syncing in E2E

Half-Shot:
This tests a reasonably sized amount of joins both after startup, and also the startup syncing itself from M->I.


Half-Shot:
Hey there people on the tail end of 2025, closing this PR as we've in practice not felt the need to work on it over the last few years and it's now rotted to the point where it would be a lot of work to fix. Anyone may pick this back up if they wish to, however!

PR #1767, opened 2023-08-07
docker: don't push on a Pull Request

gdamjan:
Since the github action triggers on a PR too, `github.ref` will be the target branch for merging (typically `develop`). In that case, don't try to login/push to docker hub.

https://docs.github.com/en/actions/learn-github-actions/contexts#github-context

Signed-off-by: Damjan Georgievski <gdamjan@gmail.com>

### DIFF
--- a/.github/workflows/docker-hub-latest.yml
+++ b/.github/workflows/docker-hub-latest.yml
@@ -11,7 +11,8 @@ env:
   DOCKER_NAMESPACE: matrixdotorg
   PLATFORMS: linux/amd64
   # Only push if this is develop, otherwise we just want to build
-  PUSH: ${{ github.ref == 'refs/heads/develop' }}
+  # On a PR github.ref is the target branch, so don't push for that either
+  PUSH: ${{ github.ref == 'refs/heads/develop' && github.event_name != 'pull_request' }}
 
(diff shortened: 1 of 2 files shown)


Half-Shot:
Looks good, happy to merge.

Half-Shot (approved):

.github/workflows/docker-hub-latest.yml
```diff
@@ -11,7 +11,8 @@ env:
   DOCKER_NAMESPACE: matrixdotorg
   PLATFORMS: linux/amd64
   # Only push if this is develop, otherwise we just want to build
-  PUSH: ${{ github.ref == 'refs/heads/develop' }}
+  # On a PR github.ref is the target branch, so don't push for that either
+  PUSH: ${{ github.ref == 'refs/heads/develop' && github.event_name != 'pull_request' }}
```
Half-Shot, line 15: Hm, the jobs might need a `env.PUSH == 'true'` perhaps?

gdamjan: ah, indeed, since PUSH is an env var.

Half-Shot:
Thanks again :)

PR #1772, opened 2023-08-14
Drop support for legacy groups

Half-Shot:
A feature from long ago when Synapse supported the concept of groups. This has been replaced by Spaces. Spaces hasn't yet added support for flairs or suggested groups, so there isn't a direct replacement.

This feature does nothing though and can be removed.


tadzik (approved):
[:(](https://www.youtube.com/watch?v=X2WH8mHJnhM)

PR #1781, opened 2023-10-01
Spec-compliant localparts for virtual matrix users

### LINKED ISSUE #1780: localparts for virtual Matrix users contain capital letters
Matrix User Identifiers are supposed to only contain lower-case letters

https://spec.matrix.org/v1.8/appendices/#user-identifiers

However, the bridge creates Localparts with capital letters, if the IRC nick contains capital letters

https://github.com/matrix-org/matrix-appservice-irc/blob/develop/src/irc/IrcServer.ts#L535

Synapse seems to not care and just accept these User IDs. 

Conduit, however only expects lower-case IDs and then bridging for IRC users with capital letters in their names fails.

see also https://gitlab.com/famedly/conduit/-/issues/388

```diff
diff --git a/src/irc/IrcServer.ts b/src/irc/IrcServer.ts
index 7bcbef57..c9cce65e 100644
--- a/src/irc/IrcServer.ts
+++ b/src/irc/IrcServer.ts
@@ -548,7 +548,7 @@ export class IrcServer {
         return renderTemplate(this.config.matrixClients.userTemplate, {
             server: this.domain,
             nick,
-        }).substring(1); // the first character is guaranteed by config schema to be '@'
+        }).substring(1).toLowerCase(); // the first character is guaranteed by config schema to be '@'
     }
 
     public claimsUserId(userId: string): boolean {
@@ -586,9 +586,7 @@ export class IrcServer {
     }
 
     public getUserIdFromNick(nick: string): string {
-        const template = this.config.matrixClients.userTemplate;
-        return template.replace(/\$NICK/g, nick).replace(/\$SERVER/g, this.domain) +
-            ":" + this.homeserverDomain;
+        return "@" + this.getUserLocalpart(nick) + ":" + this.homeserverDomain;
     }
 
     public getDisplayNameFromNick(nick: string): string {

```

Lowercasing all Localparts seems to work for me but I don't know about the consequences for other homeservers or existing deployments. Maybe this should be a config option

kbleeke:
Closes #1780 

I would like to have the option to create localparts for virtual matrix users that only contain valid characters according to 

https://spec.matrix.org/v1.8/appendices/#user-identifiers

The spec then suggest to map characters like

https://spec.matrix.org/v1.8/appendices/#mapping-from-other-character-sets

Well, the "have the option" part is still missing from the config file.

If you are generally interested in accepting a PR for this, I will do my best to comply with [Contributing.md](https://github.com/matrix-org/matrix-appservice-bridge/blob/develop/CONTRIBUTING.md#%EF%B8%8F-pull-requests)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -10,6 +10,9 @@ RUN cd freebindfree && make
 # Typescript build
 FROM node:18 as builder
 
+RUN apt-get update && apt-get install -y node-gyp --no-install-recommends
+RUN rm -rf /var/lib/apt/lists/*
+
 WORKDIR /build
 
 COPY src/ /build/src/
--- a/config.schema.yml
+++ b/config.schema.yml
@@ -413,6 +413,10 @@ properties:
                         matrixClients:
                             type: "object"
                             properties:
+                                localPartCharacterMapping:
+                                    type: "integer"
+                                    minimum: 0
+                                    maximum: 1
                                 userTemplate:
                                     type: "string"
                                     pattern: "^@.*\\$NICK"
--- a/src/irc/IrcServer.ts
+++ b/src/irc/IrcServer.ts
@@ -535,9 +535,15 @@ export class IrcServer {
     public getUserLocalpart(nick: string): string {
         // the template is just a literal string with special vars; so find/replace
         // the vars and strip the @
+
+        // https://spec.matrix.org/v1.8/appendices/#mapping-from-other-character-sets
+        const escaped = nick.replaceAll(/[A-Z_]/g, (c) => "_" + c.toLowerCase());
+        const escaped2 = escaped.replaceAll(/[^a-z0-9\.\_\-\/+]/g,
+            (c) => "=" + c.charCodeAt(0).toString(16).padStart(2, '0'));
+
         return renderTemplate(this.config.matrixClients.userTemplate, {
             server: this.domain,
-            nick,
+            nick: escaped2,
         }).substring(1); // the first character is guaranteed by config schema to be '@'
     }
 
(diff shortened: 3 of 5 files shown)


Half-Shot:
Hey there, I'd absolutely love to have this change included. It has been a long time coming. Obviously we'd need to be careful to support "legacy" identifiers as the old behaviour is widespread but I'd love to see this happen for new deployments.

Dockerfile
```diff
@@ -10,6 +10,9 @@ RUN cd freebindfree && make
 # Typescript build
 FROM node:18 as builder
 
+RUN apt-get update && apt-get install -y node-gyp --no-install-recommends
```
Half-Shot, line 13: Do you know what caused this to flag up? It's weird that you found this without changing the deps around at all.

src/irc/IrcServer.ts
```diff
@@ -582,13 +588,18 @@ export class IrcServer {
         if (!match) {
             return null;
         }
-        return match[1];
+
+        // https://spec.matrix.org/v1.8/appendices/#mapping-from-other-character-sets
+        const unescaped = match[1].replaceAll(/=([0-9a-f][0-9a-f])/g,
```
Half-Shot: I'd suggest making this behaviour configurable for now so old deployments don't break (and defaulting to off). In a future PR we can start defaulting new deployments to using it, and somewhere down the line we should stop supporting the old mechansim.

Make sense?

Half-Shot (commented):
Early thoughts

src/irc/IrcServer.ts
kbleeke: Yes, makes perfect sense.

 I haven't looked at any configuration code yet. I will update this with a proposal for a configuration option

Dockerfile
kbleeke: No idea, it just failed to compile. I wrote this on a brand new VPS. The base images were all fresh downloads. Maybe the upstream image somehow removed node-gyp?

src/irc/IrcServer.ts
```diff
@@ -80,6 +80,7 @@ export interface IrcServerConfig {
         federate: boolean;
     };
     matrixClients: {
+        matrixIdNormalization: number,
```
kbleeke: I thought that I should make this an integer, in case there are more schemes in the future

Dockerfile
kbleeke: It seems node:18.17 works node:18.18 is missing node-gyp for some reason.

config.schema.yml
```diff
@@ -413,6 +413,10 @@ properties:
                         matrixClients:
                             type: "object"
                             properties:
+                                localPartCharacterMapping:
+                                    type: "integer"
+                                    minimum: 0
+                                    maximum: 1
```
kbleeke, line 419: Is there a way to set a default value here so it doesn't break existing configs? Or does that happens automatically?

Dockerfile
Half-Shot: We've had this before where minor image versions break (and then because we target a major version, you end up not noticing).

The change looks good then. I think in the future we're just going to have to start pinning versions strictly as we can't be having this.

PR #1782, opened 2023-10-18
Ignore functional members in admin rooms

Half-Shot:
Prevents us from blocking admin rooms when a specific kind of user joins. Based off https://github.com/vector-im/element-meta/blob/develop/spec/functional_members.md, eventually superseded by proper DM support in Matrix.


AndrewFerr (approved):
LGTM

PR #1784, opened 2023-11-06
Libera Chat bridge is down

snan:


Half-Shot (approved):
Ugh, entirely forgot about this. Can you sign off the PR and then this can be merged?

PR #1787, opened 2024-01-06
Return promises inside Promise.all map

f0x52:
Currently this code doesn't return anything inside the `map()` callback, resulting in an array with undefined elements, and `Promise.all()` immediately returning.
https://github.com/matrix-org/matrix-appservice-irc/blob/b0fb7387ab22465804377dc4f5e81d6fa7a5ddb8/src/bridge/MatrixHandler.ts#L766-L768

Fixed it by adding a `return` but could also be styled as 
```js
await Promise.all(Object.keys(members).map((u) => this.membershipQueue.leave(event.room_id, u, req)));`
```

Caught this when I added my standard eslint config to my fork, might be useful to include the `array-callback-return` rule here too.


f0x52:
The eslint rule also catches this line, though there it doesn't really matter as the return value isn't used anyways (but forEach would be more semantically correct)
https://github.com/matrix-org/matrix-appservice-irc/blob/b0fb7387ab22465804377dc4f5e81d6fa7a5ddb8/src/bridge/RoomAccessSyncer.ts#L459-L461

Half-Shot (approved):
This is fine, but needs a signoff before I can merge it.

f0x52:
Signed-off-by: f0x <github@cthu.lu>

tadzik:
I think this is okay to merge, even without the changelog (it's a very internal change afaic).

PR #1788, opened 2024-01-22
Node18 updates and Timer -> Timeout change

warthog9:
Minor changes really but this is what got the bridge running for me on Node18 on Debian bookworm.  Mostly a minor update to the package resolution so it all matches, as well as changing the Timer to Timeout, which is what (clear|set)Timeout is using now seemingly.  Confirmed it's up and running and not seemingly doing anything amiss so submitting this upstream.

### DIFF
--- a/changelog.d/1788.misc
+++ b/changelog.d/1788.misc
@@ -1,0 +1,1 @@
+Fixup types for Timers so the bridge works with newer node versions.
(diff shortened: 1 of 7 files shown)


Half-Shot (approved):
Looks 100% sane to me. Can you add a changelog file so we can merge?

Half-Shot:
Signoff is in the commit.

PR #1790, opened 2024-01-25
Taking out LiberaChat

altf4arnold:
Hey. I saw that there were no commits since a long time. Since things have changed during that time, I added those changes.

### DIFF
--- a/changelog.d/1790.doc
+++ b/changelog.d/1790.doc
@@ -1,0 +1,1 @@
+Deleting LiberaChat from the list of bridged networks and adding documentation why it's not in there.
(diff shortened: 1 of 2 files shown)


morguldir:
See #1784

altf4arnold:
> See #1784

Indeed. I did not see it. But it seems to be stuck since a little while (also, does not provide further information for the reasons)

Half-Shot:
Hey, entirely my fault for missing that old PR again. I think is slightly more concise and since it came first, I'm going to merge it. Thanks for giving us a prod though!

PR #1797, opened 2024-03-18
Test both variants of message replies

tadzik:


tadzik:
Replaced by https://github.com/matrix-org/matrix-appservice-irc/pull/1799

PR #1798, opened 2024-03-19
Update our dependencies

tadzik:
Some are held back due to them requiring ESM (to be considered separately). That'd be `escape-string-regexp` and `quick-lru`.

`jasmine` from 4.0 onwards introduces a weird (seemingly) async bug with `room setup link endpoint should create a M<--->I link for a channel that has capital letters in it`. Afair we're planning to move away from Jasmine anyway, so it's probably not worth spending time on that now.

`matrix-bot-sdk` is held back due to it dropping `nedb`, which we still support – perhaps it's the right time to drop it here as well though?


src/datastore/postgres/PgDataStore.ts
```diff
@@ -453,7 +453,7 @@ export class PgDataStore implements DataStore, ProvisioningStore {
                     "server"
                 ],
             ),
-            [counter, homeserver || "*", server.domain],
+            [counter, homeserver || "*", server.domain] as any[],
```
tadzik, line 457: New version of @types/pg is very generic-happy, and insists on these being an array of a single type. explicitely casting them as `any` seems like the most sensible option here.

src/datastore/postgres/PgDataStore.ts
```diff
@@ -650,7 +650,7 @@ export class PgDataStore implements DataStore, ProvisioningStore {
         if (res.rowCount === 0) {
             return undefined;
         }
-        else if (res.rowCount > 1) {
+        else if (res.rowCount! > 1) {
```
tadzik, line 655: The reason for this sillyness is here: https://github.com/brianc/node-postgres/pull/2967

package.json
```diff
@@ -44,47 +44,47 @@
     "matrix-appservice-bridge": "^9.0.1",
```
Half-Shot: You can still update to 10, you just need to add nedb as a dependency :)

Half-Shot (commented):
This looks sane, but I think we *can* update matrix-appservice-bridge. Agree with deprecating nedb soon though!

PR #1799, opened 2024-03-20
Add tests for various forms of rich replies

tadzik:
Cleaned up and expanded https://github.com/matrix-org/matrix-appservice-irc/pull/1797

### DIFF
--- a/src/bridge/MatrixHandler.ts
+++ b/src/bridge/MatrixHandler.ts
@@ -393,8 +393,10 @@ export class MatrixHandler {
      * @param {Object} event : The Matrix member event.
      */
     private _onMemberEvent(req: BridgeRequest, event: OnMemberEventData) {
-        if (event.origin_server_ts) {
-            this.memberJoinTs.set(`${event.room_id}/${event.state_key}`, event.origin_server_ts);
+        if (event.content.membership === 'join') {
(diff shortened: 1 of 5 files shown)


src/bridge/MatrixHandler.ts
```diff
@@ -386,6 +393,9 @@ export class MatrixHandler {
      * @param {Object} event : The Matrix member event.
      */
     private _onMemberEvent(req: BridgeRequest, event: OnMemberEventData) {
+        if (event.origin_server_ts) {
```
Half-Shot: This is only joins, right?

Half-Shot (approved):
Looks reasonable, thanks for improving the test coverage!

src/bridge/MatrixHandler.ts
tadzik: Ooh, should be, yes.

Half-Shot: Ah, also we need to bound these by leave time too. Or maybe just delete the map entry, given you won't be replying to anything after you've left.

tadzik: Interestingly, it still works in this state, since your subsequent join will reset the map entry –so you won't be able to see messages from before your *most recent* join.

Still, I'd rather be explicit and have a test for it than rely on this quirk.

tadzik: (ffcba6f2866f35577424931d5949a743e2fcf497)

PR #1805, opened 2024-05-29
Integrate MediaProxy to bridge authenticated Matrix media (MSC3916)

tadzik:
Depends on https://github.com/matrix-org/matrix-appservice-bridge/pull/444


turt2live:
just to confirm: does this mean to say MSC3916 instead of MSC3910? 3910 is a closed, but related, MSC.

tadzik:
> just to confirm: does this mean to say MSC3916 instead of MSC3910? 3910 is a closed, but related, MSC.

Yes, correct - I copied the old number from the MAB PR :)

heftig:
Is a release with this planned before Synapse defaults to disallowing unauthenticated media?

tadzik:
> Is a release with this planned before Synapse defaults to disallowing unauthenticated media?

Yes, that's the plan.

src/config/BridgeConfig.ts
```diff
@@ -13,7 +13,6 @@ export interface BridgeConfig {
     };
     homeserver: {
         url: string;
-        media_url?: string;
```
Half-Shot, line 16: It would be kind somewhere to fail to start if this is still configured?

src/bridge/IrcBridge.ts
```diff
@@ -255,9 +260,10 @@ export class IrcBridge {
             log.info(`Adjusted dropMatrixMessagesAfterSecs to ${newConfig.homeserver.dropMatrixMessagesAfterSecs}`);
         }
 
-        if (oldConfig.homeserver.media_url !== newConfig.homeserver.media_url) {
-            oldConfig.homeserver.media_url = newConfig.homeserver.media_url;
-            log.info(`Adjusted media_url to ${newConfig.homeserver.media_url}`);
+        if (JSON.stringify(oldConfig.ircService.mediaProxy) !== JSON.stringify(newConfig.ircService.mediaProxy)) {
+            await this._mediaProxy?.close();
```
Half-Shot, line 263: Very nice!

src/bridge/IrcBridge.ts
```diff
@@ -209,6 +213,7 @@ export class IrcBridge {
             defaultTtlMs: 10 * 60 * 1000, // 10 mins
         });
         this.matrixBanSyncer = this.config.ircService.banLists && new MatrixBanSync(this.config.ircService.banLists);
+
```
Half-Shot: rando space?

config.sample.yaml
```diff
@@ -607,6 +601,18 @@ ircService:
     # Ignore users mentioned in a io.element.functional_members state event when checking admin room membership
     ignoreFunctionalMembersInAdminRooms: false
 
+  # Config for the media proxy, required to serve publically accessible URLs to authenticated Matrix media
+  mediaProxy:
+    # To generate a .jwk file:
+    # $ yarn -s ts-node src/generate-signing-key.ts > signingkey.jwk
```
Half-Shot: Should we make a script entry for this to shortcut it a bit.

And is `ts-node` available by default?

Half-Shot (commented):
This looks good on the whole, only a few suggestions. Should we have tests though?

config.sample.yaml
tadzik: Good point, `ts-node` is just a dev-dependency so far. Maybe we can even make it built-in, like with generating registration files (https://matrix-org.github.io/matrix-appservice-irc/latest/bridge_setup.html#4-registration)?

src/config/BridgeConfig.ts
tadzik: I'd keep it optional (maybe with a warning?) to make sure you can rollback easily without changing the configs too much.

We do require `mediaProxy` being configured now, so users will surely notice the change.

src/bridge/IrcBridge.ts
tadzik: Dropped it in 395a8c1f0a848e57f09c5f81cd25a425840f7768

tadzik:
> This looks good on the whole, only a few suggestions. Should we have tests though?

We have some tests covering the case of "do these URLs look like MediaProxy's URLs" – I'm not sure how much further we want to take this, if it all.

config.sample.yaml
Half-Shot: how chunky is it?

Half-Shot:
> > This looks good on the whole, only a few suggestions. Should we have tests though?
> 
> We have some tests covering the case of "do these URLs look like MediaProxy's URLs" – I'm not sure how much further we want to take this, if it all.

I think that's sufficient, upstream has tests that the MP works?

src/generate-signing-key.ts
```diff
@@ -0,0 +1,11 @@
+import { webcrypto } from 'node:crypto';
```
Half-Shot: To be honest, this could be a JavaScript file and save us a ts-node. We're not really even using any types here other than the Node API.

spec/e2e/jest.config.js
```diff
@@ -2,7 +2,7 @@
 module.exports = {
   preset: 'ts-jest',
   testEnvironment: 'node',
-  reporters: [['github-actions', {silent: false}], 'summary'],
+  // reporters: ['text', ['github-actions', {silent: false}], 'summary'],
```
Half-Shot: oh no!

Half-Shot (approved):
I see no problems here. I have one concern about the test reporter :)

spec/e2e/jest.config.js
tadzik: Oh no.

config.sample.yaml
tadzik: (We ended up converting it to raw JS).

Issue #1521, opened 2022-02-08
Long quotes in replies


progval:
**Describe the bug**

Seen earlier today (10:42 UTC) on #matrix-irc on Libera

```
10:42:43 <jess[m]> > <@ilmari:ilmari.org> `!listrooms` says
10:42:43 <jess[m]> > > You are connected, but not joined to any channels.
10:42:43 <jess[m]> >
10:42:43 <jess[m]> > but `/whois` from the IRC side says
10:42:43 <jess[m]> >
10:42:44 <jess[m]> > > There is no such nick ilmari_
10:42:44 <jess[m]> still having problems?
```

**To Reproduce**

No idea. Possibly just replying to a multi-line message.

**Expected behavior**

Should probably be just this:

```
<jess[m]> ilmari_: still having problems?
```

Half-Shot:
That looks more like @jessopo quoted a message rather than a reply?

progval:
I found another example:

```
{
  "content": {
[...]

PR #1807, opened 2024-06-14
Initial support for ban syncing + minor improvements and bugfixes

funderscore1:
Few improvements to the bridge:

**Change format for bridging file uploads and codeblocks to something more natural to IRC users**
For file uploads:
```
<f_[mtrx]> https://matrix.org/_matrix/media/v3/download/something/some-other-thing/image.png (20KiB)
```
For code blocks:
```
<f_[mtrx]> https://matrix.org/_matrix/media/v3/download/something/some-other-thing
```
**Change default format for replies, again, to something that's a bit more natural**
New format:
```
<f_[mtrx]> f_: "hi everyone!..." <- hi!
```

**Handle replying to self**
New format (long):
```
<f_[mtrx]> <f_[mtrx]> what is this
<f_[mtrx]> ^
```

**Use "\`" instead of an "M" when the nick starts with an invalid character**
This is more in-line with what a usual IRC user would do, and less noticeable.

**Fix bug when replying to a long message**
See https://github.com/matrix-org/matrix-appservice-irc/issues/1521

**In `!cmd`, don't require commands be uppercase**
IRC commands are, for the most part, case-insensitive. For example, with `NICK`:
```
NICK funderscore1
:funderscore123!~0@4G4AABC5T.tor-irc.dnsbl.oftc.net NICK :funderscore1
nick funderscore12
:funderscore1!~0@4G4AABC5T.tor-irc.dnsbl.oftc.net NICK :funderscore12
Nick funderscore1
:funderscore12!~0@4G4AABC5T.tor-irc.dnsbl.oftc.net NICK :funderscore1
NIcK funderscore12
:funderscore1!~0@4G4AABC5T.tor-irc.dnsbl.oftc.net NICK :funderscore12
```
Or `PING`:
```
Ping :D
:helix.oftc.net PONG helix.oftc.net :D
PING :D
:helix.oftc.net PONG helix.oftc.net :D
pIng :D
:helix.oftc.net PONG helix.oftc.net :D
ping :D
:helix.oftc.net PONG helix.oftc.net :D
```

**Fix a formatting issue in `!help`**

Doesn't need much explanation.

## Initial support for bridging Matrix bans to IRC
This is the big one. The bridge used to treat matrix bans as kicks, thus only kick the IRC user that has been banned on Matrix.

This resulted in IRC users still annoying other IRC users while the Matrix side didn't notice anything at all.

Now, matrix bans are bridged as `/mode +b ${nick}!*@*` and `/kick ${nick}`. Unbanning is still unimplemented, so moderators will need to manually send a raw command to the admin room: `!cmd mode #channel -b ${nick}!*@*`.

---

Feedback welcome!

### DIFF
--- a/src/bridge/MatrixHandler.ts
+++ b/src/bridge/MatrixHandler.ts
@@ -1373,7 +1373,7 @@ export class MatrixHandler {
         const bridgeIntent = this.ircBridge.getAppServiceBridge().getIntent();
         // strips out the quotation of the original message, if needed
         const replyText = (body: string): string => {
-            const REPLY_REGEX = /> <(.*?)>(.*?)\n\n([\s\S]*)/;
+            const REPLY_REGEX = /> <(.*?)>(.*?)\n\n([\s\S]*)/s;
             const match = REPLY_REGEX.exec(body);
             if (match === null || match.length !== 4) {
                 return body;
--- a/src/bridge/MatrixHandler.ts
+++ b/src/bridge/MatrixHandler.ts
@@ -1252,7 +1252,6 @@ export class MatrixHandler {
             // we check event.content.body since ircAction already has the markers stripped
             const codeBlockMatch = event.content.body.match(/^```(\w+)?/);
             if (codeBlockMatch) {
-                const type = codeBlockMatch[1] ? ` ${codeBlockMatch[1]}` : '';
                 event.content = {
                     ...event.content,
                     body:    `${httpUrl}`
--- a/src/bridge/MatrixHandler.ts
+++ b/src/bridge/MatrixHandler.ts
@@ -1177,8 +1177,8 @@ export class MatrixHandler {
             if (codeBlockMatch) {
                 const type = codeBlockMatch[1] ? ` ${codeBlockMatch[1]}` : '';
                 event.content = {
-                    msgtype: "m.emote",
-                    body:    `sent a${type} code block: ${httpUrl}`
+                    ...event.content,
+                    body:    `${httpUrl}`
                 };
             }
             else {
@@ -1209,7 +1209,7 @@ export class MatrixHandler {
             // Modify the event to become a truncated version of the original
             //  the truncation limits the number of lines sent to lineLimit.
 
-            const msg = '\n...(truncated)';
+            const msg = '\n(truncated)';
 
             const sendingEvent: MatrixMessageEvent = { ...event,
                 content: {
--- a/src/bridge/MatrixHandler.ts
+++ b/src/bridge/MatrixHandler.ts
@@ -67,7 +67,7 @@ export const DEFAULTS: MatrixHandlerConfig = {
     replySourceMaxLength: 32,
     shortReplyTresholdSeconds: 5 * 60,
     shortReplyTemplate: "$NICK: $REPLY",
-    longReplyTemplate: "<$NICK> \"$ORIGINAL\" <- $REPLY",
+    longReplyTemplate: "$NICK: \"$ORIGINAL\" <- $REPLY",
     truncatedMessageTemplate: "(full message at <$URL>)",
     ignoreFunctionalMembersInAdminRooms: false,
 };
--- a/src/bridge/MatrixHandler.ts
+++ b/src/bridge/MatrixHandler.ts
@@ -652,7 +652,7 @@ export class MatrixHandler {
 
     private async _onKick(req: BridgeRequest, event: MatrixEventKick, kicker: MatrixUser, kickee: MatrixUser) {
         req.log.info(
-            "onKick %s is kicking/banning %s from %s (reason: %s)",
+            "onKick %s is kicking %s from %s (reason: %s)",
             kicker.getId(), kickee.getId(), event.room_id, event.content.reason || "none"
         );
         this._onMemberEvent(req, event);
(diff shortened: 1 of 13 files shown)


funderscore1:
This is now ready for review.

funderscore1:
ping?

funderscore1:
gentle ping - could anyone review this please?

src/bridge/MatrixHandler.ts
```diff
@@ -1395,10 +1472,19 @@ export class MatrixHandler {
...
+            else {
+                replyTemplate = this.config.longReplyTemplate;
+            }
         }
         else {
-            replyTemplate = this.config.shortReplyTemplate;
+            if (cachedEvent.sender !== event.sender) {
+                // Someone[m] pinging themself is weird
+                replyTemplate = this.config.shortReplyTemplate;
+            }
```
tadzik: `else` doesn't exist here, leaving replyTemplate possibly undefined. This currently causes a compilation error:

```
src/bridge/MatrixHandler.ts:1490:47 - error TS2454: Variable 'replyTemplate' is used before being assigned.

1490         const formattedReply = renderTemplate(replyTemplate, {
                                                   ~~~~~~~~~~~~~


```

src/bridge/MatrixHandler.ts
```diff
@@ -741,7 +744,81 @@ export class MatrixHandler {
...
+            // Matrix on IRC banning, work out which IRC user to ban.
+            let server = null;
+            for (let i = 0; i < ircRooms.length; i++) {
+                if (ircRooms[i].server.claimsUserId(banned.getId())) {
+                    server = ircRooms[i].server;
+                    break;
+                }
+            }
+            if (!server) {
+                return; // kicking a bogus user
```
tadzik, line 776: Let's have at least a `warn` here to know that something unexpected is happening. Would help us investigate it in the future, if we end up here for wrong reasons.

src/bridge/MatrixHandler.ts
```diff
@@ -741,7 +744,81 @@ export class MatrixHandler {
...
+            }
+            const bannedNick = server.getNickFromUserId(banned.getId());
+            if (!bannedNick) {
+                return; // bogus virtual user ID
```
tadzik, line 780: As above.

src/bridge/MatrixHandler.ts
```diff
@@ -741,7 +744,81 @@ export class MatrixHandler {
...
+                    "%s has no client instance to send kick from. Cannot kick.",
+                    sender.getId()
+                );
+                return;
+            }
+            // we may be bridging this matrix room into many different IRC channels, and we want
+            // to kick this user from all of them.
+            for (let i = 0; i < ircRooms.length; i++) {
+                if (ircRooms[i].server.domain !== server.domain) {
+                    return;
```
tadzik, line 796: Shouldn't this be a `continue`?

src/irc/BridgedClient.ts
```diff
@@ -510,6 +510,26 @@ export class BridgedClient extends EventEmitter {
         await c.send("KICK", channel, nick, reason);
     }
 
+    public async ban(nick: string, channel: string): Promise<void> {
+        if (this.state.status !== BridgedClientStatus.CONNECTED) {
+            return; // we were never connected to the network.
```
tadzik, line 515: I was about to point out quiet failures here, but I see that it follows `onKick()`. Not a fan, but I'm not going to complain about this now, the PR is big enough already.

src/irc/BridgedClient.ts
```diff
@@ -740,7 +760,7 @@ export class BridgedClient extends EventEmitter {
             }
             // Add arbitrary letter prefix. This is important for guest user
             // IDs which are all numbers.
-            n = "M" + n;
+            n = "`" + n;
```
tadzik, line 763: Now it's no longer a letter prefix.

I get the idea of the backtick being more IRC-natural, but 1) I'd expect it as a suffix rather than a prefix, so that usual autocompletion still works in clients, and 2) this is just for guest accounts so I wouldn't worry about it too much.

src/models/IrcAction.ts
```diff
@@ -54,18 +54,18 @@ export class IrcAction {
                 return new IrcAction(matrixAction.type, matrixAction.text, matrixAction.ts);
             case "image":
                 return new IrcAction(
-                    "emote", "uploaded an image: " + matrixAction.text, matrixAction.ts
+                    "message", "" + matrixAction.text, matrixAction.ts
```
tadzik, line 57: I wonder if we want these to be templatable as well. Not a real CR issue, just thinking out loud while we're here.

src/models/MatrixAction.ts
```diff
@@ -220,12 +220,12 @@ export class MatrixAction {
 
                 if (filename) {
                     url += `/${encodeURIComponent(filename)}`;
-                    text = `${fileSize} < ${url} >`;
+                    text = `${url} ${fileSize}`;
```
tadzik, line 223: The PR note mentions size in parens (`<f_[mtrx]> https://matrix.org/_matrix/media/v3/download/something/some-other-thing/image.png (20KiB)`), here there aren't any. Which one is intentional?

(Maybe this is also template-worthy).

tadzik (changes requested):
Sorry for the long wait.

I generally agree with everything, there's just a few nitpicks. Nice work!

Can we get test to selfReplyTemplate as well? Should be easy enough to cargo-cult the other ones. This would give us an excuse to cover the 4th case (the one that's currently a compilation error).

src/irc/BridgedClient.ts
tadzik: On second thoughts, I wonder where does the letter (non-number?) prefix requirement actually come from, and why does M or the backtick fix it.

Trying on ergo locally, both `123` and ``123` are valid nicknames – where does this being required actually come into play?

funderscore1: Most IRCd's don't allow numbers and `-` at the start of nicknames. Ergo is one of the rare (only?) IRCd's that allow these. https://modern.ircdocs.horse/#nick-message says:

> Servers MUST allow at least all alphanumerical characters, square and curly brackets ([]{}), backslashes (\), and pipe (|) characters in nicknames, **and MAY disallow digits as the first character**

Numbers are usually not allowed as the first character in nicknames, as UIDs start with those.

> this is just for guest accounts so I wouldn't worry about it too much

I met a whole lot of people with numbers at the start of their nickname.

> I get the idea of the backtick being more IRC-natural, but 1) I'd expect it as a suffix rather than a prefix, so that usual autocompletion still works in clients

Not sure what you mean there.

src/models/MatrixAction.ts
funderscore1: `${fileSize}` has parentheses already.

src/irc/BridgedClient.ts
funderscore1: Yes, it is basically onKick, but bans before kicking. Shall I resolve the other comments above?

src/bridge/MatrixHandler.ts
funderscore1: Hm, I wonder why I never noticed that.

src/models/IrcAction.ts
funderscore1: I'm not sure how useful that would be, IMO. This is what IRC users do - just send a link, I don't know why should the bridge behave differently (e.g. by sending an entire `/me` saying `that person sent an image` before the actual link).

src/bridge/MatrixHandler.ts
funderscore1: That would mean changing `onKick()`'s behaviour as well, to be consistent

src/irc/BridgedClient.ts
tadzik: Sorry, got confused between this feature and another. This is fine, and IMO a better choice than "M"

src/bridge/MatrixHandler.ts
funderscore1: TODO for next time.

tadzik (commented):
LGTM but we're failing all sorts of tests

tadzik:
Plenty of test regressions here. Tricky to track down, since we're changing so many things at once.

Test "should allow arbitrary IRC commands to be issued" fails presumably due to command casing changes in 7956631eaeaca8e890792f9ec8f5258808410eaa

fb0a326d68b88531c4b93fc912e898ca1e43e5e5 changes some Matrix->IRC messages from `emote` to `message`, which trips up the tests that expect `action` instead of `message`. This manifests itself as a timeout, since the expected "action" never arrives, so the tests waits too long.

I know it's late in the process, but I think it'd make sense to split this PR into a few smaller ones, so that the big features (like the ban syncing) don't block the minor things (like case-insensitive command handling). Not only it'll make it obvious what broke which test, it'll make the changelog much neater – we currently expect these to be a single item pointing to a single PR, and mega-PRs like this one would make it a lot trickier to process.

tadzik:
One linting error away and the changelog, and we should be good to go

tadzik:
I've split it into https://github.com/matrix-org/matrix-appservice-irc/pull/1821, https://github.com/matrix-org/matrix-appservice-irc/pull/1822 and https://github.com/matrix-org/matrix-appservice-irc/pull/1823, hopefully I didn't miss anything.

PR #1821, opened 2024-09-12
Support for ban syncing

tadzik:
Split off from https://github.com/matrix-org/matrix-appservice-irc/pull/1807, credit to @funderscore1

### DIFF
--- a/spec/integ/kicking.spec.js
+++ b/spec/integ/kicking.spec.js
@@ -176,6 +176,118 @@ describe("Kicking", () => {
 });
 
 
+describe("Banning", () => {
+
+    const {env, config, test} = envBundle();
+
+    const mxUser = {
+        id: "@flibble:wibble",
+        nick: "M-flibble"
+    };
+
+    const ircUser = {
+        nick: "bob",
+        localpart: config._server + "_bob",
+        id: `@${config._server}_bob:${config.homeserver.domain}`
+    };
+
+    const ircUserKicker = {
+        nick: "KickerNick",
+        localpart: config._server + "_KickerNick",
+        id: "@" + config._server + "_KickerNick:" + config.homeserver.domain
+    };
+
+    beforeEach(async () => {
+        await test.beforeEach(env);
+
+        // accept connection requests from eeeeeeeeveryone!
+        env.ircMock._autoConnectNetworks(
+            config._server, mxUser.nick, config._server
+        );
+        env.ircMock._autoConnectNetworks(
+            config._server, ircUser.nick, config._server
+        );
+        env.ircMock._autoConnectNetworks(
+            config._server, config._botnick, config._server
+        );
+        // accept join requests from eeeeeeeeveryone!
+        env.ircMock._autoJoinChannels(
+            config._server, mxUser.nick, config._chan
+        );
+        env.ircMock._autoJoinChannels(
+            config._server, ircUser.nick, config._chan
+        );
+        env.ircMock._autoJoinChannels(
+            config._server, config._botnick, config._chan
+        );
+
+        // we also don't care about registration requests for the irc user
+        env.clientMock._intent(ircUser.id)._onHttpRegister({
+            expectLocalpart: ircUser.localpart,
+            returnUserId: ircUser.id
+        });
+
+        await test.initEnv(env);
+
+        // make the matrix user be on IRC
+        await env.mockAppService._trigger("type:m.room.message", {
+            content: {
+                body: "let me in",
+                msgtype: "m.text"
+            },
+            user_id: mxUser.id,
+            room_id: config._roomid,
+            type: "m.room.message"
+        });
+        const botIrcClient = await env.ircMock._findClientAsync(config._server, config._botnick);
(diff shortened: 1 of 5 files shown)


tadzik:
Looks good to me, and I've heard it's been successfully running on pixie.town for a while.

spec/integ/kicking.spec.js
```diff
@@ -176,6 +176,118 @@ describe("Kicking", () => {
 });
 
 
+describe("Banning", () => {
```
Half-Shot, line 179: file says kicking
test says banning

new test file time?

src/bridge/MatrixHandler.ts
```diff
@@ -741,7 +741,81 @@ export class MatrixHandler {
...
+                );
+                return;
+            }
+            // we may be bridging this matrix room into many different IRC channels, and we want
+            // to kick this user from all of them.
+            for (let i = 0; i < ircRooms.length; i++) {
+                if (ircRooms[i].server.domain !== server.domain) {
+                    return;
+                }
+                senderClient.ban(bannedNick, ircRooms[i].channel);
```
Half-Shot, line 795: Do we need to await either of these?

src/irc/BridgedClient.ts
```diff
@@ -510,6 +510,26 @@ export class BridgedClient extends EventEmitter {
...
+        if (!channel.startsWith("#")) {
+            return; // PM room
+        }
+
+        const c = this.state.client;
+
+        this.log.debug("Banning %s from channel %s", nick, channel);
+
+        // best effort ban
+        await c.send("MODE", channel, "+b", nick + "!*@*");
```
Half-Shot, line 530: I'm always a bit skeptical about these. I presume we're sure this is reasonably implementation independent. 

We don't want to offer a ban command config option or anything like that?

src/bridge/IrcBridge.ts
```diff
@@ -1254,18 +1254,20 @@ export class IrcBridge {
...
+                // we know it is a kick (or a rescinded invite)
+                const isKick = target.getId() !== sender.getId();
+                if (isKick) {
                     await this.matrixHandler.onKick(request, memberEvent as unknown as MatrixEventKick, sender, target);
                 }
                 else {
                     await this.matrixHandler.onLeave(request, memberEvent, target);
                 }
             }
+            else if (event.content.membership === "ban") {
```
Half-Shot, line 1268: How do unban?

Half-Shot (changes requested):
This is fine, but this isn't ban syncing. This needs to handle the case of unbanning too, or at the very least some technical documentation to explain why we don't do that.

And actually I'd love some documentation on our docs site on what users can expect from kicks / bans.

src/irc/BridgedClient.ts
funderscore1: Why would this be implementation dependent?

src/bridge/IrcBridge.ts
funderscore1: Not implemented because I didn't wrap my head on how unbanning works on Matrix and how to check that.

src/irc/BridgedClient.ts
funderscore1: I'd love to know how we could offer ban config when banning from Matrix.

src/bridge/MatrixHandler.ts
funderscore1: Why?

f0x52:
I think this could nicely be supplemented with an `!irc ban <mask>` command which sets +b modes, to allow for more flexible matching on nick/user/host parts.

Afaik removing a ban on the matrix side consists of the user's `m.room.member` state event changing to anything that isn't `ban`, so you could check the `prev_content` key in state update events to unset the mode

src/bridge/MatrixHandler.ts
Half-Shot: Because you're doing two operations here, and I'm not sure if one should proceed if the other fails? At the very least, run them both concurrently but await the result so we can throw an error. Node typically crashes the process these days if a unawaited promise throws.

funderscore1: I think if `+b` fails (not sure why it would) and `/kick` works that's better than if it fails and `/kick` isn't ran because of that.

funderscore1: but sure I'll look into that

funderscore1:
> I think this could nicely be supplemented with an `!irc ban <mask>` command which sets +b modes, to allow for more flexible matching on nick/user/host parts.
> 
> Afaik removing a ban on the matrix side consists of the user's `m.room.member` state event changing to anything that isn't `ban`, so you could check the `prev_content` key in state update events to unset the mode

@f0x52 I'll look into that, thanks

spec/integ/kicking.spec.js
funderscore1: How about renaming `kicking.spec.js` to `kicking-banning.spec.js`? I think they go well in one file.

PR #1822, opened 2024-09-12
Various formatting adjustments

tadzik:
Formatting-related changes from https://github.com/matrix-org/matrix-appservice-irc/pull/1807, including:

 - Better matching IRC conventions when bridging file uploads
 - Changing default reply-to format
 - Replacing "M" letter prefix with "`" when nick starts with an invalid character
 - A fix for #1521
 - Sensible handling of self-replies

Credit to @funderscore1


tadzik (commented):
Github won't let me approve this since I'm technically the author – but LGTM, good to merge IMO.

PR #1823, opened 2024-09-12
Don't require commands be uppercase when using `!cmd`

tadzik:
Split off from https://github.com/matrix-org/matrix-appservice-irc/pull/1807, credit to @funderscore1

### DIFF
--- a/spec/integ/admin-rooms.spec.js
+++ b/spec/integ/admin-rooms.spec.js
@@ -873,7 +873,7 @@ describe("Admin rooms", function() {
             const newChannel = "#coffee";
 
             // Expect the following commands to be sent in order
-            const recvCommands = ["JOIN", "TOPIC", "PART", "STUPID"];
+            const recvCommands = ["JOIN", "TOPIC", "PART", "STUPID", "lowercase"];
 
             let cmdIx = 0;
             env.ircMock._whenClient(roomMapping.server, userIdNick, "send",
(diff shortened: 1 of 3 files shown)


spec/integ/admin-rooms.spec.js
```diff
@@ -885,14 +885,12 @@ describe("Admin rooms", function() {
...
 
-            // 5 commands should be executed
-            // rubbishserver should not be accepted
+            // 4 commands should be executed
             const commands = [
                 `!cmd ${roomMapping.server} JOIN ${newChannel}`,
                 `!cmd ${roomMapping.server} TOPIC ${newChannel} :some new fancy topic`,
                 `!cmd ${roomMapping.server} PART ${newChannel}`,
-                `!cmd ${roomMapping.server} STUPID COMMANDS`,
-                `!cmd rubbishserver SOME COMMAND`];
```
Half-Shot, line 895: we've lost the test for an invalid server here?

Half-Shot (approved):
Seems reasonable aside from a lost test?

spec/integ/admin-rooms.spec.js
```diff
@@ -885,14 +885,12 @@ describe("Admin rooms", function() {
                     cmdIx++;
                 });
 
-            // 5 commands should be executed
-            // rubbishserver should not be accepted
+            // 4 commands should be executed
             const commands = [
                 `!cmd ${roomMapping.server} JOIN ${newChannel}`,
```
Half-Shot, line 889: We still need to test lower case though?

Half-Shot (commented):
oh and

spec/integ/admin-rooms.spec.js
funderscore1: With this change "rubbishserver" is now interpreted as a command rather than as a server.

Half-Shot: right...we still need to interpret servers here.

funderscore1: They are interpreted correctly when valid, but `rubbishserver` matches the new regex.

Perhaps that could be replaced by `rubbish^server!` or something?

tadzik: Or we could have two servers configured to ensure that the first string needs to match a server.

I think it's okay to let this one slide though. We're not breaking anything anything except wrong usage of commands, and I don't think there's much of a risk of a server being called `quit` or something that could land a user in (minor) trouble.

spec/integ/admin-rooms.spec.js
tadzik: Added in 18460950

PR #1832, opened 2024-11-22
Bridge K-lines into Matrix policy lists

tadzik:
Currently, while K-lined users cannot connect to IRC and continue their abuse there, they're not being banned on the Matrix side. This allows to not only hang around in bridged Matrix rooms until manually removed from them, but also continue their abuse on Matrix itself.

This enhances our banSyncing capabilities to enable us to not only to subscribe to existing policylists, but also maintain our own. Each IRC network gets its own policy room (so that they can be subscribed to by mjolnir etc independently) which serves not only as a persistency layer for `MatrixBanSync` which stops us from pointlessly trying to (re)connect K-lined users, but also enables us to protect the wider Matrix network from those who proved to be abusive already.

Currently this does not bother to re-check if the K-line is still active – temporarily banned users have to manually arrange for being removed from the policy list, which the bridge will catch up and reconnect them if needed. We may want to consider allowing them to trigger the re-check manually, perhaps by issuing an explicit `!reconnect` in the bridge status room.

### DIFF
--- a/config.sample.yaml
+++ b/config.sample.yaml
@@ -668,6 +668,10 @@ ircService:
     # to identify banned users, rooms and servers.
     rooms:
       - "#matrix-org-coc-bl:matrix.org"
+    # Optional: Rooms to sync K-lined users to
+    serverBanLists:
+      # Keys are domains matching entries in ircService.servers
+      irc.example.com: "#exampleircd:matrix.org"
 
 # Options here are generally only applicable to large-scale bridges and may have
 # consequences greater than other options in this configuration file.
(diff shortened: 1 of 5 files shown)


jkbzh:
A naive question, shouldn't you take into account G-Lines in your patch? And also, will this keep bridges from trying to connect over and over to an irc server from which they are banned? I see this behavior on an irc server I admin. The bridges just ignore the G-Line so the server keeps closing the connection, and right away the bridges re-open it. This is forcing us to apply the ban at the TCP layer. 

It is as if the bridges were designed to always try to reconnect when a connection is closed, but they are ignoring the G-Line and K-Line reasons in the connection shutdown reply.  If I understood correctly, this is part of what your patch will solve. 

If you could take into account G-Line reasons, it would be great! 

Thanks.

PR #1853, opened 2025-08-05
MessageDiff: generate proper regexes on message tail modification

oI0ck:
This PR fixes faulty logic generating bad regexes on message edits in a following case:

In a user sent out a message before finishing it, with an unfinished last word, like here:
`how many wood would a woodchuck c`
and then edited it to
`how many wood would a woodchuck chuck if a woodchuck would chuck wood`,
the bridge would output `s/c/chuck if a woodchuck would chuck wood/`, which,
if evaluated, produces `how many wood would a woodchuck if a woodchuck would chuck woodhuck chuck`.

The bridge did not take the possibility of the replaced part occurring more than one time into the account.

These changes fix it by introducing the regex dollar operator for matching line endings in this very specific case.


oI0ck:
Pinging :3

Issue #1857, opened 2025-09-23
Receiving `IO.ELEMENT.MSC4190.M_APPSERVICE_LOGIN_UNSUPPORTED` from Synapse upon registering Matrix user


anoadragon453:
We are seeing the following exceptions being generated on the matrix.org instance of this bridge:

```
Error: IO.ELEMENT.MSC4190.M_APPSERVICE_LOGIN_UNSUPPORTED: This server uses OAuth2, so the inhibit_login parameter must be set to true for appservice registrations.
  File "/opt/oftcirc/matrix-appservice-irc-3.0.5/node_modules/@vector-im/matrix-bot-sdk/lib/http.js", line 95, col 15, in doHttpRequest
    throw new MatrixError_1.MatrixError(errBody, response.statusCode);
  File "node:internal/process/task_queues", line 95, col 5, in process.processTicksAndRejections
  File "/opt/oftcirc/matrix-appservice-irc-3.0.5/node_modules/@vector-im/matrix-bot-sdk/lib/metrics/decorators.js", line 19, col 32, in descriptor.value
    const result = await originalMethod.apply(this, args);
  File "/opt/oftcirc/matrix-appservice-irc-3.0.5/node_modules/@vector-im/matrix-bot-sdk/lib/appservice/Intent.js", line 276, col 32, in Intent.ensureRegistered
    const result = await this.client.doRequest("POST", "/_matrix/client/v3/register", null, {
  File "/opt/oftcirc/matrix-appservice-irc-3.0.5/node_modules/@vector-im/matrix-bot-sdk/lib/metrics/decorators.js", line 78, col 32, in descriptor.value
    const result = await originalMethod.apply(this, args);
  File "/opt/oftcirc/matrix-appservice-irc-3.0.5/node_modules/matrix-appservice-bridge/lib/components/intent.js", line 892, col 17, in Intent.ensureRegistered
    await this.botSdkIntent.ensureRegistered();
  File "/opt/oftcirc/matrix-appservice-irc-3.0.5/node_modules/matrix-appservice-bridge/lib/components/intent.js", line 503, col 9, in Intent.setDisplayName
    await this.ensureRegistered();
  File "/opt/oftcirc/matrix-appservice-irc-3.0.5/lib/bridge/IrcBridge.js", line 949, col 9, in IrcBridge.getMatrixUser
    await userIntent.setDisplayName(displayName); // will also register this user
[...]

PR #1858, opened 2025-10-20
Update matrix-appservice-bridge to fix `IO.ELEMENT.MSC4190.M_APPSERVICE_LOGIN_UNSUPPORTED` error with MSC4190 enabled homeservers

jaywink:
Also remove the extra `matrix-bot-sdk` import since we get it through `matrix-appservice-bridge` and should not be defining two competing versions.

Refs: https://github.com/matrix-org/matrix-appservice-irc/issues/1857

Node version needed updating to the requirement from matrix-appservice-bridge. Docker also needed a base image bump as the build against buster was failing. The node version update caused some code changes where TypeScript with the new types was unhappy about existing code.


src/bridge/IrcBridge.ts
```diff
@@ -1582,20 +1582,20 @@ export class IrcBridge {
...
-            catch (ex) {
-                // We may not have permissions to do so, which means we are basically stuffed.
-                log.warn(`Could not send bridge info event to new room: ${ex}`);
+            if (bridgeInfoEvent) {
+                try {
+                    await this.bridge.getIntent().sendStateEvent(
+                        newRoomId,
+                        bridgeInfoEvent.type,
+                        bridgingEvent.state_key,
+                        bridgingEvent.content
```
AndrewFerr, line 1591: Hmm, was this supposed to use the state key & content of the `bridgeInfoEvent` all along?

If so, this could be moved back outside of the scope of `if (bridgingEvent)`.

AndrewFerr (changes requested):
The workflow files also need updating to run on Node 22.  Consider updating the `.node-version` file & setting all workflow files to use it via `node-version-file` (which is currently done only by the E2E tests workflow).

src/bridge/IrcBridge.ts
jaywink: Hard to say, we would need someone with sufficient experience on the bridge to make that call. Certainly I don't think we should be introducing such a change in this pr, as it changes functionality unrelated to this fix. The change I made is to guard against passing `undefined` to `sendStateEvent` which down the line afaict will cause:

```
const path = "/_matrix/client/v3/rooms/"
            + encodeURIComponent(roomId) + "/state/"
            + encodeURIComponent(type) + "/"
            + encodeURIComponent(stateKey);
```

Where

```
> encodeURIComponent(undefined)
'undefined'
```

Which is probably not what we want. Granted, we should probably not be even hitting this as if `bridgingEvent` is not defined/null, we would have an exception calling `bridgingEvent.state_key`. But TS seems to have figured this out and become stricter, unsure how this even passed any previous checks? Thus some kind of change is needed in this pr, and this seemed the least invasive.

```
src/bridge/IrcBridge.ts:1591:21 - error TS18048: 'bridgingEvent' is possibly 'undefined'.
1591                     bridgingEvent.state_key,
                         ~~~~~~~~~~~~~
src/bridge/IrcBridge.ts:1592:21 - error TS18048: 'bridgingEvent' is possibly 'undefined'.
1592                     bridgingEvent.content
```

Basically whatever can be done to resolve this is fine for me as long as we don't introduce behaviour change, suggest we introduce that in a separate pr, if necessary, as you say maybe this is just a bug.

AndrewFerr: Good point, it's better to be conservative for now & keep this PR focused on just the MSC4190 fix.

AndrewFerr (changes requested):
The hanging workflows should update their runner image from [the now-unsupported Ubuntu 20.04](https://github.blog/changelog/2025-01-15-github-actions-ubuntu-20-runner-image-brownout-dates-and-other-breaking-changes/#ubuntu-20-image-is-closing-down) to something more recent.

AndrewFerr (approved):

AndrewFerr (changes requested):
ESLint deps should be updated, because the versions currently in use don't support the bumped TypeScript version.

spec/util/irc-client-mock.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable @typescript-eslint/no-require-imports */
```
AndrewFerr, line 1: Any reason to not update these `require`s?
Bonus: `defer` is unused, so it can be removed.
```diff --git
diff --git a/spec/util/irc-client-mock.js b/spec/util/irc-client-mock.js
index f7f543b0..cccd7ee5 100644
--- a/spec/util/irc-client-mock.js
+++ b/spec/util/irc-client-mock.js
@@ -1,12 +1,10 @@
-/* eslint-disable @typescript-eslint/no-require-imports */
 /*
  * Mock replacement for 'irc'.
  */
 "use strict";
-const Promise = require("bluebird");
-const util = require("util");
-const EventEmitter = require('events').EventEmitter;
-const defer = require("../../lib/promiseutil").defer;
+import Promise from "bluebird";
+import util from "util";
+import { EventEmitter } from "events";
 let instanceEmitter, clientEmitter;
 const DELIM = "_DELIM_";
 const instances = {
```

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Freebind build
-FROM debian:buster-slim as freebind
+FROM debian:bookworm-slim as freebind
```
AndrewFerr: Nitpick, as long as we're updating this line
```suggestion
FROM debian:bookworm-slim AS freebind
```

Dockerfile
```diff
@@ -8,7 +8,7 @@ RUN git clone https://github.com/matrix-org/freebindfree.git
 RUN cd freebindfree && make
 
 # Typescript build
-FROM node:20 as builder
+FROM node:22 as builder
```
AndrewFerr: Nitpick #2
```suggestion
FROM node:22 AS builder
```

spec/util/test.js
```diff
@@ -1,15 +1,18 @@
...
 const ircMock = require("./irc-client-mock");
-const MockAppService = require("./app-service-mock");
+import MockAppService from "./app-service-mock";
+import testConfig from "../util/test-config.json";
 
 const USING_PG = process.env.IRCBRIDGE_TEST_ENABLEPG === "yes";
 
+// eslint-disable-next-line no-import-assign
 clientMock["@global"] = true;
 ircMock["@global"] = true;
```
AndrewFerr, line 17: Here's a way to avoid needing to ignore any rules, which seems to work when I lint/test locally:
```diff --git
diff --git a/spec/util/test.js b/spec/util/test.js
index 55ae5665..2d880d39 100644
--- a/spec/util/test.js
+++ b/spec/util/test.js
@@ -5,22 +5,21 @@ import { Client } from "pg";
 import Promise from "bluebird";
 import { AppServiceRegistration, Intent } from "matrix-appservice-bridge";
 import * as clientMock from "./bot-sdk-mock";
-// eslint-disable-next-line @typescript-eslint/no-require-imports
-const ircMock = require("./irc-client-mock");
+import * as ircMock from "./irc-client-mock";
 import MockAppService from "./app-service-mock";
 import testConfig from "../util/test-config.json";
 
 const USING_PG = process.env.IRCBRIDGE_TEST_ENABLEPG === "yes";
 
-// eslint-disable-next-line no-import-assign
-clientMock["@global"] = true;
-ircMock["@global"] = true;
 const main = proxyquire("../../lib/main.js", {
     "matrix-appservice": {
         AppService: MockAppService,
         "@global": true
     },
-    "matrix-org-irc": ircMock,
+    "matrix-org-irc": {
+        ...ircMock,
+        "@global": true
+    }
 });
 
 // Log the test case. Jasmine is a global var.
```

src/datastore/postgres/PgDataStore.ts
```diff
@@ -728,7 +728,7 @@ export class PgDataStore implements DataStore, ProvisioningStore {
         let currentVersion = await this.getSchemaVersion();
         while (currentVersion < PgDataStore.LATEST_SCHEMA) {
             log.info(`Updating schema to v${currentVersion + 1}`);
-            // eslint-disable-next-line @typescript-eslint/no-var-requires
+            // eslint-disable-next-line @typescript-eslint/no-var-requires,@typescript-eslint/no-require-imports
             const runSchema = require(`./schema/v${currentVersion + 1}`).runSchema;
```
AndrewFerr, line 732: For the sake of avoiding `require()`, consider using a dynamic `import` instead:
```suggestion
            const runSchema = (await import(`./schema/v${currentVersion + 1}`)).runSchema;
```

src/main.ts
AndrewFerr, line 1: Consider:
```diff --git
diff --git a/src/main.ts b/src/main.ts
index cb4c1285..cd051be0 100644
--- a/src/main.ts
+++ b/src/main.ts
@@ -104,10 +104,8 @@ export async function runBridge(
     }
 
     const maxSockets = config.advanced?.maxHttpSockets ?? 1000;
-    // eslint-disable-next-line @typescript-eslint/no-var-requires,@typescript-eslint/no-require-imports
-    require("http").globalAgent.maxSockets = maxSockets;
-    // eslint-disable-next-line @typescript-eslint/no-var-requires,@typescript-eslint/no-require-imports
-    require("https").globalAgent.maxSockets = maxSockets;
+    (await import("http")).globalAgent.maxSockets = maxSockets;
+    (await import ("https")).globalAgent.maxSockets = maxSockets;
 
     // run the bridge
     const ircBridge = new IrcBridge(config, reg, testOpts);
```

.github/workflows/test.yml
```diff
@@ -63,9 +71,13 @@ jobs:
...
+      - name: Use Node.js ${{ matrix.node_version }}
+        uses: actions/setup-node@v3
+        with:
+          node-version: ${{ matrix.node_version }}
       - run: yarn --strict-semver --frozen-lockfile
       - run: yarn test
         env:
           IRCBRIDGE_TEST_PGDB: "ircbridge_integtest"
-          IRCBRIDGE_TEST_PGURL: "postgresql://postgres_user:postgres_password@postgres"
+          IRCBRIDGE_TEST_PGURL: "postgresql://postgres_user:postgres_password@localhost"
```
AndrewFerr, line 82: FTR `localhost` is indeed the correct hostname to use for this.  Source: https://docs.github.com/en/actions/tutorials/use-containerized-services/create-postgresql-service-containers#running-jobs-directly-on-the-runner-machine

AndrewFerr (approved):
The test failures in Node 24 are because of `nedb` having rotted, which ought to be fixed in the bridge SDK (see https://github.com/matrix-org/matrix-appservice-bridge/pull/521).

And the E2E test failure happened even before this PR.

With that said, this PR looks good to go.  My comments are all non-critical suggestions (though I do think they're good to implement :slightly_smiling_face: )

spec/util/irc-client-mock.js
jaywink: Yeah, there were some issues caused by this, so I rolled back the converting to require. The issues happened elsewhere importing this file afaicr. In the interest of shipping this somewhat urgent bug fix I'm going to leave this as is now without spending more time on it, can be converted later the next time this comes up in an editor.

spec/util/test.js
jaywink: As per https://github.com/matrix-org/matrix-appservice-irc/pull/1858#discussion_r2459426450 I would like to avoid doing more adjustments to things, if possible, given time sensitivity and getting this bugfix done has already taken more time than was allocated to it. The change could be done as a separate pr to clean up, as it's not a functional change and thus doesn't need to go in to the next release.

src/datastore/postgres/PgDataStore.ts
jaywink: As per https://github.com/matrix-org/matrix-appservice-irc/pull/1858#discussion_r2459426450 I would like to avoid doing more adjustments to things, if possible, given time sensitivity and getting this bugfix done has already taken more time than was allocated to it. The change could be done as a separate pr to clean up, as it's not a functional change and thus doesn't need to go in to the next release.

src/main.ts
jaywink: As per https://github.com/matrix-org/matrix-appservice-irc/pull/1858#discussion_r2459426450 I would like to avoid doing more adjustments to things, if possible, given time sensitivity and getting this bugfix done has already taken more time than was allocated to it. The change could be done as a separate pr to clean up, as it's not a functional change and thus doesn't need to go in to the next release.

